PR #2, opened 2024-07-02
Houdini: Convert any ROP to a publishable AYON instance

### LINKED ISSUE #16: Better communication between publisher and rops

In Houdini addon we faced a problem when ROP node's parameters needs to be updated to match user selection in the publisher. 
e.g. changing the render target requires updating some parameters on the ROP node respectively.
image

Currently, The logic responsible for setting ROP's attributes to favorite values, lives in the extractors.
which is against pyblish concepts.

https://github.com/ynput/ayon-core/blob/6df1a97446d485badcd91c5ae353c5a7962ac757/server_addon/houdini/client/ayon_houdini/plugins/publish/extract_render.py#L25-L43

We need to figure out a way to sync the ROP's state with the selected options in the publisher.

There was an experimental PR https://github.com/ynput/ayon-core/pull/418 but it increased the complexity of the plugin.
Its idea was to set node's parameters when clicking save in the publisher and set the options to match the node when resetting the publisher.

### Additional context:

_No response_

BigRoy:
Implement Generic ROP publish from https://github.com/ynput/ayon-core/pull/542

---

## Changelog Description

This PR implements an idea for "lower level publishing" in Houdini. This implement **Generic ROP publishing**. Just create any Houdini ROP node (or custom Rop node HDA) and publish your product types from it!

> This PR adds a **Generic Creator**:
> Currently, It injects the ayon meta data onto ROP nodes to behave as though they were produced by various creators from Ayon. This ensures that validators and collectors are retained without loss.
> e.g. using Generic creator while setting product type to pointcache should yield similar/same results as using pointcache creator
> Here is an example using labs karma node and make it behave as though it was made by karma creator. 
> https://github.com/ynput/ayon-core/pull/542#issuecomment-2179333623

## Additional info

_As part of the Ynput Houdini Workgroup sessions I developed this quick prototype to expose a way to batch publish and ingest files. Consider it more of an exploration of what's possible then a "drop it in production now" ready-to-go solution._

#### Explainer

Yes, this requires some explanation. Here you go.

https://github.com/ynput/ayon-core/assets/2439881/5f767493-10bd-41f6-b54f-89313a96da33

What I forgot to add is that it currently still relies on detecting what the output files are for a ROP node based on a "file" parm that is often unique in Houdini per node. If anyone knows a way to just query the expected output files for a ROP node (similar to what PDG/TOPs seems to do I'd love to know!) but otherwise we'll just need to expand that list.

However, I also played with the idea of having "custom file list" attributes on the Node that when enabled could override the "Collector" logic and would instead use that list of files as the publish instance's files. So that e.g. one instance could also publish multiple representations. For that, @MustafaJafar did this lovely non-functional 'prototype' but it does get the idea across:

mustafa_collect_multiple_representations_to_one_instance

#### Demo scene file

The demo scene file:

[ynts_char_hero_pdg_v012.zip](https://github.com/ynput/ayon-core/files/15407437/ynts_char_hero_pdg_v012.zip)

## TODO 📔 ✏️ 

- [x] Use `$OS` as default _variant_ name.
- [x] Add "OnCreated" callback to allow a studio to configure which nodes to automatically imprint as 'publishable'. [See Fabia's first video here at 08:10](https://github.com/ynput/ayon-core/pull/542#issuecomment-2128044486) and [code here](https://github.com/fabiaserra/ayon-core/blob/release/alkemyx/server_addon/houdini/client/ayon_houdini/startup/OnCreated.py)
    - [ ] This is currently a first iteration - should still be exposed to project settings and allow user configuration.
- [x] The product types being an enum with predefined standards per node type.
- [x] The comment field being available on the ROP nodes
    - [x] This does actually already work if you'd add the product type to `ayon+settings://core/publish/collect_comment_per_instance` ✅ - I guess this is fine for now.
- [x] Add "Publish" button directly on the node similar to "self publish" or whatever we had for others
- [ ] Allow exposing to settings which node types may need to be "automatically imprinted" on create for a studio to customize, with also settings for allowed product types (the default type, default variant name, etc.)
- [ ] Implement good support for multiple representations, e.g. [like this](https://github.com/BigRoy/ayon-core/pull/6) and [Fabia's second video near the end](https://github.com/ynput/ayon-core/pull/542#issuecomment-2128044486).
- [ ] Make sure [this API approach](https://gist.github.com/MustafaJafar/bd2a388e4a6aa3613d64a186ebb6660c?permalink_comment_id=5067094#gistcomment-5067094) remains a goal of the PR.
- [ ] Keep in mind [Fabia's usage where their AX Publish Submitter](https://github.com/ynput/ayon-core/pull/542#issuecomment-2135399068)  can submit just the publish of existing files to the farm and create e.g. a nuke dependency job that generates custom previews of the intermediate media, etc.
- [ ] The product name is currently an 'exposed field' but should actually be locked off to the user for manual editing and should _always_ adhere to the `get_product_name` logic of the Creator and the product types so that they adhere to project templates correctly.
    - [ ] We need to make sure this isn't an expression that is heavy and re-evaluates the whole time. May need to be based on callback on "product type" and "variant" changes.

## Known issues / Blockers 📔 🔥 

- The publisher, creator and create context can't differentiate attributes per instance, as such it can't e.g. differentiate whether "split export and render job" makes sense or not for a particular instance's attribute definitions. See [comment](https://github.com/ynput/ayon-houdini/pull/2#discussion_r1667159224)
- It will only detect the single file defined by the ROP node's output path and does not recognize whether e.g. a USD ROP or GLTF rop may generate other sidecar files (which often also if moved along should also remap paths, etc. so it might be non-trivial to support those out of the box)
    - This basically makes it useless for GLTF exports, and in many cases useless for USD (if it were to write multiple files which also need remapping) and rendering (which may generate export files, need dedicated Deadline plug-ins to maybe render certain outputs (e.g. Husk) or may generate multiple sequences of files (like AOVs to separate files).
- We do not know or detect the colorspace of the output files (there is no 'generic' way for ANY ROP to return what colorspace it may result in.
- If certain ROPs may override the output FPS other than the workfile, we do not know it.
- Due to Creators enforcing a single product type for which the attribute definitions are populated/created based on the Creator's product type we run into issues with a generic Creator that basically switches product type on the instances dynamically. See [comments here](https://github.com/ynput/ayon-houdini/pull/2#discussion_r1674639415).

## Testing notes:

1. Check out the branch
2. Check out the explainer video
3. Test the demo scene file
4. Comment with great ideas on how to improve

Resolves #16


krishna8008:
Also, it would be great, if the node changed color or added a sticker on top when published, especially since this helps to identify the publish node easily in the scene. Not a very important one, but helps little.

BigRoy:
> Also, it would be great, if the node changed color or added a sticker on top when published, especially since this helps to identify the publish node easily in the scene. Not a very important one, but helps little.

Interesting request - when would you like the 'clear' this color or sticker? I can imagine it otherwise always having the sticker after the first publish you've done from the workfile.

MustafaJafar (commented):
I've a question about collect batch families type.

client/ayon_houdini/plugins/create/create_generic.py
```diff
@@ -0,0 +1,735 @@
...
+
+        Attributes defined in this method are exposed in
+            publish tab in the publisher UI.
+        """
+
+        render_target_items = {
+            "local": "Local machine rendering",
+            "local_no_render": "Use existing frames (local)",
+            "farm": "Farm Rendering",
+        }
```
MustafaJafar, line 665: Is it possible to add some condition here to add more items based on the instance ? 

I tried but I found out that it works per creator not per instance.

BigRoy: No - unfortunately not. The current Creator system design _and_ the Publisher UI do not allow it unfortunately. @antirotor I believe is aware of it and it's part of the "Extended Publisher" epic.

client/ayon_houdini/plugins/publish/collect_houdini_batch_families.py
```diff
@@ -0,0 +1,19 @@
...
+    targets = ["local", "remote"]
+    label = "Collect Data for Cache"
+
+    def process(self, instance):
+        # Do not allow `productType` to creep into the pyblish families
+        # so that e.g. any regular plug-ins for `pointcache` or alike do
+        # not trigger.
+        instance.data["family"] = "generic"
+        # TODO: Do not add the dynamic 'rop' family in the collector?
+        instance.data["families"] = ["generic", "rop"]
```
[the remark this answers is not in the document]
BigRoy: Actually - I didn't do that, because I believe that'd make it so that also the instance anatomy data collected for the instance would be the "generic" product type (which may affect the names for your resulting products or output files (I'm not entirely sure).

I recall there being a reason for doing it 'later' in publishing - but I'll try to reproduce some issues and maybe add a comment. But if it does work, and there are no side effects (🥁) then I'd love to remove this and rely on the Creator, yes.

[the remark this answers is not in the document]
BigRoy: I had a look at this - unfortunately we can't currently.

[This is the Collector](https://github.com/ynput/ayon-core/blob/2072f2fbe2bdaa09c96dd3c849bff65785c3b3bb/client/ayon_core/plugins/publish/collect_from_create_context.py#L89-L91) that turns the `CreateContext` 's `CreatedInstance` items into the pyblish instances. As you can see it sets the instance's family to the product type - which makes sense, but also means that we can't override the family explicitly separate from the product type.

We want to preserve e.g. product type "pointcache" because that's what the publisher attributes are also based on (like certain export options specific for a product type) whilst avoiding the usage of the regular product type plug-ins. We may need to redesign how the familie work for the 'original legacy rop instances' so that we maybe can differentiate which plug-ins are exposed where.

This is again, a CreateContext / Publisher system design issue or I'm missing how I could best redesign this to work around that.

BigRoy:
Adding screenshot of what it looks like now:

image

krishna8008:
Hey!

Just testing this branch, we cannot publish anything from houdini(workfiles/caches etc), somehow its getting conflict, I've attatched the error when we tried to publish workfiles with this branch

```Creator: Generic (io.ayon.creators.houdini.publish)

Error: The attempted operation failed.
Parameter name 'AYON_self_publish' is invalid or already exists

  File "C:\Users\avril3975\AppData\Local\Ynput\AYON\addons\core_0.4.4-dev.1\ayon_core\pipeline\create\context.py", line 2484, in _save_instance_changes
    creator.update_instances(update_list)
  File "C:\Users\avril3975\AppData\Local\Ynput\AYON\addons\houdini_0.3.12-dev.1\ayon_houdini\plugins\create\create_generic.py", line 393, in update_instances
    self.imprint(
  File "C:\Users\avril3975\AppData\Local\Ynput\AYON\addons\houdini_0.3.12-dev.1\ayon_houdini\plugins\create\create_generic.py", line 641, in imprint
    self.create_attribute_def_parms(instance_node, created_instance)
  File "C:\Users\avril3975\AppData\Local\Ynput\AYON\addons\houdini_0.3.12-dev.1\ayon_houdini\plugins\create\create_generic.py", line 624, in create_attribute_def_parms
    node.setParmTemplateGroup(parm_group)
  File "C:\PROGRA~1/SIDEEF~1/HOUDIN~1.547/houdini/python3.9libs\hou.py", line 16733, in setParmTemplateGroup
    return _hou.OpNode_setParmTemplateGroup(self, parm_template_group, rename_conflicting_parms)
Parameter name 'AYON_self_publish' is invalid or already exists
```

BigRoy:
> Hey!
> 
> Just testing this branch, we cannot publish anything from houdini(workfiles/caches etc), somehow its getting conflict, I've attatched the error when we tried to publish workfiles with this branch
> 
> ```
> 
> Error: The attempted operation failed.
> Parameter name 'AYON_self_publish' is invalid or already exists
> 
> Traceback (most recent call last):
>   File "C:\Users\avril3975\AppData\Local\Ynput\AYON\addons\core_0.4.4-dev.1\ayon_core\pipeline\create\context.py", line 2484, in _save_instance_changes
>     creator.update_instances(update_list)
>   File "C:\Users\avril3975\AppData\Local\Ynput\AYON\addons\houdini_0.3.12-dev.1\ayon_houdini\plugins\create\create_generic.py", line 393, in update_instances
>     self.imprint(
>   File "C:\Users\avril3975\AppData\Local\Ynput\AYON\addons\houdini_0.3.12-dev.1\ayon_houdini\plugins\create\create_generic.py", line 641, in imprint
>     self.create_attribute_def_parms(instance_node, created_instance)
>   File "C:\Users\avril3975\AppData\Local\Ynput\AYON\addons\houdini_0.3.12-dev.1\ayon_houdini\plugins\create\create_generic.py", line 624, in create_attribute_def_parms
>     node.setParmTemplateGroup(parm_group)
>   File "C:\PROGRA~1/SIDEEF~1/HOUDIN~1.547/houdini/python3.9libs\hou.py", line 16733, in setParmTemplateGroup
>     return _hou.OpNode_setParmTemplateGroup(self, parm_template_group, rename_conflicting_parms)
> hou.OperationFailed: The attempted operation failed.
> Parameter name 'AYON_self_publish' is invalid or already exists
> ```

Thanks @krishnaavril - should be solved now. It will now log a warning on creation.
It's because it tries to create it twice **somehow?** you have the creation of self publish button enabled in settings.

But that would only have been the case if you triggered the generic creation on a node that was already a publishable instance. I've changed the code, but would still be lovely if you could share a screen capture of how you get it to the state that it breaks.

client/ayon_houdini/plugins/create/create_generic.py
```diff
@@ -0,0 +1,735 @@
...
+                    creator_attributes[creator_key] = value
+                    del node_data[key]  # remove from original
+
+            node_data["creator_attributes"] = creator_attributes
+            node_data["publish_attributes"] = publish_attributes
+
+            created_instance = CreatedInstance.from_existing(
+                node_data, self
+            )
+            self._add_instance_to_context(created_instance)
```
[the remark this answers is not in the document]
BigRoy: Unfortunately - this won't work, see [comment](https://github.com/ynput/ayon-houdini/pull/2#discussion_r1674639415) here. I'll add a todo in the code instead.

client/ayon_houdini/plugins/publish/collect_houdini_batch_families.py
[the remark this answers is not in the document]
BigRoy: Update the code with`TODO` explaining the issue. Whether there are better ways, not sure. @iLLiCiTiT @antirotor

client/ayon_houdini/plugins/create/create_generic.py
BigRoy: @MustafaJafar it may be that this PR https://github.com/ynput/ayon-core/pull/860 by @iLLiCiTiT may solve this. The question then becomes _WHAT_ would define which of the targets would be available?

krishnaavril:
Thanks, Yes I do have the self-publish button enabled in Ayon settings.

```But that would only have been the case if you triggered the generic creation on a node that was already a publishable instance.```
No, I tried this with our inhouse otl, adding the parameter and otl name hardcoded in the ayon code which allows publishing our own cache and render nodes.

Will test these pushed changes today, thanks

**Comment by krishnaavril** [comment]
I've tested the code with the labs karma node, I think the issue is still there

```Creator: Generic (io.ayon.creators.houdini.publish)

Error: The attempted operation failed.
Parameter name 'AYON_self_publish' is invalid or already exists

  File "C:\Users\avril3975\AppData\Local\Ynput\AYON\addons\core_0.4.5-dev.3\ayon_core\pipeline\create\context.py", line 1278, in _save_instance_changes
    creator.update_instances(update_list)
  File "C:\Users\avril3975\AppData\Local\Ynput\AYON\addons\houdini_0.3.14-dev\ayon_houdini\plugins\create\create_generic.py", line 392, in update_instances
    self.imprint(
  File "C:\Users\avril3975\AppData\Local\Ynput\AYON\addons\houdini_0.3.14-dev\ayon_houdini\plugins\create\create_generic.py", line 640, in imprint
    self.create_attribute_def_parms(instance_node, created_instance)
  File "C:\Users\avril3975\AppData\Local\Ynput\AYON\addons\houdini_0.3.14-dev\ayon_houdini\plugins\create\create_generic.py", line 623, in create_attribute_def_parms
    node.setParmTemplateGroup(parm_group)
  File "C:\PROGRA~1/SIDEEF~1/HOUDIN~1.547/houdini/python3.10libs\hou.py", line 16733, in setParmTemplateGroup
    return _hou.OpNode_setParmTemplateGroup(self, parm_template_group, rename_conflicting_parms)
Parameter name 'AYON_self_publish' is invalid or already exists
```

after create was done and when I clicked on publish this error popped up!

client/ayon_houdini/plugins/create/create_generic.py
```diff
@@ -0,0 +1,751 @@
...
+            "AYON_self_publish",
+            "Publish",
+            help="Directly publishes only this node and any inputs with a "
+                 "comment entered in a popup prompt. All other instances will "
+                 "be disabled in the publish.",
+            script_callback="from ayon_houdini.api.lib import "
+                            "self_publish; self_publish()",
+            script_callback_language=hou.scriptLanguage.Python,
+        )
+        ayon_folder.addParmTemplate(publish_button_parm)
```
MustafaJafar, line 450: This should fix the bug 
```suggestion
        publish_button_parm = parm_group.find("AYON_self_publish")
        if not publish_button_parm:
            publish_button_parm = hou.ButtonParmTemplate(
                "AYON_self_publish",
                "Publish",
                help="Directly publishes only this node and any inputs with a "
                    "comment entered in a popup prompt. All other instances will "
                    "be disabled in the publish.",
                script_callback="from ayon_houdini.api.lib import "
                                "self_publish; self_publish()",
                script_callback_language=hou.scriptLanguage.Python,
            )
            ayon_folder.addParmTemplate(publish_button_parm)
```

MustafaJafar: Also, the above logic always creates self publish button as it  doesn't take the value of the following setting into consideration: 
- `ayon+settings://houdini/general/add_self_publish_button`

For reference, the super class `plugin.HoudiniCreator` already adds the self publish button in `apply_settings` method (which is not overridden in `CreateHoudiniGeneric`)
That's why we may have a publish button duplicate.

https://github.com/ynput/ayon-houdini/blob/c3ab366c047e63881fa4cd95de79284a8605bc11/client/ayon_houdini/api/plugin.py#L295-L297

MustafaJafar: we can also do this which leverage `self.add_publish_button` that is set in `apply_settings`.

```suggestion
        publish_button_parm = parm_group.find("AYON_self_publish")
        if not publish_button_parm and self.add_publish_button:
            publish_button_parm = hou.ButtonParmTemplate(
                "AYON_self_publish",
                "Publish",
                help="Directly publishes only this node and any inputs with a "
                    "comment entered in a popup prompt. All other instances will "
                    "be disabled in the publish.",
                script_callback="from ayon_houdini.api.lib import "
                                "self_publish; self_publish()",
                script_callback_language=hou.scriptLanguage.Python,
            )
            ayon_folder.addParmTemplate(publish_button_parm)
```
But, this has two downsides:
1. it won't add the button until we change any attribute (as it won't be marked as a change on its own)
2. the button will be added at the end of the list because we are using `.addParmTemplate` and folders doesn't support `insertBefore`.
image

client/ayon_houdini/api/node_wrap.py
```diff
@@ -0,0 +1,41 @@
...
+        pre_create_data={
+            "node": node
+        }
+    )
+
+
+# TODO: Move this choice of automatic 'imprint' to settings so studio can
+#   configure which nodes should get automatically imprinted on creation
+# TODO: Do not import and reload the creator plugin file
+from ayon_houdini.plugins.create import create_generic
```
iLLiCiTiT, line 26: Please, never import plugins.

client/ayon_houdini/startup/scripts/OnCreated.py
```diff
@@ -0,0 +1,18 @@
...
+# As such, preferably the code here should run fast to avoid slowing down node
+# creation. Note: It does not trigger on existing nodes for scene open nor on
+# node copy-pasting.
+from ayon_core.lib import env_value_to_bool, is_dev_mode_enabled
+from ayon_houdini.api import node_wrap
+
+
+# Allow easier development by automatic reloads
+# TODO: remove this
+if is_dev_mode_enabled():
```
iLLiCiTiT, line 11: I don't agree with this, but I'm not responsible for houdini repository.

Do you know it will "reload" the module only for newly imported files (only plugins), and may break other imports because of module pointers cleanup (don't know if still happens in Py3).

client/ayon_houdini/plugins/create/create_generic.py
```diff
@@ -0,0 +1,751 @@
...
+            # but explicitly collected
+            node_path = node.path()
+            node_data["instance_id"] = node_path
+            node_data["instance_node"] = node_path
+            node_data["families"] = self.get_publish_families()
+
+            # Read creator and publish attributes
+            publish_attributes = {}
+            creator_attributes = {}
+            for key, value in dict(node_data).items():
```
iLLiCiTiT, line 364: Why it's not stored as json string?

client/ayon_houdini/plugins/create/create_generic.py
```diff
@@ -0,0 +1,751 @@
...
+    product_type = "generic"
+    icon = "male"
+    description = "Make any ROP node publishable."
+
+    render_target = "local_no_render"
+    default_variant = USE_DEFAULT_NODE_NAME
+    default_variants = ["Main", USE_DEFAULT_NODE_NAME]
+
+    # TODO: Move this to project settings
+    node_type_product_types: Dict[str, NodeTypeProductTypes] = {
```
iLLiCiTiT, line 223: Why are not the types resolved during collection?

client/ayon_houdini/startup/scripts/OnCreated.py
BigRoy: Correct, this should not be in live code - this is just there for my own dev. Ignore it for now, the "remove this" todo completely applies.

client/ayon_houdini/plugins/create/create_generic.py
BigRoy: So that the attributes are actually accessible as options on the node for the user to click. We want them to be like native houdini parms so houdini users can do with them using houdini logic whatever they want - that's the whole point. ;)

client/ayon_houdini/plugins/create/create_generic.py
BigRoy: These are supposed to be the "defaults" for new nodes to be created - not the ones currently set on the instance.

client/ayon_houdini/startup/scripts/OnCreated.py
```diff
@@ -0,0 +1,18 @@
...
+    import importlib
```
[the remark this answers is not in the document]
BigRoy: Yes, but code should actually be removed - see the todo.

client/ayon_houdini/api/node_wrap.py
BigRoy: > Please, never import plugins. Shouldn't node_type_product_types from the creator be moved here (or to different pipeline file)?

It should be taken from settings instead - or stored elsewhere.

iLLiCiTiT: I'm not sure about settings. Those values don't seem like something what TD can fill, they actually look like they are quite pre-determined.

BigRoy: They are - until you get into the realm of custom HDAs (custom nodes) a studio develops on their own inside of Houdini (or get shipped with addons/plugins for Houdini and are not native to Houdini). We may not know about ALL nodes from our codebase.

client/ayon_houdini/plugins/create/create_generic.py
iLLiCiTiT: So `"handleStart" = 0` will become `"handleStart": True`.

BTW considering in progress feature of create context: "Value change based callbacks to update attribute definitions or values." will break this assumption that you can freely change create and publish attributes with no consequences.

BigRoy: ### Does exposing these publisher UI attributes on the node actually make sense?

> BTW considering in progress feature on create context: "Value change based callbacks to update attribute definitions or values." will break this assumption that you can freely change create and publish attributes with no consequences.

Yes, makes sense. There's just a plethora of reasons why one WOULD want to actually exploit this type of full control in Houdini but I'm not confident whether those are production use cases that really make sense over time.

@fabiaserra @krishnaavril @MustafaJafar how much use would you say there is in _actually_ having all these publisher UI options exposed as separate Houdini parms? Like, would you really ever have those parms reference some other attribute in a production scenario? Or is setting those via the publisher UI sufficient (it would allow us to create more dynamic UI over time without 'unexpected changes' coming back from Houdini).

See [here](https://github.com/ynput/ayon-houdini/pull/2#issuecomment-2223867770) for a screenshot with some of those exposed attributes.

---

> So "handleStart" = 0 will become "handleStart": True (as an example).

Luckily those are not exposed like that in Houdini 🙈 - but yes, that definitely points to a crucial bug that would need solving. However, 'setting' them om the attribute definitions in publisher UI themselves would then automatically turn that back into 0 and 1 so it may be 'ok' but it's definitely illogical code. Thanks for point it out!

fabiaserra: I'm not sure I understand the questions but let me try explain why:

> how much use would you say there is in actually having all these publisher UI options exposed as separate Houdini parms?

What do you mean by "separate"? If you just mean if there's a point on having these options exposed on the Houdini nodes natively... yeah, there's a lot of use, it allows artists to use Houdini workflows and not be limited by another UI, making it very easy entry point for Houdini users (the advanced options are probably confusing and irrelevant for anyone though?)

> Like, would you really ever have those parms reference some other attribute in a production scenario? 

Potentially, that's the beauty of Houdini

> Or is setting those via the publisher UI sufficient 

Not as it is, but not because just of the UI setting parameters... the whole publish framework being coupled with the export of the nodes is very limited IMO as I have already explained a few times

MustafaJafar: Additionally, the node parameters and the publisher attributes are synced. 
From the node to the publisher and from the publisher to the node. 
That logic is implemented in `collect_instances` by creating the instance manually and adding it to the context and `update_instances` by the `imprint` method.

client/ayon_houdini/api/node_wrap.py
MustafaJafar: So, basically, pipeline admins would need to expand that list as needed e.g. to add their own types or like what krishna did at early stage of this PR https://github.com/ynput/ayon-core/pull/542#issuecomment-2179333623 when he tested `labs::karma::2.0` node that we don't usually support in the current status of the addon, correct?

MustafaJafar: Does moving `node_type_product_types` to settings resolve this?

client/ayon_houdini/plugins/create/create_generic.py
MustafaJafar: Once `node_type_product_types` is moved to settings, this line should be replaced by applying/reading the settings?

client/ayon_houdini/plugins/create/create_generic.py
MustafaJafar: > is use existing frames (farm) not supported yet?

Render targets is not yet implemented for all product types.

MustafaJafar: @BigRoy  I gave it a test and it actually works although it was hard for me to learn how to use it. I can't really tell if there are any alternative solutions but, we can discuss that on discord.
https://github.com/ynput/ayon-core/pull/860#pullrequestreview-2326349076

client/ayon_houdini/plugins/create/create_generic.py
krishnaavril: > ### Does exposing these publisher UI attributes on the node actually make sense?
> > BTW considering in progress feature on create context: "Value change based callbacks to update attribute definitions or values." will break this assumption that you can freely change create and publish attributes with no consequences.
> 
> Yes, makes sense. There's just a plethora of reasons why one WOULD want to actually exploit this type of full control in Houdini but I'm not confident whether those are production use cases that really make sense over time.
> 
> @fabiaserra @krishnaavril @MustafaJafar how much use would you say there is in _actually_ having all these publisher UI options exposed as separate Houdini parms? Like, would you really ever have those parms reference some other attribute in a production scenario? Or is setting those via the publisher UI sufficient (it would allow us to create more dynamic UI over time without 'unexpected changes' coming back from Houdini).
> 
> See [here](https://github.com/ynput/ayon-houdini/pull/2#issuecomment-2223867770) for a screenshot with some of those exposed attributes.
> 
> In this case in particular the **Creator Attributes** and **Publish Attributes**
> 
> image
> 
> > So "handleStart" = 0 will become "handleStart": True (as an example).
> 
> Luckily those are not exposed like that in Houdini 🙈 - but yes, that definitely points to a crucial bug that would need solving. However, 'setting' them om the attribute definitions in publisher UI themselves would then automatically turn that back into 0 and 1 so it may be 'ok' but it's definitely illogical code. Thanks for point it out!



> ### Does exposing these publisher UI attributes on the node actually make sense?
> > BTW considering in progress feature on create context: "Value change based callbacks to update attribute definitions or values." will break this assumption that you can freely change create and publish attributes with no consequences.
> 
> Yes, makes sense. There's just a plethora of reasons why one WOULD want to actually exploit this type of full control in Houdini but I'm not confident whether those are production use cases that really make sense over time.
> 
> @fabiaserra @krishnaavril @MustafaJafar how much use would you say there is in _actually_ having all these publisher UI options exposed as separate Houdini parms? Like, would you really ever have those parms reference some other attribute in a production scenario? Or is setting those via the publisher UI sufficient (it would allow us to create more dynamic UI over time without 'unexpected changes' coming back from Houdini).
> 
> See [here](https://github.com/ynput/ayon-houdini/pull/2#issuecomment-2223867770) for a screenshot with some of those exposed attributes.
> 
> In this case in particular the **Creator Attributes** and **Publish Attributes**
> 
> image
> 
> > So "handleStart" = 0 will become "handleStart": True (as an example).
> 
> Luckily those are not exposed like that in Houdini 🙈 - but yes, that definitely points to a crucial bug that would need solving. However, 'setting' them om the attribute definitions in publisher UI themselves would then automatically turn that back into 0 and 1 so it may be 'ok' but it's definitely illogical code. Thanks for point it out!

Hello @BigRoy ! I think it was completly fine having properties in ayon publisher window, Honestly Me and my teams only visualise purpose, there is no change scenario seen in production from our end. (I was writing this coz we tested this rop publish when houdini is in core)



Can we write a logic where ayon filters the rop node inside any studio specific otl?
This will help to identify what kind of files will be writting(pointcache/renders), also this will help on creating generic for any otl(without we hardcoded in the script), this should essentially open the possiblity of adding generic for any otl with no code changes(or a ayon frontend parameter will be good too)
Like we know studio uses own otls, we can put in the documentation that to add picture parameter added in there otls, thus it refers the path and publishes.

Let me know if this make sense.

client/ayon_houdini/plugins/create/create_generic.py
krishnaavril: Adding use exiting frames using farm is essential, there is a [dicussion](https://discord.com/channels/517362899170230292/517382145552154634/1287415312139943967) happened in couple of days ago, about this feature.
Local publish is tkaing huge amount of time and leaving houdini freezed until the publish get done. This a time saving for artists.

client/ayon_houdini/api/node_wrap.py
BigRoy: > Does moving `node_type_product_types` to settings resolve this?

Yes - but for now I've shuffled the imports around. But moving it to settings is still very relevant down the line.

client/ayon_houdini/api/lib.py
```diff
@@ -370,6 +370,9 @@ def read(node):
             except json.JSONDecodeError:
                 # not a json
                 pass
+        elif parameter.parmTemplate().type() == hou.parmTemplateType.Toggle:
```
iLLiCiTiT, line 373:
```suggestion
        elif parameter.parmTemplate().type() is hou.parmTemplateType.Toggle:
```

BigRoy:
I'm closing this PR in favor of https://github.com/ynput/ayon-houdini/pull/122

That PR allows to do what we're doing something similar to what we're doing here but more on the level of Houdini - in a way that can also easily be exposed to PDG. It's lower level but would allow much greater flexibility.

PR #4, opened 2024-07-03
Chore: Require newer ayon core

iLLiCiTiT:
## Description
After PR https://github.com/ynput/ayon-core/pull/295 the addon require ayon-core > 0.4.0 .

## Additional information
This should be already part of that PR and ayon-houdini release 0.3.9 .

### DIFF
--- a/package.py
+++ b/package.py
@@ -5,6 +5,6 @@ version = "0.3.10-dev.1"
 client_dir = "ayon_houdini"
 
 ayon_required_addons = {
-    "core": ">0.3.2",
+    "core": ">0.4.0",
 }
 ayon_compatible_addons = {}


BigRoy (approved):
Description and change makes sense - I'm not sure how to confirm the compatibility check itself actually works but I suppose that's an existing feature that we know should work?

iLLiCiTiT:
> Description and change makes sense - I'm not sure how to confirm the compatibility check itself actually works but I suppose that's an existing feature that we know should work?

Happens on bundle creation on AYON server.

BigRoy:
So I tested this out and it does work:

image
image
image

However, the PR description makes it sound like anything ABOVE 0.3.9 should be fine - but the compatibility check only allows **above** 0.4.0 and does not allow exactly 0.4.0. Is that correct?

iLLiCiTiT:
>  but the compatibility check only allows above 0.4.0 and does not allow exactly 0.4.0. Is that correct?

It should allow greater than 0.4.0 . Core release 0.4.0 does not contain changes that current houdini develop requires. The changes are in current develop of core addon.

PR #26, opened 2024-07-05
Load USD loaders with filepath from entity URI

BigRoy:
Changes the load behavior for the loaders that load to LOPs (e.g. `/stage`) to use AYON Entity URIs that can be resolved with the [AYON USD Resolver](https://github.com/ynput/ayon-usd-resolver).

### TODO

- [x] Use ayon api endpoint to resolve to the entity URI
- [x] Allow toggle in project settings between loading to AYON entity URI or regular filepath.
- [x] The Load Asset LOP may need custom work because it manages the filepath using HDA utils maybe? This may need a dedicated "use entity URI" toggle on the HDA itself, could be a follow up issue/PR to this one.
    - [x] Find a way to set the "default" using studio choice from settings when creating _new nodes_. This may need to be a "on created" callback using some sort of cached settings.

### Additional context

Example URI after load is:
```
ayon+entity://ayontest/asset/char_roy&product=usdAsset&version=v004&representation=usd
```

On the Load Asset LOP and Load Shot LOP there's now a toggle that defines whether it resolves as entity URI or not:

image

### Testing notes

Build package and upload new addon.
Test with settings enabled/disabled:
image

```
ayon+settings://houdini/load/LOPLoadAssetLoader/use_ayon_entity_uri
ayon+settings://houdini/load/LOPLoadShotLoader/use_ayon_entity_uri
ayon+settings://houdini/load/USDSublayerLoader/use_ayon_entity_uri
ayon+settings://houdini/load/USDReferenceLoader/use_ayon_entity_uri
ayon+settings://houdini/load/SopUsdImportLoader/use_ayon_entity_uri
```

1. All loaders should still work. 
2. The entity URIs should be correct and work with AYON USD Resolver
3. Updating/managing of versions should still work.


Lypsolon (commented):
in not sure if we should ask the user to use uri at every step of the way tbh. 
there are multiple Prs now where you have to enable Uri usage. 

i dont say we should not allow for fine grain control if needed but i believe we should have one big button that says Use uri. and then the option to disable that for individual Systems.

BigRoy:
> in not sure if we should ask the user to use uri at every step of the way tbh. there are multiple Prs now where you have to enable Uri usage.
> 
> i dont say we should not allow for fine grain control if needed but i believe we should have one big button that says Use uri. and then the option to disable that for individual Systems.

It's not the user by the way, it's the studio "admin" defining that once in settings. So yes, it now provides granularity. We could do it differently by making all these enums that are three options:

- Use default
- Enabled
- Disabled

And when "Use Default" is set it uses the big button.
Happy to hear of better ideas, but I definitely wanted to offer some granularity but wasn't sure what the best approach would be to maintain but also making it easy to manage the setting.

@iLLiCiTiT @Innders ideas?

antirotor (changes requested):
There is a slight glitch in URIs:

This URI works for me:
```
ayon+entity://USD_DEMO/assets/box?product=usdAsset&version=v002&representation=usd
```

and this is filled by this PR:
```
ayon+entity://USD_DEMO/assets/box&product=usdAsset&version=v002&representation=usd
```

The issue is here: `...box&product...` - it should use question mark.

This is handled by server call if I am not mistaken, so please disregard this if this is the case and we need to fix it there.

BigRoy:
> There is a slight glitch in URIs:
> 
> This URI works for me:
> 
> ```
> ayon+entity://USD_DEMO/assets/box?product=usdAsset&version=v002&representation=usd
> ```
> 
> and this is filled by this PR:
> 
> ```
> ayon+entity://USD_DEMO/assets/box&product=usdAsset&version=v002&representation=usd
> ```
> 
> The issue is here: `...box&product...` - it should use question mark.
> 
> This is handled by server call if I am not mistaken, so please disregard this if this is the case and we need to fix it there.

Correct - I fixed it this morning: https://github.com/ynput/ayon-backend/pull/329
You'll need latest `dev` tag of the `ayon-docker` to get the fix.

@antirotor can you test and approve if it works?

Do note that there was also an issue with using `os.path.normpath` on these paths because it'd strip the double forward slash - so please check each loader individually to see if they are not lacking those slashes (and even better if you can test whether the resolver resolves it correctly!)

antirotor (commented):
So it works now, resolves correctly too. One exception is **Load USD to SOPs** - not sure if this is actually resolvable by the resolver?

image

antirotor (approved):
works now as expected

PR #29, opened 2024-07-05
Use task entity attributes instead of folder attributes

### LINKED ISSUE #14: Use Task level attributes instead of folder

### Current Behavior:

Houdini addon uses folder level attributes e.g. frame range

### Expected Behavior:

It should use task level attributes.

### Version

1.0.0

Windows

### Steps To Reproduce:

This is an example,
1. On AYON Server, change frame range on task
2. In Houdini, use `Set Frame Range`, frame range will be reset to folder frame range not task.

### Relevant log output:

_No response_

### Additional context:

_No response_

BigRoy:
### Changes

Use task entity attributes instead of folder attributes

### Additional context

Fixes #14

### DIFF
--- a/client/ayon_houdini/api/lib.py
+++ b/client/ayon_houdini/api/lib.py
@@ -14,7 +14,6 @@ from ayon_core.lib import StringTemplate
 from ayon_core.settings import get_current_project_settings
 from ayon_core.pipeline import (
     Anatomy,
-    get_current_folder_path,
     registered_host,
     get_current_context,
     get_current_host_name,
--- a/client/ayon_houdini/api/lib.py
+++ b/client/ayon_houdini/api/lib.py
@@ -402,18 +402,12 @@ def reset_framerange(fps=True, frame_range=True):
     if frame_range:
 
         # Set Start and End Frames
-        frame_start = task_entity.get("frameStart")
-        frame_end = task_entity.get("frameEnd")
+        task_attrib = task_entity["attrib"]
+        frame_start = task_attrib.get("frameStart", 0)
+        frame_end = task_attrib.get("frameEnd", 0)
 
-        if frame_start is None or frame_end is None:
-            folder_path = get_current_folder_path()
-            task_name = task_entity["name"]
-            log.warning("No edit information found for '%s' > '%s'",
-                        folder_path, task_name)
-            return
-
-        handle_start = task_entity.get("handleStart", 0)
-        handle_end = task_entity.get("handleEnd", 0)
+        handle_start = task_attrib.get("handleStart", 0)
+        handle_end = task_attrib.get("handleEnd", 0)
 
         frame_start -= int(handle_start)
         frame_end += int(handle_end)
--- a/client/ayon_houdini/api/lib.py
+++ b/client/ayon_houdini/api/lib.py
@@ -22,7 +22,7 @@ from ayon_core.pipeline import (
 )
 from ayon_core.pipeline.create import CreateContext
 from ayon_core.pipeline.template_data import get_template_data
-from ayon_core.pipeline.context_tools import get_current_folder_entity
+from ayon_core.pipeline.context_tools import get_current_task_entity
 from ayon_core.tools.utils import PopupUpdateKeys, SimplePopup
 from ayon_core.tools.utils.host_tools import get_tool_by_name
 
@@ -35,12 +35,12 @@ log = logging.getLogger(__name__)
 JSON_PREFIX = "JSON:::"
 
 
-def get_folder_fps(folder_entity=None):
-    """Return current folder fps."""
+def get_entity_fps(entity=None):
+    """Return current task fps or fps from an entity."""
 
-    if folder_entity is None:
-        folder_entity = get_current_folder_entity(fields=["attrib.fps"])
-    return folder_entity["attrib"]["fps"]
+    if entity is None:
+        entity = get_current_task_entity(fields=["attrib.fps"])
+    return entity["attrib"]["fps"]
 
 
 def get_output_parameter(node):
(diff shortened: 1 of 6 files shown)


client/ayon_houdini/plugins/publish/collect_asset_handles.py
```diff
@@ -50,9 +50,9 @@ def process(self, instance):
 
         attr_values = self.get_attr_values_from_data(instance.data)
         if attr_values.get("use_handles", self.use_asset_handles):
-            folder_attributes = instance.data["folderEntity"]["attrib"]
-            handle_start = folder_attributes.get("handleStart", 0)
-            handle_end = folder_attributes.get("handleEnd", 0)
+            task_attributes = instance.data["taskEntity"]["attrib"]
```
BigRoy, line 53: This may need to have a fallback to `folderEntity` because `task` may not be required for publish instances. Correct @iLLiCiTiT ?

Does that mean the 'label' changes here from Folder to Task across these plug-ins is actually counter intuitive? Any better names we can think of that represents' the Instance's folder/task attributes?

iLLiCiTiT: There is `ValidateInstanceInContextHoudini` but it is optional, so probably yes.

client/ayon_houdini/api/lib.py
```diff
@@ -646,43 +644,38 @@ def get_output_children(output_node, include_sops=True):
...
-        ))
+        print(f"No resolution information found in entity: '{entity}'")
         return None
 
     return int(resolution_width), int(resolution_height)
 
 
-def set_camera_resolution(camera, folder_entity=None):
-    """Apply resolution to camera from folder entity of the publish"""
+def set_camera_resolution(camera, task_entity=None):
```
iLLiCiTiT, line 672:
```suggestion
def set_camera_resolution(camera, entity=None):
```

client/ayon_houdini/api/lib.py
```diff
@@ -729,23 +725,25 @@ def get_current_context_template_data_with_folder_attrs():
...
+            attributes["frameStartHandle"] = frame_start - handle_start
+        if frame_end is not None and handle_end is not None:
+            attributes["frameEndHandle"] = frame_end + handle_end
 
     template_data = get_template_data(
         project_entity, folder_entity, task_entity, host_name
     )
     template_data["root"] = anatomy.roots
     template_data["folderAttributes"] = folder_attributes
+    template_data["taskAttributes"] = task_attributes
```
iLLiCiTiT, line 746: Where are `folderAttributes` and `taskAttributes` actually used?

BigRoy: These I believe are used by the feature to update houdini session variables based on context changes. So you could e.g. store houdini-specific variables like the folder or tasks's frame range to be used dynamically inside parms.

Basically related to this feature by @MustafaJafar I believe: https://github.com/ynput/OpenPype/pull/5651

MustafaJafar: > Where are `folderAttributes` and `taskAttributes` actually used?

It's used in this feature to manage Houdini variables. 
image

and this how it look like in Houdini.
image

client/ayon_houdini/api/lib.py
```diff
@@ -392,30 +391,29 @@ def parm_values(overrides):
...
+            folder_path = get_current_folder_path()
+            task_name = task_entity["name"]
+            log.warning("No edit information found for '%s' > '%s'",
+                        folder_path, task_name)
             return
 
-        handle_start = folder_attributes.get("handleStart", 0)
-        handle_end = folder_attributes.get("handleEnd", 0)
+        handle_start = task_entity.get("handleStart", 0)
+        handle_end = task_entity.get("handleEnd", 0)
```
MustafaJafar, line 416:
```suggestion
        attributes = task_entity["attrib"]
        frame_start = attributes.get("frameStart")
        frame_end = attributes.get("frameEnd")

        if frame_start is None or frame_end is None:
            folder_path = get_current_folder_path()
            task_name = task_entity["name"]
            log.warning("No edit information found for '%s' > '%s'",
                        folder_path, task_name)
            return

        handle_start = attributes.get("handleStart", 0)
        handle_end = attributes.get("handleEnd", 0)
```

MustafaJafar (commented):
Should this one updated too ? 

https://github.com/ynput/ayon-houdini/blob/701d46ce94f79955519ff262f250369785a7ccc4/client/ayon_houdini/plugins/load/load_camera.py#L124

MustafaJafar:
Other than my comments. it works on my side.

client/ayon_houdini/api/lib.py
BigRoy: Ah yes, obviously! Thanks. Even, better.... let me kill that folder fallback since it shouldn't be there.

BigRoy:
> Should this one updated too ?
> 
> https://github.com/ynput/ayon-houdini/blob/701d46ce94f79955519ff262f250369785a7ccc4/client/ayon_houdini/plugins/load/load_camera.py#L124

This should already be the case? :/ 

https://github.com/ynput/ayon-houdini/blob/e20fbe045ed1480d01e529ccac054ec8a6c344b7/client/ayon_houdini/plugins/load/load_camera.py#L124

BigRoy:
Feel free to re-test @MustafaJafar and merge if it works.

MustafaJafar:
> This should already be the case? :/

oh, I think I've mistaken `context["folder"]` and folder entities.

client/ayon_houdini/api/lib.py
```diff
@@ -14,15 +14,14 @@
 from ayon_core.settings import get_current_project_settings
 from ayon_core.pipeline import (
     Anatomy,
-    get_current_project_name,
     get_current_folder_path,
```
MustafaJafar, line 17:
```suggestion
```

MustafaJafar (approved):
It works on my side.
Thank you \^^/

Issue #32, opened 2024-07-08
Load Asset LOP: Representation does not update when saving to another context


BigRoy:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior:

When the AYON Load Asset LOP is set to load using e.g. folder path `$AYON_FOLDER_PATH` then upon change of that Houdini variable the 'parameter changed' callback does not seem to trigger for the HDA and thus the representation ID remains the old value - which seems counter-intuitive.

![image](https://github.com/ynput/ayon-houdini/assets/2439881/4ae07173-2682-43d4-8fdc-4221d6b90010)

### Expected Behavior:

On change of one of the variables, like `$AYON_FOLDER_PATH` preferably we're able to respond to that change and then re-evaluate the loaded representation id.

### Version

1.0.0

### What platform you are running on?

Windows

### Steps To Reproduce:

1. Load using Load Asset LOP
2. Save scene to another context.
3. Load asset lop is not updated to new context.

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the bug report.

MustafaJafar:
I think the source of the problem is the representation name.
we should update the representation name before calling `get_representation_id` 

https://github.com/ynput/ayon-houdini/blob/701d46ce94f79955519ff262f250369785a7ccc4/client/ayon_houdini/api/hda_utils.py#L281


![image](https://github.com/ynput/ayon-houdini/assets/20871534/7388c0a3-47c6-448e-81e9-d33d8a0183ec)

So, I have to change it manually, in order to continue the execution flow as expected.
![image](https://github.com/ynput/ayon-houdini/assets/20871534/406e0d04-7405-48c0-9958-7d6ccccddb84)

BigRoy:
This will be fixed by https://github.com/ynput/ayon-houdini/pull/112 because the expressions will respond to the parameter value changes.

PR #39, opened 2024-07-11
asset HDA LOP: Implement and use  select product dialog

MustafaJafar:
## Changelog description
Use dialog to pick product and remove redundant Houdini parameter.

Animation_88

## Additional info
There's a bug where the representation name is not updating. I can reproduce it in develop branch.
oh, there's an issue for it already https://github.com/ynput/ayon-houdini/issues/32

## Testing Notes:
1. use asset lop.
2. try the product picker.
3. everything should work as usual.


BigRoy:
Thanks @MustafaJafar 

I personally like it less because it's further away from where you click so I have a feeling it's less quick this way, usually. But I'm fine with merging it if everyone likes it.

Could you add a screen recording on how it behaves without this PR? That way anyone can just quickly comment on the visual differences on how it behaves.

MustafaJafar:
> Could you add a screen recording on how it behaves without this PR? That way anyone can just quickly comment on the visual differences on how it behaves.

I already have that, let me update the PR description.

moonyuet (approved):

BigRoy:
Had another look - I actually love that this removes the Product Type parm. Simplifies a lot of things.

However, I would love for the dialog to feel more like a dropdown menu so that it popups near the clicked location so that it behaves similar to what Houdini does or right click context menus do. I'll look into that.

MustafaJafar:
@BigRoy , I almost forgot about this PR 😄

BigRoy:
> @BigRoy , I almost forgot about this PR 😄

I've had it in my mind - just not at high priority... but looking at the fact that I'm now creating more HDAs.. maybe I do want to focus on this rather sooner than later. Hmm... hopefully later this week before my vacation.

MustafaJafar:
> > @BigRoy , I almost forgot about this PR 😄
> 
> I've had it in my mind - just not at high priority... but looking at the fact that I'm now creating more HDAs.. maybe I do want to focus on this rather sooner than later. Hmm... hopefully later this week before my vacation.

So, what is missing according to your comment https://github.com/ynput/ayon-houdini/pull/39#issuecomment-2225310932 above is to place the window near the mouse post. 
I'm a little noob. 
I was able to get the mouse location, but no idea how to set the location correctly.
```python
    def showEvent(self, event):
        # Position the window near the mouse position 
        pos = QtGui.QCursor.pos()
        geo = QtCore.QRect(pos.x(), pos.y())
        self.setGeometry(geo)

        return super(SelectFolderPathDialog, self).showEvent(event)
```

BigRoy:
@MustafaJafar I've pushed some changes could you give it a go and of course, break it if possible?

image

PR #40, opened 2024-07-12
Chore: Fix copy in create package

iLLiCiTiT:
## Description
Fix copy to output dir in create package script.

### DIFF
--- a/create_package.py
+++ b/create_package.py
@@ -132,10 +132,7 @@ def safe_copy_file(src_path: str, dst_path: str):
         return
 
     dst_dir: str = os.path.dirname(dst_path)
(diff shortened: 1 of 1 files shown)


BigRoy:
To test this I should just run it to create package at explicit output dir that does not exist before I run it?

BigRoy (approved):
This worked to create the package in that 'destination' folder even if the packages folder did not previously exist.
```
/ayon-houdini/create_package.py --output %DESTINATION%
```

Also worked with
```
/ayon-houdini/create_package.py
```
Even if `packages` folder did not exist yet in the `ayon-houdini` git folder.

PR #41, opened 2024-07-13
Creators: Remove legacy `pop` of the 'active' data on create of instance.

BigRoy:
## Changelog Description

Creators: Remove legacy `pop` of the 'active' data on create of instance.

## Additional info

This was a leftover from legacy creator where the active state was defined by the Bypass state of the node instead of a dedicated parm for the publisher.

⚠️ **Question:**

There also seems to be this [left-over Collect Active State plug-in](https://github.com/ynput/ayon-houdini/blob/a2285258185afb9db9337163204cb6b2e8b8f787/client/ayon_houdini/plugins/publish/collect_active_state.py#L27-L31). Which currently would be unable to activate any inactive instances as far as I know because new publisher would have those instances already deactivated from CreateContext. However, it will disable instances if the node were currently set to "bypass" but the instance was active in publisher. 

**Question:** Should we also remove the Collect Active State plug-in to avoid this unwanted side effect?

---

This fixes the default active state creation for https://github.com/ynput/ayon-core/pull/774 combined with https://github.com/ynput/ayon-houdini/pull/36

## Testing notes:

1. Creators should still work
2. The active state should be imprinted and collected the same way as before this PR.
3. Double check HDA creator and publishing works, because that one always has been special. :)

### DIFF
--- a/client/ayon_houdini/plugins/create/create_alembic_camera.py
+++ b/client/ayon_houdini/plugins/create/create_alembic_camera.py
@@ -17,7 +17,6 @@ class CreateAlembicCamera(plugin.HoudiniCreator):
     def create(self, product_name, instance_data, pre_create_data):
         import hou
 
-        instance_data.pop("active", None)
         instance_data.update({"node_type": "alembic"})
 
         instance = super(CreateAlembicCamera, self).create(
--- a/client/ayon_houdini/plugins/create/create_arnold_ass.py
+++ b/client/ayon_houdini/plugins/create/create_arnold_ass.py
@@ -20,7 +20,6 @@ class CreateArnoldAss(plugin.HoudiniCreator):
     def create(self, product_name, instance_data, pre_create_data):
         import hou
 
-        instance_data.pop("active", None)
         instance_data.update({"node_type": "arnold"})
         creator_attributes = instance_data.setdefault(
             "creator_attributes", dict())
--- a/client/ayon_houdini/plugins/create/create_arnold_rop.py
+++ b/client/ayon_houdini/plugins/create/create_arnold_rop.py
@@ -26,7 +26,6 @@ class CreateArnoldRop(plugin.HoudiniCreator):
                 creator_attributes[key] = pre_create_data[key]
 
         # Remove the active, we are checking the bypass flag of the nodes
-        instance_data.pop("active", None)
         instance_data.update({"node_type": "arnold"})
 
         # Add chunk size attribute
--- a/client/ayon_houdini/plugins/create/create_bgeo.py
+++ b/client/ayon_houdini/plugins/create/create_bgeo.py
@@ -15,8 +15,6 @@ class CreateBGEO(plugin.HoudiniCreator):
 
     def create(self, product_name, instance_data, pre_create_data):
 
-        instance_data.pop("active", None)
-
         instance_data.update({"node_type": "geometry"})
         creator_attributes = instance_data.setdefault(
             "creator_attributes", dict())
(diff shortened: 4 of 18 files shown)


moonyuet (approved):
Tested by creating new instances, the active data are still located in the extra attributes.
Looks good.
image
image

MustafaJafar (approved):
It works.

MustafaJafar:
> ⚠️ **Question:**
> 
> There also seems to be this [left-over Collect Active State plug-in](https://github.com/ynput/ayon-houdini/blob/a2285258185afb9db9337163204cb6b2e8b8f787/client/ayon_houdini/plugins/publish/collect_active_state.py#L27-L31). Which currently would be unable to activate any inactive instances as far as I know because new publisher would have those instances already deactivated from CreateContext. However, it will disable instances if the node were currently set to "bypass" but the instance was active in publisher.
> 
> **Question:** Should we also remove the Collect Active State plug-in to avoid this unwanted side effect?

What about adding these few lines [here](https://github.com/ynput/ayon-houdini/blob/a2285258185afb9db9337163204cb6b2e8b8f787/client/ayon_houdini/api/plugin.py#L251) and remove `Collect Active State` plug-in ?
 
```python
            node = hou.node(node_path)
            if hasattr(node, "isBypassed"):
                node_data["active"] = not node.isBypassed()
```

BigRoy:
> > ⚠️ **Question:**
> > There also seems to be this [left-over Collect Active State plug-in](https://github.com/ynput/ayon-houdini/blob/a2285258185afb9db9337163204cb6b2e8b8f787/client/ayon_houdini/plugins/publish/collect_active_state.py#L27-L31). Which currently would be unable to activate any inactive instances as far as I know because new publisher would have those instances already deactivated from CreateContext. However, it will disable instances if the node were currently set to "bypass" but the instance was active in publisher.
> > **Question:** Should we also remove the Collect Active State plug-in to avoid this unwanted side effect?
> 
> What about adding these few lines [here](https://github.com/ynput/ayon-houdini/blob/a2285258185afb9db9337163204cb6b2e8b8f787/client/ayon_houdini/api/plugin.py#L251) and remove `Collect Active State` plug-in ? (btw the it works on my side.)
> 
> ```python
>             node = hou.node(node_path)
>             if hasattr(node, "isBypassed") and node.isBypassed():
>                 node_data["active"] = False
> ```
> 
> Side note: I think your question is related to this issue #16

We could - but I believe the relying on the "bypass" state was removed on purpose to avoid interfering with a native houdini feature that artists may use for something else. @antirotor was back then involved with that decision-making.

So if we were to stick with that I'd say just remove that collector (and then of course still validate that nodes that are set to bypass do actually actively render the ROPs the way we want them to.) (they should still work in publishing then.)

MustafaJafar:
> We could - but I believe the relying on the "bypass" state was removed on purpose to avoid interfering with a native houdini feature that artists may use for something else. @antirotor was back then involved with that decision-making.
> 
> So if we were to stick with that I'd say just remove that collector (and then of course still validate that nodes that are set to bypass do actually actively render the ROPs the way we want them to.) (they should still work in publishing then.)

I'd appreciate shedding some light on these feature. 

The only thing I was able to think of is publishing existing files as Houdini will skip rendering without not raising any errors.

image

BigRoy:
> > We could - but I believe the relying on the "bypass" state was removed on purpose to avoid interfering with a native houdini feature that artists may use for something else. @antirotor was back then involved with that decision-making.
> > So if we were to stick with that I'd say just remove that collector (and then of course still validate that nodes that are set to bypass do actually actively render the ROPs the way we want them to.) (they should still work in publishing then.)
> 
> I'd appreciate shedding some light on these feature.
> 
> The only thing I was able to think of is publishing existing files as Houdini will skip rendering without not raising any errors. (btw, it was my first time to know that this actually works - I disabled validate bypass validator.) image

Thanks - we could make the Bypass validation optional and force the `RopNode.render` with the `ignore_bypass_flags=True` arguments but seems more counter-intuitive even. (Especially because that may hold even more untrue for what e.g. a deadline render farm may do with those ROPs if they were bypassed).

I'll merge this PR.

The Collect Active State collector being removed is in a separate PR #42

PR #42, opened 2024-07-15
Remove Collect Active State plug-in

BigRoy:
As discussed [here]() the Collect Active State plug-in was a left over from the old publish system that used the bypass state to activate/deactivate the instances in the publisher.

This currently, with the new publisher, does nothing. It still continues to process it, due to the issue explained [here](https://github.com/ynput/ayon-houdini/pull/42#issuecomment-2228300804) so the plug-in was basically already doing nothing.

Anyway, it's not an issue at all because even with it continuing processing we have a dedicated validator for the ROP's bypass state:

image

So it would still get caught.

### Additional Context

~~I'm actually quite confused @iLLiCiTiT that setting "False" here on the instance still means the instance gets processed by further plug-ins. It should've actually disabled it for futher processing. Are we unable to set the active state of the instance during the publishing?~~ Solved: see [comment](https://github.com/ynput/ayon-houdini/pull/42#issuecomment-2228300804)


iLLiCiTiT:
> I'm actually quite confused @iLLiCiTiT that setting "False" here on the instance still means the instance gets processed by further plug-ins. It should've actually disabled it for futher processing. Are we unable to set the active state of the instance during the publishing?

It should disable future processing of instance plugins, context plugins have to handle it on their own.

BigRoy:
> It should disable future processing of instance plugins, context plugins have to handle it on their own.

Hmm - testing it like this, you're right:
```python
import pyblish.api


class TestA(pyblish.api.InstancePlugin):
    order = pyblish.api.CollectorOrder + 0.1
    families = ["*"]
    label = "TEST: Deactivate all instances"

    def process(self, instance):
        instance.data.update({
            "active": False,
            "publish": False
        })


class TestB(pyblish.api.InstancePlugin):
    order = pyblish.api.CollectorOrder + 0.2
    families = ["*"]
    label = "TEST: Run over all instances"

    def process(self, instance):
        self.log.info(f"Processing {instance}...")
```

---

The plug-in didn't continue past these lines:
```python
        # Must have node to check for bypass state
        if len(instance) == 0:
            return
```
Because in current Houdini implementation it's not in `instance` anymore, but just in `instance.data["instance_node"]`.
That solves that!

So it's working as expected.

Still this plug-in can be removed just fine 👍 (especially because it didn't do anything apparently anyway)

iLLiCiTiT (approved):

PR #45, opened 2024-07-16
Overwrite the last publish

### LINKED ISSUE #18: Overwrite the last publish

### Current Behavior:

Currently, there is no overwrite to the latest version in Ayon. When we publish from Houdini, it becomes a pain when large caches need to be saved.
image
image

### Expected Behavior:

Can have a feature to overwrite the latest version cache?

### Version

1.0.0

Windows

### Steps To Reproduce:

When general BGEO.sc cache was publishing, we can see a option to overwrite the latest cache in ayon publish options

### Relevant log output:

_No response_

### Additional context:

_No response_

[cuID:[AY-5773](https://app.clickup.com/t/86bz5dbdz)]

MustafaJafar:
## Changelog Description
Rendering is expensive, sometimes it is helpful only to re-render changed frames and reuse existing.

Todo list:
- [x] **ExtractLastPublished**: A generic extractor to copy the files of the latest published version `last_version_published_files` .
It only copies files that do not need to be fixed.
- [x] Allow `render_rop` to override ROP's frame range
- [x] Support frames to fix for different product types
  - [x] Cache product types
  - [x] Review (OpenGl) 
  - [x] Render
- [x] ~~Test on farm~~ The entire logic is local. it won't run on farm.
- [x] ~~Add profiles to `CollectFramesFixDef` settings (Here is it https://github.com/ynput/ayon-core/pull/839)~~
- [x] Copy `CollectFramesFixDef` from `ayon-core`, more info check https://github.com/ynput/ayon-core/pull/839#issuecomment-2293777570
- [x] Add settings for `CollectFramesFixDef`.
- [x] Update frame range so that integrator include frames to fix

> [!NOTE]
> This PR is the Houdini equivalent for https://github.com/ynput/OpenPype/pull/4362 although in Houdini it can be a little more complex.
>
> ---
> **Some user notes:**
> 1. when extending frame range. you need to mark them as frame to fix.
> 2. Frames to fix if output is a single file with no frame. it'll be ignored.
> 3. When publishing render with multiple separate AOVs while selecting `Local machine rendering`, please don't use `frames_to_fix`, you'll mostly get `missing frames` error. 
> 4. When publishing render with multiple separate AOVs while selecting `Use existing frames (local`) , it'll be the artist responsibility to fill in any missing frames. Note,` ExtractLastPublished` will still work. so if you rendered the entire frame range, then `ExtractLastPublished` will overwrite any existing franes according to the `frames_to_fix` value.

## Testing notes:
1. In Houdini create a product
2. In publisher tab, set frames to fix to your desired value
3. publish. check "Extract ROP" and "Extract Last Published" log.


MustafaJafar:
Some screenshots form publishing a Bgeo pointcache.
image
image

Cleaning up log
image

MustafaJafar:
Currently, This PR works with local publishes properly.
Few notes:
1. When publishing render with multiple separate AOVs while selecting `Local machine rendering`, please don't use frames_to_fix, you'll mostly get missing frames error.
  More info:
  ```
  # Render the entire frame range if any of the AOVs does not have a
  # previously rendered version. This situation breaks the publishing.
  # because There will be missing frames as ROP nodes typically cannot render different
  #  frame ranges for each AOV; they always use the same frame range for all AOVs.
  ```
2. When publishing render with multiple separate AOVs while selecting `Use existing frames (local)` , it'll be the artist responsibility to fill in any missing frames. Note, `ExtractLastPublished`  will still work. so if you rendered the entire frame range, then `ExtractLastPublished` will overwrite them according to the `frames_to_fix` value.

---
The situation with farm submission: it doesn't work on farm.
1. The target of [CollectFramesFixDef](https://github.com/ynput/ayon-core/blob/3b53093d12f2fc181858f3815677fd23ca7cb49a/client/ayon_core/plugins/publish/collect_frames_fix.py#L23) plugin is set to `local`, so it won't work on farm. (this is the collector responsible for processing the `frames_to_fix` user input and listing the files of the last rendered version.)
2. It's possible to copy  the files of the last rendered version before submitting the job to farm **and propagate the frames to fix to the deadline job submission.** which should effectively render only the frames to fix and publish. but, this won't work with render products. because the logic in `ExtractLastPublished` expects the AOV instances to exist which only happens with local rendering. 

What do you recommend ?

moonyuet:
> Currently, This PR works with local publishes properly. Few notes:
> 
> 1. When publishing render with multiple separate AOVs while selecting `Local machine rendering`, please don't use frames_to_fix, you'll mostly get missing frames error.
>    More info:
> 
> ```
> # Render the entire frame range if any of the AOVs does not have a
> # previously rendered version. This situation breaks the publishing.
> # because There will be missing frames as ROP nodes typically cannot render different
> #  frame ranges for each AOV; they always use the same frame range for all AOVs.
> ```
> 
> 2. When publishing render with multiple separate AOVs while selecting `Use existing frames (local)` , it'll be the artist responsibility to fill in any missing frames. Note, `ExtractLastPublished`  will still work. so if you rendered the entire frame range, then `ExtractLastPublished` will overwrite them according to the `frames_to_fix` value.
> 
> The situation with farm submission: it doesn't work on farm.
> 
> 1. The target of [CollectFramesFixDef](https://github.com/ynput/ayon-core/blob/3b53093d12f2fc181858f3815677fd23ca7cb49a/client/ayon_core/plugins/publish/collect_frames_fix.py#L23) plugin is set to `local`, so it won't work on farm. (this is the collector responsible for processing the `frames_to_fix` user input and listing the files of the last rendered version.)
> 2. It's possible to copy  the files of the last rendered version before submitting the job to farm **and propagate the frames to fix to the deadline job submission.** which should effectively render only the frames to fix and publish. but, this won't work with render products. because the logic in `ExtractLastPublished` expects the AOV instances to exist which only happens with local rendering.
> 
> What do you recommend ?

I guess it is not supposed to work on farm when there are already rendered frames in the renders folder(aka skipping to submit to farm). I will test it to see how it goes. (Should be working smiliarly to Nuke.)

moonyuet:
When testing the PR, it is pretty concise for the image-rendering related product type(as it has render target options)
image
But it is quite confusing for me to test on the review and cache-related product type as it does not give any info whether I am using general local rendering(render locally and publish) or using existing frames (use existing frames for publishing). 
Guess it is better to have render target options similar to the render host to identify which one to use for publishing (For cache: local, local(existing frames) and farm; For review: local, local(existing frames)), if it doesn't cause any anti-pattern for the implementation of the generic node by @BigRoy

client/ayon_houdini/plugins/publish/collect_frames_fix.py
```diff
@@ -0,0 +1,101 @@
...
+            #   representation 'context' data.
+            repre_context = repre["context"]
+            product_type = repre_context.get("product", {}).get("type")
+            if not product_type:
+                product_type = repre_context.get("family")
+            if "*" not in self.families and product_type not in self.families:
+                continue
+
+            for file_info in repre.get("files"):
+                published_files.append(file_info["path"])
```
BigRoy, line 73: This seems fishy - isn't the product type defined on the product or the version? Getting it from the `context` of the representation seems a bit odd since that context is just the context for the template data?

client/ayon_houdini/api/lib.py
```diff
@@ -146,11 +146,26 @@ def validate_fps():
     return True
 
 
-def render_rop(ropnode):
+def render_rop(ropnode, frame_range=None):
```
BigRoy, line 149: Why don't we just match the defaults of houdini?
```suggestion
def render_rop(ropnode, frame_range=tuple()):
```

client/ayon_houdini/plugins/publish/extract_last_published.py
```diff
@@ -0,0 +1,89 @@
...
+        frames_to_fix = clique.parse(frames_to_fix, "{ranges}")
+        
+        anatomy = instance.context.data["anatomy"]
+
+        for file_path, frame in last_published_and_frames.items():
+            if frame is None:
+                continue
+            file_path = anatomy.fill_root(file_path)
+            if not os.path.exists(file_path):
+                continue
```
BigRoy, line 79: So if a source frame just happens to not exist we ignore it? Seems like something that we should error on?

client/ayon_houdini/plugins/publish/extract_last_published.py
```diff
@@ -0,0 +1,89 @@
...
+            target_file_name = frames_and_expected.get(frame)
+            if not target_file_name:
+                continue
```
BigRoy, line 82: Same here... if we can't define where it ends up we ignore it silently? Does that seem right? It's collected into `last_version_published_files` data because it was data we needed to transfer, right? Or did it just contain all last version's files and we're filtering here.

I have a feeling we need a generic way (that hopefully also is not houdini-specific!) to:

- Define the frames we want to re-render/re-export in the current publish -> The frames to render
- Those not include but still within frame start and frame end of the instance are transferred from last version -> The list we need to copy from last version.
- In validation we need to ensure the frames to copy from last version actually exist (so it actually makes sense to try and ignore those and we're safe to continue?)

client/ayon_houdini/plugins/publish/collect_frames_fix.py
MustafaJafar: I wish if I was able to tell. I desperately copied it.
For reference,  https://github.com/ynput/ayon-core/pull/839#issuecomment-2293777570

client/ayon_houdini/plugins/publish/extract_last_published.py
MustafaJafar: Not actually, because we can use this feature to extend the frame range of a the last published render. 
In that case, the frame won't exist.

But, maybe we need to check if that frame is included in the frames to fix. IDK.
I was not able to make it error when I tested many scenarios. e.g. extend frames, fix frames.

client/ayon_houdini/plugins/publish/extract_last_published.py
MustafaJafar: I agree, I tried to make it generic although I was mostly stealing code from Nuke.

MustafaJafar: What is the case when those files don't exist ?
- wrong `frames to fix` input ? 
- Or  the published files don't exist ? i.e. they were deleted manually without removing them from the database ? 
- Or a storage connection error resulting in making the files look like they are missed ? which I assume will result in an error anyway during the publishing.

client/ayon_houdini/api/lib.py
MustafaJafar: yeah, correct. I forgot that tuples are immutable so this won't make us fall in the pass by reference trap. 
Many thanks... https://github.com/ynput/ayon-houdini/pull/45/commits/61c2fae3f61405c70b31f814ea8931e3f71ce34e

client/ayon_houdini/plugins/publish/extract_last_published.py
BigRoy: So - what defines the final published frame range for the new version (or what we're writing now?)

Is it, the frame start and frame end specified on the instance? Is it only the frames to fix? Is it always whatever existed in last publish plus the current frames to fix? But what if the frames to fix is e.g. 100-120 and previous publish was 80-90 - do we allow the gap? (Pretty sure AYON doesn't like that currently on published sequences?)

client/ayon_houdini/plugins/publish/collect_frames_fix.py
moonyuet: @BigRoy This collector is inherited from the code which lives in the ayon-core where it is used for collecting frames fix in Nuke, which also has the same functionality of overwriting the last publish. You can find this PR helpful: https://github.com/ynput/ayon-core/pull/839

client/ayon_houdini/plugins/publish/collect_frames_fix.py
```diff
@@ -0,0 +1,101 @@
...
+
+        instance.data["last_version_published_files"] = published_files
+        self.log.debug("last_version_published_files::{}".format(
```
moonyuet, line 76:
```suggestion
        self.log.debug("last_version_published_files: {}".format(
```

BigRoy:
@MustafaJafar just checking back in.

- What's the state of this PR? It has some open todo checkboxes - when do you need e.g. more of my review?
- How much does it match 'other' DCC implementations? (Like how good can we make this minimal code related to Houdini and allow the functionality to be shared between host integrations?)

MustafaJafar:
Hey, 
I refreshed my mind a bit regarding this PR. 
Since most of Houdini exports are sequences of files. so, it was considered since the beginning to make this PR work with as many AYON Houdini Products.

> @MustafaJafar just checking back in.
> 
> * What's the state of this PR? It has some open todo checkboxes - when do you need e.g. more of my review?

Currently, there's a bug regards the frame range of the instances as `frame_to_fix` is not considered.
We had a discussion about it but we didn't come up with a conclusion. 
As far as I know, it should be frame range + frame to fix. and we should add a validator to expose any frame gaps. 


> * How much does it match 'other' DCC implementations? (Like how good can we make this minimal code related to Houdini and allow the functionality to be shared between host integrations?)

As far as I know, this is only implemented in Nuke. and I've no idea how to make it generic.
The collector `CollectFramesFixDef` is indeed generic and it was copied as we couldn't come up with a decision check https://github.com/ynput/ayon-core/pull/839#issuecomment-2293777570

The extractor `ExtractLastPublished` is mostly generic but I had add some logic related to Houdini to get the list `expected_filenames` and staging directory.

MustafaJafar:
I eventually though about adding a validator that:
1. Check if the frames to fix include any gaps.
2. Check if the Rop frame range is not correct.
3. Offer a help to update the scene frame range. (It may be confusing. as it doesn't touch the ROP nodes and doesn't help in most of the time. But, I think it's good to have it to initiate a discussion about what would be a good help action).
4. I don't like to optimize and beautify the code in the validator, but couldn't come up with better ideas atm.

image
image

MustafaJafar:
I've finished all the TODOs. and I think this function is working (at least as a initial support and we can enhance it over time).
By defaults it work with all families in Houdini. (since most of the AYON Houdini products are sequence of files and the plugin will/should take care if the product is one frame or one file with no frames.)

Please note, this feature only work locally.

Currently, for Artists to specify frames to fix, they need to have a frame range in their ROPs that covers the frames to fix.
The feature should copy the frames from the last publish before rendering. 
when rendering, the frames to fix value overrides the ROP frame range leading it to render only the specified range.
_I think I don't cove an edge case (if a frame in the last publish is missing, I should tell the ROP node to render it).. it might not be a much problem because we check for missing frames after rendering._

The ground truth for publishing and integration is the frame range of the ROP node.

I've no idea how to make this feature generic. since different addons have different ways for setting frame range in their pyblish workflow. also, we may use different keys for files and frames. however, we ensure achieving a consistent representation data for the integrator.

client/ayon_houdini/plugins/publish/extract_last_published.py
```diff
@@ -0,0 +1,99 @@
...
+class ExtractLastPublished(plugin.HoudiniExtractorPlugin):
+    """
+    Generic Extractor that copies files from last published
+    to staging directory.
+    It works only if instance data includes "last_version_published_files"
+    and there are frames to fix.
+
+    The files from last published are base of files which will be extended/fixed for specific
+    frames.
+    """
```
BigRoy, line 20: I don't get it. If this is _generic_? How come `ayon-core` does not have this yet... but it does have `CollectFrameToFix` plug-in? Shouldn't there be some other implementation of this then somewhere that does _exactly_ this?

EDIT: AH, it's from Nuke [here](https://github.com/ynput/ayon-nuke/blob/43726e8ff8ca7a57f9169995b03447a413af8658/client/ayon_nuke/plugins/publish/extract_render_local.py#L70-L218).

This should come with a massive `# TODO` block screaming that this is taken 1:1 from `ayon-nuke` and should transition to `ayon-core`. Or we transition that now I'd say.

client/ayon_houdini/plugins/publish/extract_last_published.py
BigRoy: > I agree, I tried to make it generic although I was mostly stealing code from Nuke.

When doing such a copy - it's good to mark that very clear in the code... especially with such a feature that should really be taken OUT of Nuke and into core... Because this plug-in should be generic, not to Houdini, not to Nuke, but something that can all share.

client/ayon_houdini/plugins/publish/extract_render.py
```diff
@@ -58,8 +59,29 @@ def process(self, instance):
...
-            ropnode = hou.node(instance.data.get("instance_node"))
-            render_rop(ropnode)
+            # FIXME Render the entire frame range if any of the AOVs does not have a
+            # previously rendered version. This situation breaks the publishing.
+            # because There will be missing frames as ROP nodes typically cannot render different
+            #  frame ranges for each AOV; they always use the same frame range for all AOVs.
+            rop_node = hou.node(instance.data.get("instance_node"))
+            frames_to_fix = clique.parse(instance.data.get("frames_to_fix", ""), "{ranges}")
+
+            if len(set(frames_to_fix)) > 1:
```
BigRoy, line 69: Why would this only be for two frames and more?
```suggestion
            if len(set(frames_to_fix)) > 0:
```

client/ayon_houdini/plugins/publish/extract_rop.py
```diff
@@ -33,9 +34,31 @@ def process(self, instance: pyblish.api.Instance):
...
+                self.log.debug(
+                    "Rendering frames to fix [{f1}, {f2}]".format(
+                        f1=frame_range[0],
+                        f2=frame_range[-1]
+                    )
+                )
+                frame_range = (
+                    int(frame_range[0]), int(frame_range[-1]), instance.data["byFrameStep"]
+                )
+                render_rop(rop_node, frame_range=frame_range)
```
BigRoy, line 58: This logic feels incorrect. If we do a `byFrameStep` for each separate chunk. Aren't we stepping from the wrong start frame and hence offsetted in each chunk.

Also, we may as well off load this to the `render_rop` function so we don't need to duplicate the logic.

So that we just do `render_rop(rop_nopes, frames_to_render)`?

client/ayon_houdini/plugins/publish/validate_frame_range_frames_to_fix.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+class UpdateSceneFrameRangeAction(RepairAction):
+    label = "Update scene frame range by frames to fix"
+
+
+class ValidateFrameRangeFramesToFix(plugin.HoudiniInstancePlugin):
+    """Validate Frame Range Frames to Fix.
+
+    This validator checks if the rop node covers the entire frame range, including any frames that require correction.
```
BigRoy, line 20: Cosmetics. Exceeds max line length?

client/ayon_houdini/plugins/publish/validate_frame_range_frames_to_fix.py
```diff
@@ -0,0 +1,118 @@
...
+        if not frames_to_fix.is_contiguous():
+            title = "Invalid Frames To Fix"
+            message = (
+                "Frames to fix [{0[frames_to_fix]}] include gaps.\n"
+                "Please, use a value that has no gaps."
+                .format(instance.data)
+            )
+
+            cls.log.error("Frames to fix include gaps.")
+            return (rop_node.path(), title, message)
```
BigRoy, line 70: Sorry - why can't we fix `1001-1002,1005-1010`?

client/ayon_houdini/plugins/publish/validate_frame_range_frames_to_fix.py
```diff
@@ -0,0 +1,118 @@
...
+        if not frame_end >= int(frame_range[-1]):
+            frame_end = frame_range[-1]
+
+        frame_start = int(frame_start)
+        frame_end = int(frame_end)
+
+        # Set frame range
+        hou.playbar.setFrameRange(frame_start, frame_end)
+        hou.playbar.setPlaybackRange(frame_start, frame_end)
+        hou.setFrame(frame_start)
```
BigRoy, line 118: Sorry - I don't understand? Why would I want to update my scene frame range to the frames to fix?

client/ayon_houdini/plugins/publish/extract_last_published.py
MustafaJafar: > transition that now I'd say.

I've no problem doing that but I'm afraid we fell in the same situation as https://github.com/ynput/ayon-core/pull/839#issuecomment-2293777570

client/ayon_houdini/plugins/publish/collect_frames_fix.py
```diff
@@ -0,0 +1,101 @@
...
+from ayon_core.lib.attribute_definitions import (
+    TextDef,
+    BoolDef
+)
+from ayon_core.pipeline.publish import AYONPyblishPluginMixin
+
+from ayon_houdini.api import plugin
+
+
+class CollectFramesFixDef(
```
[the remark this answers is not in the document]
MustafaJafar: https://github.com/ynput/ayon-houdini/pull/45/commits/224723d58fd06b34a40b3b65b5425a81fd44853a

client/ayon_houdini/plugins/publish/extract_render.py
MustafaJafar: This avoids publish instances with one frame/file. e.g. HDA products, alembic products.

client/ayon_houdini/plugins/publish/extract_rop.py
MustafaJafar: That's a good idea.  but, it'll be frames to fix which is a string containing the frame range `"1001-1003"`.
`render_rop(rop_nopes, frames_to_fix)` .. I may need some time to think about it. because that string doesn't contain any info about the frame step.

> This logic feels incorrect. If we do a `byFrameStep` for each separate chunk. Aren't we stepping from the wrong start frame and hence offsetted in each chunk.

I missed this 😢 Any ideas about considering this ?

client/ayon_houdini/plugins/publish/validate_frame_range_frames_to_fix.py
MustafaJafar: https://github.com/ynput/ayon-houdini/pull/45/commits/5534e3eb5e3f3b6691309393e51be1c401f85069

client/ayon_houdini/plugins/publish/validate_frame_range_frames_to_fix.py
MustafaJafar: tbh, I want to remove it. but, I kept to initiate a discussion about if should we provide any fixes ?
Most of the ROP nodes are following the scene frame range.

client/ayon_houdini/plugins/publish/validate_frame_range_frames_to_fix.py
MustafaJafar: oh, I mistaken that by the comment above https://github.com/ynput/ayon-houdini/pull/45#discussion_r1724006751
I think since we are validating that ROP's frame range covers the frames to fix, then we can just fix `1001-1002,1005-1010` fine

MustafaJafar: fixed here https://github.com/ynput/ayon-houdini/pull/45/commits/9161e09fe93bf1dc85ea6b6c8cab00d5506d8a80

MustafaJafar: Marking this as unresolved again. 
We should be able to fix `1001-1002,1005-1010` just fine. but a new edge case popped up. 

What happens if the last published has frames `1001-1002` and my frames to fix are `1003-1005, 1007`
- `1001-1002` are copied. 
- `1003-1005` and `1007-1007` will be rendered. 
- then we will have missing frame `1006`.
- extract render will error because of the that missing frame without giving a hint that's happened because of the gap in the frame range.. 

I think to solve this we need to get the number of frames in the last publish + the number of frames to fix. 
and compare that to the ROP frame range to expose any potential missed frames. 
This is also the same case you mentioned earlier https://github.com/ynput/ayon-houdini/pull/45#discussion_r1723580252

client/ayon_houdini/plugins/publish/validate_frame_range_frames_to_fix.py
BigRoy: Let's remove it - or override the actual frame parm (which I'm also not sure of). Let's move repair for now.

MustafaJafar: removed.
https://github.com/ynput/ayon-houdini/pull/45/commits/38720902d6e19cd0e8bde8bad1176cde1d853f27

MustafaJafar:
Hey, let me sum up the previous conversations: 

This PR has some caveats: 
1. `render_rop` may step from the wrong frame.
2. when extending the frame range while having gaps in frames to fix. (the publish will fail because of missing frames) -> [example](https://github.com/ynput/ayon-houdini/pull/45#discussion_r1743334789)
3. Nothing in extract last publish error if the last published file doesn't exist. and we may need to add more logging. 

we need to keep in mind that at some point, we would need to make it generic to work with different Host Addons. this includes the collector and extractor of frames to fix. which typically compute the frame to fix, copy that last published.

client/ayon_houdini/plugins/publish/extract_rop.py
BigRoy: The solution to this @MustafaJafar - is whenever you specify `frames_to_fix` to FORCE `byFrameStep` to 1.0. If you still want to "step frames" for frames to fix, you should then do, e.g. `1-100x5` (does clique parse support that?)

MustafaJafar: > `1-100x5` (does clique parse support that?)

No.

MustafaJafar: https://github.com/ynput/ayon-houdini/pull/45/commits/6be6494e0ccca718e21c839d5d380d403d32dceb

client/ayon_houdini/plugins/publish/extract_last_published.py
```diff
@@ -0,0 +1,107 @@
...
+            if frame and frame not in frames_to_fix:
+                self.log.debug("Copying '{}' -> '{}'".format(file_path, out_path))
+                shutil.copy(file_path, out_path)
+
+
+    def get_expected_files_and_staging_dir(self, instance):
+        """ Get expected file names or frames.
+
+        This method includes Houdini specific code.
+        """
```
BigRoy, line 83: Can we get arguments and return with types in the docstring?

MustafaJafar: Thank you. Here it's.
https://github.com/ynput/ayon-houdini/pull/45/commits/8cef57a3c46eb3277b1dd5d9542aea3587eb055b

MustafaJafar:
Hey, 
I didn't go through the new commits yet. 
Anyways, Here's a test run from Mantra Render. _This test included publishing 2 more AOVs but, I didn't include them in the log to avoid lengthy logs._

Log from the publisher. 
- **Frames to Fix:** `1002-1004`
- **Mantra ROP Frame Range:** `1001-1005`
- **Extract Last Published:**
    ```
    DEBUG: Copying '\\storage\work\ayon_projects/Robo/Assets/Character/roboangry/publish/render/render_beauty/v032/robo_roboangry_render_beauty_v032.1001.exr' -> '//storage/work/ayon_projects/Robo/Assets/Character/roboangry/work/cfx/pyblish/renders/mantra_ropTest/mantra_ropTest.1001.beauty.exr'
    DEBUG: Copying '\\storage\work\ayon_projects/Robo/Assets/Character/roboangry/publish/render/render_beauty/v032/robo_roboangry_render_beauty_v032.1005.exr' -> '//storage/work/ayon_projects/Robo/Assets/Character/roboangry/work/cfx/pyblish/renders/mantra_ropTest/mantra_ropTest.1005.beauty.exr'
    ```
- **Extract Render:**
    ```
    DEBUG: Rendering /out/mantra_ropTest
    DEBUG: Rendering frames to fix [1002, 1004]
    ```
- **Removed Files:**
    ```
    \\storage\work\ayon_projects\Robo\Assets\Character\roboangry\work\cfx\pyblish\renders\mantra_ropTest\mantra_ropTest.1001.beauty.exr
    \\storage\work\ayon_projects\Robo\Assets\Character\roboangry\work\cfx\pyblish\renders\mantra_ropTest\mantra_ropTest.1002.beauty.exr
    \\storage\work\ayon_projects\Robo\Assets\Character\roboangry\work\cfx\pyblish\renders\mantra_ropTest\mantra_ropTest.1003.beauty.exr
    \\storage\work\ayon_projects\Robo\Assets\Character\roboangry\work\cfx\pyblish\renders\mantra_ropTest\mantra_ropTest.1004.beauty.exr
    \\storage\work\ayon_projects\Robo\Assets\Character\roboangry\work\cfx\pyblish\renders\mantra_ropTest\mantra_ropTest.1005.beauty.exr
    \\storage\work\ayon_projects\Robo\Assets\Character\roboangry\work\cfx\pyblish\renders\mantra_ropTest\mantra_ropTest_h264focal_length_burnin.mp4
    ```

Also, the published files look fine.

client/ayon_houdini/plugins/publish/extract_last_published.py
```diff
@@ -0,0 +1,119 @@
...
+    label = "Extract Last Published"
+    targets = ["local"]  # Same target as `CollectFramesFixDef`
+    families = ["*"]
+
+    def process(self, instance):
+        frames_to_fix = instance.data.get("frames_to_fix")
+        if not frames_to_fix:
+            self.log.debug("Skipping, No frames to fix.")
+            return
+
```
MustafaJafar, line 40: I wonder if should we add this one here too.
```suggestion
        if not instance.data.get("integrate", True):
            self.log.debug("Skipping collecting frames to fix data for "
                           "instance because instance is set to not integrate")
            return
            
```

moonyuet: I guess yes as you are not gonna extract the files and transfer them to the publish folder.

MustafaJafar: well, yes. 
But it's also worth mentioning that instances with `integrate == False` don't usually have any last published files so I always get 
```
Skipping, No last publish found.
```

MustafaJafar:
I gave it one more test local and farm render. 
After merging PR https://github.com/ynput/ayon-houdini/pull/93 
I was able to test this PR and getting the same results mentioned in https://github.com/ynput/ayon-houdini/pull/45#issuecomment-2340035905 without modifying anything in the code manually.

on farm, we don't propagate the `frames_to_fix` also, we don't set render job frames to the `frames_to_fix`. 
Therefore, the render job will render the whole range. in my case it's `1001-1005`. while `frames_to_fix` was `1002-1004` .
image

dee-ynput:
> Also, we are keeping in mind that at some point, we would need to make it generic to work with different Host Addons. this includes the collector and extractor of frames to fix. which typically compute the frame to fix, copy that last published.

Here is an issue on ayon-core to explore a generic implementation:
https://github.com/ynput/ayon-core/issues/879

client/ayon_houdini/api/lib.py
```diff
@@ -149,11 +149,23 @@ def validate_fps():
     return True
 
 
-def render_rop(ropnode):
+def render_rop(ropnode, frame_range=tuple()):
```
iLLiCiTiT, line 152: Could this have default value `None`?

client/ayon_houdini/plugins/publish/extract_last_published.py
MustafaJafar: https://github.com/ynput/ayon-houdini/pull/45/commits/c96e05415d6f939bfc8f2abad9b9ccdb6d1b254f

client/ayon_houdini/api/lib.py
MustafaJafar: https://github.com/ynput/ayon-houdini/pull/45/commits/e8922f236c4e267622cd444ca455e0ad2d6bbe6b

client/ayon_houdini/api/plugin.py
```diff
@@ -354,3 +355,36 @@ class HoudiniExtractorPlugin(publish.Extractor):
...
+                        f2=frame_range[-1]
+                    )
+                )
+                # for step to be 1 since clique doesn't support steps.
+                frame_range = (
+                    int(frame_range[0]), int(frame_range[-1]), 1
+                )
+                render_rop(rop_node, frame_range=frame_range)
+        else:
+            render_rop(rop_node)
```
iLLiCiTiT, line 390: Why is everybody making huge if statements?
```suggestion
        if len(set(frames_to_fix)) < 2:
            render_rop(rop_node)
            return

        # Render only frames to fix
        for frame_range in frames_to_fix.separate():
            frame_range = list(frame_range)
            frame_range.sort()
            first_frame = int(frame_range[0])
            last_frame = int(frame_range[-1])
            self.log.debug(
                f"Rendering frames to fix [{first_frame}, {last_frame}]"
            )
            # for step to be 1 since clique doesn't support steps.
            frame_range = (first_frame, last_frame, 1)
            render_rop(rop_node, frame_range=frame_range)
```

client/ayon_houdini/api/plugin.py
```diff
@@ -354,3 +355,35 @@ class HoudiniExtractorPlugin(publish.Extractor):
...
+        if len(set(frames_to_fix)) < 2:
+            render_rop(rop_node)
+            return
+
+        # Render only frames to fix
+        for frame_range in frames_to_fix.separate():
+            frame_range = list(frame_range)
+            frame_range.sort()
+            first_frame = int(frame_range[0])
+            last_frame = int(frame_range[-1])
```
iLLiCiTiT, line 383: I just realized that there are `int` conversions, so original `frame_range` might not contain `int`? If yes, then the sort would not work.
```suggestion
            frame_range = [int(frame) for frame in frame_range]
            frame_range.sort()
            first_frame, last_frame = frame_range[0], frame_range[-1]
```

MustafaJafar: I don't think we need to sort since `frame_range` collection is ordered and converting to list won't affect that, no?

```suggestion
            frame_range = list(frame_range)
            first_frame = int(frame_range[0])
            last_frame = int(frame_range[-1])
```

> so original `frame_range` might not contain `int`?

I don't know tbh what would be the padding used in clique when using `{ranges}` keyword without specifying `padding`

iLLiCiTiT: > I don't think we need to sort since frame_range collection is ordered and converting

If that's the case, then your suggestion is correct.

iLLiCiTiT (approved):
Code looks ok, needs workflow validation and test.

client/ayon_houdini/api/lib.py
BigRoy: It could, it had before... but this matches because Houdini's default value and hence avoids the need to do `if frame_range is None: frame_range = tuple()` inside the function.

client/ayon_houdini/api/plugin.py
BigRoy: I understand the thinking - but this to me really feels less clearly that there are two options and it's one of the two.

Also, the sort is not needed because `clique.Collection` when put into `list` should be sorted by design. Right?

server/settings/publish.py
```diff
@@ -159,6 +175,15 @@ class PublishPluginsModel(BaseSettingsModel):
         "active": True,
         "families" : []
     },
+    "CollectFramesFixDefHou": {
+        "enabled": True,
+        "optional": True,
```
BigRoy, line 180: @MustafaJafar does it makes sense to have this optional? :) The plug-in itself doesn't support optional state yet but that'd be easy to implement.

BigRoy (approved):
🐌 Sorry for taking so long on my end each test-run. But I'd say this is a good MVP.

The caveats are:
- There may be unexpected results if say you'd "fix frames" to last version if your new render differs in resolution. ;)
- It **only works for local renders** not on the farm.
- We should work on abstracting this logic into core so that the logic isn't duplicated across different hosts each time.
- We should optimize which frames we copy from the previous version instead of copying all (which is also listed as todo in the code)
- We may want to make the optional "rewrite_version" toggle in the publisher UI a separate plug-in in the future in core (if that makes any sense) which would then just be a simple core functionality that allows any publish instance to "replace" an existing version regardless of whether it only fixes frames or not.

PR #48, opened 2024-07-17
Houdini: Activate host color management by default

BigRoy:
## Changelog Description

Active Host Color Management for Houdini by default.

## Additional info

This value is only relevant if in ayon core you have Global Color Management enabled. However, Houdini was the only application that had an application specific `activate_host_color_management` defaulting to False - even though others all have it True.

This meant that by default if you enabled Global Color Management it would still remain disabled for Houdini unless this toggle was checked. That's straight up confusing.

This may need to come with a minor bump and clear statement in the release notes - because they may be edge cases where a studio can get bit by this if they relied on the previous disabled state but had Global Color Management enabled.

## Testing notes

1. New default value should work

### DIFF
--- a/server/settings/imageio.py
+++ b/server/settings/imageio.py
@@ -96,7 +96,7 @@ class HoudiniImageIOModel(BaseSettingsModel):
 
 
 DEFAULT_IMAGEIO_SETTINGS = {
-    "activate_host_color_management": False,
+    "activate_host_color_management": True,
     "ocio_config": {
         "override_global_config": False,
         "filepath": []


MustafaJafar (approved):
My Approve == It works as excepted on my end.
But, I can't help on decision making. because I'm not aware of what studios are used to do. Maybe we can summon @m-u-r-p-h-y  

> This may need to come with a minor bump and clear statement in the release notes - because they may be edge cases where a studio can get bit by this if they relied on the previous disabled state but had Global Color Management enabled.

If they are copying settings values from the existent Houdini addon on their end, would it override the `activate_host_color_management` value ? 
No, it won't if they kept it off by default. Here's a screenshot where It's not visible.
image

Maybe we can just pass the pain to the backend repo as the copy settings tool doesn't inform users about the change in default settings. 
256dd183e023748f24f3cb6e044c1f96

PR #56, opened 2024-07-29
Add Load Shot LOP and Mute Layers LOP

BigRoy:
## Changelog Description

This adds two new HDAs/OTLs for Houdini for the USD workflow:

- **Load Shot LOP**: This is an equivalent to a Houdini "Sublayer" node but exposing the loader functionality in a similar manner as the Load Asset LOP.
- **Mute Layers LOP**: This is a custom HDA that allows to mute layers by wildcard, which can make it easy to mute e.g. all layers with `*modeling*` in the name, to - in that way, hopefully make it trivial to mute the layers you don't want active in your current scene - e.g. the department layer you're currently working on.

## Additional info

The Load Shot LOP makes it easy to load e.g. the current shot:
image

The Mute Layers LOP has some 'easy access' presets you can apply to mute specific layers:
image

## Testing notes:

1. Load Shot LOP and Mute Layers LOP should work and be understandable to use.
2. Both should provide a simple yet understandable "HELP" (question mark on the node)


BigRoy:
The layer muting is unfortunately based on wildcards to filename patterns. This means that also the "example defaults" are somewhat based on:
- Publish file templates (since it checks those files for the names)
- Also, those examples may not be the actual 'custom' layers a studio might define in settings: `ayon+settings://core/publish/CollectUSDLayerContributions/contribution_layers`

We may want to:
- somehow allow the defaults to be based on the defined layer names in settings 
- tweak the mute layer matching to be more reliable than solely using the filepath (yet at the same time for loaded nested filepaths that's about as much as we 'have' to be reliable on)
- be able to mute the 'current department layer' of you current task automatically: **this I'd like most!** so that we can have just a "mute current task layer" on the Load Shot node directly. However the department layers aren't "linked" to a particular task name or task type by default. To be able to do that we would need to solve https://github.com/ynput/ayon-core/issues/803

client/ayon_houdini/startup/otls/ayon_lop_load_shot.hda/ayon_8_8Lop_1load__shot_8_81.0/OnCreated
```diff
@@ -0,0 +1,6 @@
+node = kwargs["node"]
+hda_module = node.hdaModule()
+hda_module.setup_flag_changed_callback(node)
+
+node.parm("product_type").lock(True)
```
MustafaJafar, line 5:
```suggestion
```

MustafaJafar (commented):
Could we add the ayon icon in the `AYON Mute Layers` , the help section can't find it.
image

BigRoy:
> Could we add the ayon icon in the `AYON Mute Layers` , the help section can't find it. image

It should exist there. It also didn't work for me for the Load Asset LOP so I'm unsure what's wrong in this case since I assumed something on my end was just buggy and it'd never display.

Does https://github.com/ynput/ayon-houdini/pull/56/commits/ad6fe6491431b665b1088fe275d1abaa8dc28100 fix it for you @MustafaJafar ?

MustafaJafar:
I think I've hit the issue mentioned in this comment https://github.com/ynput/ayon-houdini/pull/56#issuecomment-2257101711
I ended up checking my layer stack and picking names.

Since, the variant sets (department names) are hardcoded/specified in settings, Are we able to get the reference path of the current variant ? I think it won't care about how I named my files.

image

client/ayon_houdini/startup/otls/ayon_lop_mute_layers.hda/ayon_8_8Lop_1mute__layers_8_81.0/Sections.list
```diff
@@ -0,0 +1,10 @@
+""
+DialogScript	DialogScript
+CreateScript	CreateScript
+InternalFileOptions	InternalFileOptions
+Contents.gz	Contents.gz
+TypePropertiesOptions	TypePropertiesOptions
+Help	Help
+Tools.shelf	Tools.shelf
+IconImage	IconImage
+ExtraFileOptions	ExtraFileOptions
```
MustafaJafar, line 10:
```suggestion
ExtraFileOptions	ExtraFileOptions
AYON__icon.png	AYON_icon.png
```

client/ayon_houdini/startup/otls/ayon_lop_mute_layers.hda/ayon_8_8Lop_1mute__layers_8_81.0/ExtraFileOptions
```diff
@@ -0,0 +1,2 @@
+{
+}
```
MustafaJafar, line 2:
```suggestion
{
    "AYON_icon.png/Cursor":{
		"type":"intarray",
		"value":[0,0]
	},
	"AYON_icon.png/IsExpr":{
		"type":"bool",
		"value":false
	},
	"AYON_icon.png/IsPython":{
		"type":"bool",
		"value":false
	},
	"AYON_icon.png/IsScript":{
		"type":"bool",
		"value":false
	},
	"AYON_icon.png/Source":{
		"type":"string",
		"value":"AYON_icon.png"
	}
}
```

MustafaJafar (commented):
> It should exist there. It also didn't work for me for the Load Asset LOP so I'm unsure what's wrong in this case since I assumed something on my end was just buggy and it'd never display.
> 
> It's as if on my end it never refreshes the HELP docs :)
> 
> Does [ad6fe64](https://github.com/ynput/ayon-houdini/commit/ad6fe6491431b665b1088fe275d1abaa8dc28100) fix it for you @MustafaJafar ?

It needs two additional changes to work. 
After applying the following changes, it works on my side.
image

client/ayon_houdini/startup/otls/ayon_lop_mute_layers.hda/ayon_8_8Lop_1mute__layers_8_81.0/Sections.list
BigRoy: Oh nice - this makes me wonder whether it's possible to set up 1 icon that we can share across all the HDAs instead of having to have the png in every single one of them. I guess something for another PR.

BigRoy:
> Since, the variant sets (department names) are hardcoded/specified in settings, Are we able to get the reference path of the current variant ? I think it won't care about how I named my files.

I'm not sure why/when you'd _actually_ want to mute a specific variant's layer but that may be me missing the point. Usually it's way higher level just muting a specific full department layer - but maybe on  your end you somehow have messed up not working department layers at all.

You seem to also be Loading **Shot** what is essentially an Asset Structure - so you should be loading an asset (reference) so that you can load into the scene structure the way you like instead of layering in the exact hierarchy of the asset file without control over where to put it into the scene.

BigRoy:
@MustafaJafar with you **exposed** does that mean this PR still has fixes to do? or is it actually ready for approval?

Layer muting in USD works off of filepaths (or actually `layer.identifier` I believe; which is usually the filepath or the AYON URI!? But we'll have to see that when that workflow is implemented; would just mean exposing other useful 'defaults' by then).

We can improve over time of course when issues arise.

MustafaJafar (approved):
Played with this PR and it is awesome 😉

PR #57, opened 2024-07-30
Load Asset LOP: Make Set Project or Set Folder Path appear as pop-up dialog

BigRoy:
## Changelog Description

 Make Set Project or Set Folder Path appear as pop-up dialog similar to Set Product Name

## Additional info

image

## Testing notes:

1. Load Asset LOP should work
2. The arrow button next to Project and Folder Path parm should work as intended and show a nice dialog at right location near the cursor


MustafaJafar (approved):
This is so cool! 
It integrates seamlessly with Native Houdini menus.

PR #59, opened 2024-08-02
Fix Collect USD Layers refactor and transfer `productGroup` to the extra product

BigRoy:
## Changelog Description

This fixes the Collect USD Layers logic where **Explicit layers** generated within the current solaris graph will turn into their own products.

## Additional info

You'll need to have an "Explicit" layer in your solaris graph that will be saved along:

image

## Testing notes:

1. Create a Solaris graph which generates an Explicit Save Layer as part of the graph.

E.g. this Sop Import generating a new 'layer' merged in:

image

Note the **Layer Save Path** - it must be explicit otherwise Houdini's USD ROP node will error.

2. Publish USD
3. Should work with and without USD contribution workflow


MustafaJafar (approved):
This is amazing! 

image

image

PR #62, opened 2024-08-02
fix wrong name when calling `setName`

### LINKED ISSUE #60: Creating placeholders fail whn using a regex pattern instead of a plain string.

### Current Behavior:

Creating placeholders may fail as `placeholder_data["product_name"]` is a regex pattern. which may contain characters that houdini don't like for node names. 

https://github.com/ynput/ayon-houdini/blob/56998089202d224a91650987194cca4ea92628b6/client/ayon_houdini/plugins/workfile_build/load_placeholder.py#L33-L39

### Expected Behavior:

to make it not fail.

### Version

1.0.0

Windows

### Steps To Reproduce:

1. crate a placeholder while using a regex pattern not only a plain str.

### Relevant log output:

_No response_

### Additional context:

suggested solution by `BigRoy` to add 
`force_valid_node_name=True` 
https://www.sidefx.com/docs/houdini/hom/hou/Node.html#adding-and-removing

MustafaJafar:
## Changelog Description
Follow up PR to https://github.com/ynput/ayon-houdini/pull/61 
as we missed to fix the name when updating the node names

## Testing notes:

1. update load placeholder, use a regex pattern not only a plain str.

### DIFF
--- a/client/ayon_houdini/api/workfile_template_builder.py
+++ b/client/ayon_houdini/api/workfile_template_builder.py
@@ -231,6 +231,7 @@ class HoudiniPlaceholderPlugin(PlaceholderPlugin):
         
         # Update node name
         node_name = self.get_placeholder_node_name(placeholder_data)
+        node_name = hou.text.variableName(node_name)
         placeholder_node.setName(node_name, unique_name=True)
 
     def delete_placeholder(self, placeholder):


BigRoy (approved):
As discussed on discord, this looks like the right fix. :)

For context: docs on `hou.text` are here https://www.sidefx.com/docs/houdini/hom/hou/text.html

PR #64, opened 2024-08-06
Add `ValidateWaitForRender`

### LINKED ISSUE #63: `Wait for render to complete` cause the usdrender to fail

### Current Behavior:

When selecting `Local machine rendering` while rendering, the publish may fail.

Extract Render will complain about missing frames. which happens because files won't exist at this point of time during execution unless we have `Wait for Render to Complete` enabled.

### Expected Behavior:

It should report to user that the setting is disabled.

### Version

1.0.0

Windows

### Steps To Reproduce:

1. Create USD Render
2. When publishing, select `Local machine rendering`

### Relevant log output:

_No response_

### Additional context:

_No response_

MustafaJafar:
## Changelog Description
Fix #63 

## Testing notes:
1. Create USD Render
2. When publishing, disable `Driver > wait for render to complete`

### DIFF
--- a/client/ayon_houdini/plugins/publish/validate_wait_for_render.py
+++ b/client/ayon_houdini/plugins/publish/validate_wait_for_render.py
@@ -32,7 +32,7 @@ class ValidateWaitForRender(plugin.HoudiniInstancePlugin):
             # Ignore instances without an instance node
             # e.g. in memory bootstrap instances
             self.log.debug(
-                "Skipping instance without instance node: {}".format(instance)
+                f"Skipping instance without instance node: {instance}"
             )
             return
         
--- a/client/ayon_houdini/plugins/publish/validate_wait_for_render.py
+++ b/client/ayon_houdini/plugins/publish/validate_wait_for_render.py
@@ -1,0 +1,68 @@
+# -*- coding: utf-8 -*-
+import hou
+
+import pyblish.api
+from ayon_core.pipeline import PublishValidationError
+from ayon_core.pipeline.publish import RepairAction
+
+from ayon_houdini.api import plugin
+
+
+class EnableWaitForRender(RepairAction):
+    label = "Enable WaitForRendertoComplete"
+    icon = "mdi.pencil-plus-outline"
+
+
+class ValidateWaitForRender(plugin.HoudiniInstancePlugin):
+    """Validate `WaitForRendertoComplete` is enabled.
+
+    Disabling `WaitForRendertoComplete` cause the local render to fail
+    as the publish execution continues while the render may not be finished yet.
+
+    """
+
+    order = pyblish.api.ValidatorOrder
+    families = ["usdrender"]
+    label = "Validate Wait For Render to Complete"
+    actions = [EnableWaitForRender]
+
+    def process(self, instance):
+
+        if not instance.data.get("instance_node"):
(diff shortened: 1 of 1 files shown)


client/ayon_houdini/plugins/publish/validate_wait_for_render.py
```diff
@@ -0,0 +1,68 @@
...
+import pyblish.api
+from ayon_core.pipeline import PublishValidationError
+from ayon_core.pipeline.publish import RepairAction
+
+from ayon_houdini.api import plugin
+
+
+class EnableWaitForRender(RepairAction):
+    label = "Enable WaitForRendertoComplete"
+    icon = "mdi.pencil-plus-outline"
```
moonyuet, line 13:
```suggestion

```

client/ayon_houdini/plugins/publish/validate_wait_for_render.py
```diff
@@ -0,0 +1,68 @@
...
+
+    Disabling `WaitForRendertoComplete` cause the local render to fail
+    as the publish execution continues while the render may not be finished yet.
+
+    """
+
+    order = pyblish.api.ValidatorOrder
+    families = ["usdrender"]
+    label = "Validate Wait For Render to Complete"
+    actions = [EnableWaitForRender]
```
moonyuet, line 27: Any reason why we need to have the extra class inherited from the Repair action?
```suggestion
    actions = [RepairAction]
```

client/ayon_houdini/plugins/publish/validate_wait_for_render.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def process(self, instance):
+
+        if not instance.data.get("instance_node"):
+            # Ignore instances without an instance node
+            # e.g. in memory bootstrap instances
+            self.log.debug(
+                "Skipping instance without instance node: {}".format(instance)
```
moonyuet, line 35:
```suggestion
                f"Skipping instance without instance node: {instance}"
```

moonyuet (commented):
Did a simple test, When I turned off the WaitForRendertoComplete, the validator would check on the parameter as expected.
image
When I repair it, it pass the validation as expected.
image

I just have some comments on the RepairAction, otherwise looks good and can approve.

client/ayon_houdini/plugins/publish/validate_wait_for_render.py
MustafaJafar: > Any reason why we need to have the extra class inherited from the Repair action?

yes, so that I can change the label and the icon. 
Here's how the default repair action looks.
image

Here's how the extra class looks like. which makes the button action clearer.
image

moonyuet: I would prefer just purely repair action

client/ayon_houdini/plugins/publish/validate_wait_for_render.py
MustafaJafar: implemented here, Thank you
https://github.com/ynput/ayon-houdini/pull/64/commits/9f2c93acaa9b78fe05674bef2517d2ae02b30f49

moonyuet (approved):
I have done another round of test. Looks good to me.

PR #67, opened 2024-08-08
Render layer name collector

antirotor:
## Changelog Description
This PR adds collector for render layer names - so they can be used later on in anatomy templates (either for publishing or delivery).

> [!WARNING]
Due to the dynamic nature of product templates it is difficult to determine good `group_name` from that. It should be basically `product_name` without **aov** data in it, but since the product name template might look like `{product[type]}<{Renderlayer}><_{Aov}>`, it is difficult to get it just from the resolved string.

## Testing notes:

1. With mentioned ayon-core PR enable new behavior in the settings
2. Set render product name template correctly
3. Render something (locally and on farm)

### DIFF
--- a/client/ayon_houdini/plugins/publish/collect_local_render_instances.py
+++ b/client/ayon_houdini/plugins/publish/collect_local_render_instances.py
@@ -73,6 +73,9 @@ class CollectLocalRenderInstances(plugin.HoudiniInstancePlugin):
             dynamic_data = {}
             if aov_name:
                 dynamic_data["aov"] = aov_name
+                
+            if instance.data.get("renderlayer"):
+                dynamic_data["renderlayer"] = instance.data["renderlayer"]
 
             product_name, product_group = get_product_name_and_group_from_template(
                 project_name=context.data["projectName"],
--- a/client/ayon_houdini/plugins/publish/collect_local_render_instances.py
+++ b/client/ayon_houdini/plugins/publish/collect_local_render_instances.py
@@ -1,9 +1,10 @@
 import os
-
+import warnings
 import pyblish.api
 from ayon_core.pipeline.farm.patterning import match_aov_pattern
 from ayon_core.pipeline.farm.pyblish_functions import (
-    get_product_name_and_group_from_template
+    get_product_name_and_group_from_template,
+    _get_legacy_product_name_and_group
 )
 from ayon_core.pipeline.publish import (
     get_plugin_settings,
@@ -96,13 +97,10 @@ class CollectLocalRenderInstances(plugin.HoudiniInstancePlugin,
             if instance.data.get("renderlayer"):
                 dynamic_data["renderlayer"] = instance.data["renderlayer"]
 
-            product_name, product_group = get_product_name_and_group_from_template(
-                project_name=context.data["projectName"],
-                task_entity=context.data["taskEntity"],
-                host_name=context.data["hostName"],
-                product_type=product_type,
-                variant=instance.data["productName"],
-                dynamic_data=dynamic_data,
+            product_name, product_group = self._get_product_name_and_group(
+                instance,
+                product_type,
+                dynamic_data
             )
 
             # Create instance for each AOV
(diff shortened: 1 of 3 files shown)


moonyuet (commented):
Did a simple test with both local and farm rendering mantra. It is rendered successfully.
image
I will test it with different template later to double confirm

client/ayon_houdini/plugins/publish/collect_local_render_instances.py
```diff
@@ -62,23 +65,23 @@ def process(self, instance):
...
-        )
-
-        for aov_name, aov_filepaths in expectedFiles.items():
-            product_name = product_group
 
+        for aov_name, aov_filepaths in expected_files.items():
+            dynamic_data = {}
             if aov_name:
-                product_name = "{}_{}".format(product_name, aov_name)
+                dynamic_data["aov"] = aov_name
```
MustafaJafar, line 75: It misses render layer dynamic data. 

```suggestion
            dynamic_data = {
                "renderlayer": instance.data["renderlayer"]
            }
            if aov_name:
                dynamic_data["aov"] = aov_name
```

client/ayon_houdini/plugins/publish/collect_local_render_instances.py
```diff
@@ -62,23 +65,23 @@ def process(self, instance):
...
+
+            product_name, product_group = get_product_name_and_group_from_template(
+                project_name=context.data["projectName"],
+                task_entity=context.data["taskEntity"],
+                host_name=context.data["hostName"],
+                product_type=product_type,
+                variant=instance.data["productName"],
+                dynamic_data = dynamic_data,
```
moonyuet, line 83:
```suggestion
                dynamic_data=dynamic_data,
```

MustafaJafar:
I need to say, that adding a wrong/bad value for templates name profiles can lead to this error.
image

and, they may be clueless until they open the details page.

image

So, could we do something to help about it ? 
maybe `CollectLocalRenderInstances`  message about creating instances with the same name.

moonyuet:
> I need to say, that adding a wrong/bad value for templates name profiles can lead to this error.

This product name template is not applicable for the houdini host only, guess we should check if there is wrong template name generally.

> So, could we do something to help about it ? maybe `CollectLocalRenderInstances` message about creating instances with the same name.

From your screenshot, it seems there is something wrong when creating render instances with specific name which these variants name doesn't get check which leads to the duplicates.  It is supposed that it is always checking(it works at least when we create rop instance with the same names.

client/ayon_houdini/plugins/publish/collect_local_render_instances.py
antirotor: I've added *renderlayer* in optional way so it is not broken whenever this plugin changes families (and the render layer collector doesn't). Dynamic data shouldn't be mandatory anyway.

antirotor:
I've added some help in https://github.com/ynput/ayon-core/pull/841/commits/f752940c1175cfd2bd9665264e218ed0732622c1

With product name templates (not just connected to these PRs), they need to be configured really carefuly. Any mistake can break a lot of things so it shouldn't be something artists themselves hit that often.

MustafaJafar:
Firstly, I must have a template profile for render in Houdini, otherwise the publisher will complain about having multiple instances with the same name.
image

---

Farm doesn't use the correct naming but it works locally. 
image

Local Log: 
image

From Farm job json file: 
```
{
    .
    .
    "instances": [
        {
            "aov": "beauty",
            "productGroup": "renderCfxMantra_ropTest",
            "productName": "renderCfxMantra_ropTest_beauty",
            "productType": "render",
            "renderlayer": "mantra_ropTest",
            .
            .
        },
        {
            "aov": "N",
            "productGroup": "renderCfxMantra_ropTest",
            "productName": "renderCfxMantra_ropTest_N",
            "productType": "render",
            "renderlayer": "mantra_ropTest",
            .
            .
        },
        {
            "aov": "P",
            "productGroup": "renderCfxMantra_ropTest",
            "productName": "renderCfxMantra_ropTest_P",
            "productType": "render",
            "renderlayer": "mantra_ropTest",
            .
            .
        }
    ],
    .
    .
}
```

MustafaJafar (commented):
Firstly, I must have a template profile for render in Houdini, otherwise the publisher will complain about having multiple instances with the same name.
image

---

Farm doesn't use the correct naming but it works locally. 
image

Local Log: 
image

From Farm job json file: 
```json
{
    .
    .
    "instances": [
        {
            "aov": "beauty",
            "productGroup": "renderCfxMantra_ropTest",
            "productName": "renderCfxMantra_ropTest_beauty",
            "productType": "render",
            "renderlayer": "mantra_ropTest",
            .
            .
        },
        {
            "aov": "N",
            "productGroup": "renderCfxMantra_ropTest",
            "productName": "renderCfxMantra_ropTest_N",
            "productType": "render",
            "renderlayer": "mantra_ropTest",
            .
            .
        },
        {
            "aov": "P",
            "productGroup": "renderCfxMantra_ropTest",
            "productName": "renderCfxMantra_ropTest_P",
            "productType": "render",
            "renderlayer": "mantra_ropTest",
            .
            .
        }
    ],
    .
    .
}
```

MustafaJafar (approved):
Hey, this works fine on my side. 
My local render and farm render are following my render product template `{product[type]}<_{Renderlayer}><_{Aov}>` 

For information this toggle works only fon farm.
- `ayon+settings://core/tools/creator/use_legacy_product_names_for_renders`
image

antirotor:
> For information this toggle works only for farm. Should we port its [logic](https://github.com/ynput/ayon-core/blob/d00fa72892f24bf50527838761016f8a28379c98/client/ayon_core/pipeline/farm/pyblish_functions.py#L705-L730) to `CollectLocalRenderInstances`?

ah, we definitely should...

MustafaJafar (approved):
This now is working in local and farm. 
Both respect `use_legacy_product_names_for_renders` logic.

antirotor:
Tested and it works indeed, thanks.

PR #70, opened 2024-08-09
Support publishing and loading node presets

### LINKED ISSUE #68: Render Setup Workflow alternative for Houdini

In Maya, there is "Render Setup" workflow, serializing settings in Render Settings and publishing them as JSON files with ability to load them again.

Something similar can be done for Houdini too, from simple serialization of subset of values on ROP node to HDA based solution.

As first step, simple value serialization can be enough.

..

_No response_

### Additional context:

_No response_

Internal Ticket Number: AY-6061

MustafaJafar:
## Changelog Description
Fixes https://github.com/ynput/ayon-houdini/issues/68, This PR Implements publishing and loading node presets.
A node preset is simple serialization of subset of values on ROP node

There are still some edge cases to handle (e.g. animated parameters) and some code improvements, but that doesn't affect reviewing the suggested workflow, shown in the demo video below.

## Additional info
Render setups in Houdini can be
1. a collection of nodes (you can call it: a scene fragment, or a part of a node graph)
  This is partially solved by publishing HDAs as we can publish HDAs from different Houdini contexts.
  Also, we can use `CPIO`  which should work in a similar manner as publishing node presets.
2. a node preset of my render rop node. (which is solved by this PR)

## Additional info 2
The new recopies system is an alternative solution to creating the node preset json file ourselves. 
But, this is a topic for the future as it's only supported in Hou 20.5 
https://www.sidefx.com/docs/houdini/network/recipes.html

## Demo
https://github.com/user-attachments/assets/df1ee571-07c4-4351-b5ca-a86b64a475d1

## Testing notes:
**Publishing:**
1. Create a node preset
2. Update Source node parameter 
3. Publish 
**Loading:**
1. Load a preset through the loader
2. Mange its version via inventory
3. Update target node parameter
4. Click `Apply Preset`


client/ayon_houdini/plugins/load/load_filepath.py
```diff
@@ -25,6 +25,10 @@ class FilePathLoader(plugin.HoudiniLoader):
     product_types = {"*"}
     representations = {"*"}
 
+    def _add_more_node_params(self, attr_folder, node):
```
moonyuet, line 28: we can move this function outside of the class as it is just used in `NodePresetLoader`? We dont need to pass the function here if we do that.
```suggestion
    def _add_more_node_params(self, attr_folder, node):
```

client/ayon_houdini/plugins/create/create_node_preset.py
```diff
@@ -0,0 +1,105 @@
+# -*- coding: utf-8 -*-
+"""Creator plugin for creating houdini node presets."""
+from ayon_houdini.api import plugin
+import hou
+
+
+def _update_node_parmtemplate(node, defaults):
```
moonyuet, line 7: A question: Does this function imprint the data into the instance_node already? (Or it is like the temp data publisher and it won't imprint the instance data?)
i.e. some variables you can find in almost every creator and they are missing in this creator.
```
            instance = CreatedInstance(
                self.product_type,
                product_name,
                instance_data,
                self)
            self._add_instance_to_context(instance)
            self.imprint(instance_node, instance.data_to_store())

```

MustafaJafar: The method `_update_node_parmtemplate` doesn't create the extra ayon parameters created by imprint.
It actually doesn't create any parameters at all. 
It only captures the values of any existent parameters (including ayon extra parameters) from the source node and reassigns these values to the target node.

---

I have that feeling that the method name should be refactored but I couldn't come up with a better name :/

client/ayon_houdini/plugins/load/load_filepath.py
MustafaJafar: It's actually used inside `FilePathLoader` in the middle of  `FilePathLoader.load` method. 
Then I override it in `NodePresetLoader` to add the extra custom parameters to the `Attributes` folder (`attr_folder`). 

I chose this way to avoid copying/pasting the code of `FilePathLoader`

---

I thought about using a different approach where I do something like this in the subclass `NodePresetLoader` .

```python
class NodePresetLoader(FilePathLoader):
    label = "Load Node Preset"
    order = 9
    icon = "link"
    color = "white"
    product_types = {"node_preset"}
    representations = {"json"}

    def load(self, context, name=None, namespace=None, data=None):
      container = super(NodePresetLoader, self).load(context, name=name, namespace=namespace, data=data)
      node = container["node"]
      parm_group = node.parmTemplateGroup()
      attr_folder = parm_group.find("attributes_folder") 
      ## But this won't work because Houdini overrides the parameter name.
```
image

moonyuet: In regards to your second approach, I think your first approach is better.

client/ayon_houdini/plugins/create/create_node_preset.py
moonyuet: I remember @BigRoy has written a one-time publish node(along with the generic nodes for publishing). Maybe he can give some maps on this.

client/ayon_houdini/plugins/load/load_filepath.py
```diff
@@ -128,3 +136,111 @@ def format_path(path: str, representation: dict) -> str:
...
+
+hou.pwd().setColor(hou.Color(0.7, 0.8, 0.87))
+hou.pwd().setComment("")
+hou.pwd().setGenericFlag(hou.nodeFlag.DisplayComment, True)
+if target_node and target_node.type().name() != node_type:
+    hou.pwd().setColor(hou.Color(0.8, 0.45, 0.1))
+    hou.pwd().setComment(
+        f"Target Node type '{target_node.type().name()}' doesn't match the loaded preset type '{node_type}'."
+        "Please note, Applying the preset skips parameters that doesn't exist"
+    )
```
BigRoy, line 192: Should we make this just a call to a function inside `ayon_houdini.api.lib` so that we can update it from the codebase, instead of storing the code on the node in the actual scenes?

So that we just have `lib.apply_node_preset(filepath)` or `lib.apply_node_preset(json_data)` or alike.

MustafaJafar: Very good point!
Totally yes. let me do it.

MustafaJafar: https://github.com/ynput/ayon-houdini/pull/70/commits/3818b57bda704dce2a41f44e149c6e9bf228dba2

BigRoy (commented):
I gave this PR a whirl - I guess it does what it says.

You can:

1. Create a 'fake' ROP node that will generate a `.json` dump of the parms of the node specified on that ROP.
2. Publish that to write the `.json` file.
3. Load that to create another null placeholder node pointing to that published preset.
4. Click a button on that node to apply the preset.

---

YET - I really don't see the point myself. 😕 

1. It works for ONE node only
2. If this is intended as presets for ROPs - which are also our publish instance nodes... applying the preset also overrides the target folder, target task, etc. 
3. The node stored in the scene that applies the presets I really don't see any reason for. The only upside of that is that you can see in manager which preset is 'loaded' **however** that is very unlikely the case. Once you load it, you need to apply it manually separately... and even if anyone changes any parameters - you have no clue whether the node applied to still matches the preset.

Also, we're basically matching Houdini's own 'presets' (which are odd files, sure) but I just don't grasp what this would solve over their system.

The Houdini presets can also be saved and shared within a project location.

---

If for whatever reason anyone _really_ likes this and we want to add it. I'd kill the way it loads. 

Just apply the preset via the loader to selected nodes - no manage, no custom loaded nodes that you need to browse and configure, etc.

If you really want to go crazy write some "node menu" customization that allows you to add an `AYON Presets` submenu on a node that basically lists all presets in the project or whatever.
image

But even then - I feel like we're way overthinking this. I just don't get it.

Also, there are some "preset managers" tools for Houdini out there that can do some pretty slick stuff.

---

There's also _so much_ goodies in Houdini 20.5 that do this, but more... I really feel like it's time wasted if we just don't adopt in that direction now if we want to go crazy with some nice stuff and "live with just a preset manager or houdini presets" for now.

There's some good [API](https://www.sidefx.com/docs/houdini/hom/hou/data.html), [docs on recipes](https://www.sidefx.com/docs/houdini/network/recipes.html#hom) and even nodes have `asData()` methods.

For this to be anywhere user-friendly I think we also need to define for the ticket/issue more the production use cases. 

- How often do you apply a preset? 
- Do you switch them around?
etc.

I have a feeling that now, going through the loader to find them, which then loads a node, that I need to configure, etc. I've already spend so much time that could've just been a single click on the node itself using other tools - like e.g. Houdini's own preset system (which also sucks, but it has basically the same feature set, exports all parms and reapplies all parms)

BigRoy:
Just noting there that this started from a client request but was put on hold for now - potentially in favor of Houdini 20.5+ Recipes in the future.

MustafaJafar:
After some discussion, we decided to close this PR in favor of Houdini 20.5+ Recipes.

PR #72, opened 2024-08-20
Support merged AOVs in arnold render

### LINKED ISSUE #69: Arnold ROP AOV multichannel EXR

### Current Behavior:

render/publish multi-channel exrs from Houdini arnold rop fails when rendering a multi-channel exr with all aovs combined.

after quick inspection, I've found this TODO.
https://github.com/ynput/ayon-houdini/blob/c239540d12742b61893fb2bbf4042d434e616b9d/client/ayon_houdini/plugins/publish/collect_arnold_rop.py#L73-L74

### Expected Behavior:

`CollectArnoldROPRenderProducts` should compute the correct list of frames if all aovs combined.

### Version

1.0.0

Windows

### Steps To Reproduce:

1. render/publish multi-channel exrs from Houdini arnold rop with all aovs combined.

### Relevant log output:

_No response_

### Additional context:

_No response_

Internal Ticket Number: AY-6475

MustafaJafar:
## Changelog Description
Fixes https://github.com/ynput/ayon-houdini/issues/69

## Test Runs
Here's a test run on my side. 
The setup is shown in the ROP node parameters and the created files are shown in the removed files log.
image

## Testing notes:

1. Create AYON Arnold ROP publish instance.
2. Configure the ROP node to have multiple AOVs and set them to separated files
3. Publish. AOVs with separated files should be a separated publish instance.

### DIFF
--- a/client/ayon_houdini/plugins/publish/collect_arnold_rop.py
+++ b/client/ayon_houdini/plugins/publish/collect_arnold_rop.py
@@ -70,11 +70,15 @@ class CollectArnoldROPRenderProducts(plugin.HoudiniInstancePlugin):
         multipartExr = True
 
         num_aovs = rop.evalParm("ar_aovs")
-        # TODO: Check the following logic.
-        #       as it always assumes that all AOV are not merged.
+
         for index in range(1, num_aovs + 1):
-            # Skip disabled AOVs
-            if not rop.evalParm("ar_enable_aov{}".format(index)):
+            
+            aov_enabled = rop.evalParm("ar_enable_aov{}".format(index)) 
+            aov_sep = rop.evalParm("ar_aov_separate{}".format(index))
+            aov_path = rop.evalParm("ar_aov_separate_file{}".format(index))
+            
+            # Skip disabled AOVs or AOVs with no separate aov file path
+            if not all((aov_enabled, aov_path, aov_sep)):
                 continue
 
             if rop.evalParm("ar_aov_exr_enable_layer_name{}".format(index)):
@@ -82,8 +86,15 @@ class CollectArnoldROPRenderProducts(plugin.HoudiniInstancePlugin):
             else:
                 label = evalParmNoFrame(rop, "ar_aov_label{}".format(index))
 
-            aov_product = self.get_render_product_name(default_prefix,
-                                                       suffix=label)
+            # NOTE:
+            #  we don't collect the actual AOV path but rather assume 
+            #    the user has used the default beauty path (collected above)
+            #    with the AOV name before the extension.
+            #  Also, Note that Ayon Publishing does not require a specific file name,
+            #    as it will be renamed according to the naming conventions set in the publish template.
+            aov_product = self.get_render_product_name(
+                prefix=default_prefix, suffix=label
+            )
             render_products.append(aov_product)
             files_by_aov[label] = self.generate_expected_files(instance,
                                                                aov_product)


client/ayon_houdini/plugins/publish/collect_arnold_rop.py
```diff
@@ -70,20 +70,31 @@ def process(self, instance):
...
                 label = evalParmNoFrame(rop, "ar_aov_label{}".format(index))
 
-            aov_product = self.get_render_product_name(default_prefix,
-                                                       suffix=label)
+            # NOTE:
+            #  we don't collect the actual AOV path but rather assume 
+            #    the user has used the default beauty path (collected above)
+            #    with the AOV name before the extension.
+            #  Also, Note that Ayon Publishing does not require a specific file name,
+            #    as it will be renamed according to the naming conventions set in the publish template.
```
BigRoy, line 94: Sorry - why do we assume that? Isn't it trivial to 'fix' that?

Also it seems that all [this logic](https://github.com/ynput/ayon-houdini/blob/a7c1caa098a7e623956023420366687a44a3c2f6/client/ayon_houdini/plugins/publish/collect_arnold_rop.py#L133-L145) isn't doing anything at all for Arnold. `${AOV}` and `$AOV` and `%AOV%` do nothing with the Arnold ROP. So that logic can be removed? Or?

Because we're now also due to this failing to correctly detect the images if they just happen to have a very different path set?

BigRoy: We should somehow differentiate the logic between:

- What are the expected file paths? (We know that, because that's explicitly defined on the parm, yay!)
- What will we make the product name for the AOV - preferably that's the main product name, but with a suffix of the AOV or alike? So we'll somehow need to find a part of the filename the user has defined that we can use for that.

moonyuet: > Sorry - why do we assume that? Isn't it trivial to 'fix' that?
> 
> Also it seems that all [this logic](https://github.com/ynput/ayon-houdini/blob/a7c1caa098a7e623956023420366687a44a3c2f6/client/ayon_houdini/plugins/publish/collect_arnold_rop.py#L133-L145) isn't doing anything at all for Arnold. `${AOV}` and `$AOV` and `%AOV%` do nothing with the Arnold ROP. So that logic can be removed? Or?
> 
> Because we're now also due to this failing to correctly detect the images if they just happen to have a very different path set?

We need to improve (and even remove) the logic above to make sure the AOV files included in the expected file paths. 
And we need to think about how to deal with the cryptomatte too: https://help.autodesk.com/view/ARNOL/ENU/?guid=arnold_for_houdini_shaders_ah_Cryptomatte_html

kartikgupta:
Thannks @MustafaJafar and @BigRoy for looking into it.

There are parts of the pull request that I don't understand completely.
Is it trying to support separated AOVs, or combined AOVs?
Looking at the previous comments, the discussion seems to be revolving around how to resolve the aov file paths.

But what we're looking to get is that it shouldn't look for the AOV file paths at all and just integrate the overall file path mentioned in `ar_picture` attribute of the rop, instead of integrating each AOV.
Not sure if there's something I'm missing here.

BigRoy:
> But what we're looking to get is that it shouldn't look for the AOV file paths at all and just integrate the overall file path mentioned in `ar_picture` attribute of the rop, instead of integrating each AOV.
> Not sure if there's something I'm missing here.

It should do that with the current changes as it would now ignore active AOVs that are not explicitly set to be a separate file.

> Is it trying to support separated AOVs, or combined AOVs?

In the comments we're trying to make sure it does both correctly - but the part that allows combined AOVs should already have been solved looking at the code.

Anyway, I'll do some testing.

kartikgupta:
Thanks @BigRoy for clearing up the misunderstanding!

client/ayon_houdini/plugins/publish/collect_arnold_rop.py
BigRoy: So this PR fixes a current critical issue already - and the 'naming' issue here isn't new and was already present. I'd say, let's merge this now and keep the comment and create a new issue @MustafaJafar to track for us that we should fix this naming for the separate layers down the line.

BigRoy (approved):
This fixes at least being able render MERGED aovs.

We should follow up with a new issue and PR after this to solve the remaining TODO and the notes given in [this comment](https://github.com/ynput/ayon-houdini/pull/72#discussion_r1745608538).

_Note_ that I have not tested whether Cryptomattes break the current "merged" AOV publishing so that remains unconfirmed.

PR #80, opened 2024-08-27
USD ROP allow remapping explicit layer save paths to be remapped to AYON Entity URI

BigRoy:
## Changelog Description

Allow remapping to AYON Entity URI instead of resolved publish filepath for layer save paths directly written out from the USD ROP that also turn into their own publishable instance.

## Testing notes:

Enable `Extract USD` attribute `use_ayon_entity_uri` in settings.

1. Create a Solaris stage that generates an additional layer save path (explicit layer save path)
2. This layer save path will introduce an additional publish instance.
3. With the setting enabled that instance should be remapped to an AYON Entity URI
4. With the setting disabled that instance should be remapped to the resolved publish path (same as before this PR)

*When using _Merge LOP_ to merge multiple explicit layer save paths (e.g. separate SOP imports with explicit layer save paths) keep in mind that Merge LOP default merge mode "Simple merge" will discard them. More details, see [here](https://github.com/ynput/ayon-houdini/pull/80#issuecomment-2362864108)

### DIFF
--- a/client/ayon_houdini/plugins/publish/extract_usd.py
+++ b/client/ayon_houdini/plugins/publish/extract_usd.py
@@ -18,7 +18,7 @@ class ExtractUSD(plugin.HoudiniExtractorPlugin):
     label = "Extract USD"
     families = ["usdrop"]
 
-    use_ayon_entity_uri = True
+    use_ayon_entity_uri = False
 
     def process(self, instance):
 
--- a/client/ayon_houdini/plugins/publish/extract_usd.py
+++ b/client/ayon_houdini/plugins/publish/extract_usd.py
@@ -3,11 +3,11 @@ from typing import List, AnyStr
 
 import pyblish.api
 
+from ayon_core.pipeline.entity_uri import construct_ayon_entity_uri
 from ayon_core.pipeline.publish.lib import get_instance_expected_output_path
 from ayon_houdini.api import plugin
 from ayon_houdini.api.lib import render_rop
 from ayon_houdini.api.usd import remap_paths
-from ayon_core.pipeline.entity_uri import construct_ayon_entity_uri
 
 import hou
 
--- a/client/ayon_houdini/plugins/publish/extract_usd.py
+++ b/client/ayon_houdini/plugins/publish/extract_usd.py
@@ -3,19 +3,25 @@ from typing import List, AnyStr
 
 import pyblish.api
 
+from ayon_core.lib import BoolDef
+from ayon_core.pipeline import AYONPyblishPluginMixin
 from ayon_core.pipeline.publish.lib import get_instance_expected_output_path
 from ayon_houdini.api import plugin
 from ayon_houdini.api.lib import render_rop
 from ayon_houdini.api.usd import remap_paths
+from ayon_core.pipeline.entity_uri import construct_ayon_entity_uri
 
 import hou
 
 
-class ExtractUSD(plugin.HoudiniExtractorPlugin):
+class ExtractUSD(plugin.HoudiniExtractorPlugin, AYONPyblishPluginMixin):
 
     order = pyblish.api.ExtractorOrder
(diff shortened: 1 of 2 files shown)


BigRoy:
> im not sure what is happening but it dose not create the output i would have expected.
> 
> i belive its only extracting 1 of the explicit save paths
> 
> here is an image to what i see not sure how i would test or debug things here. Screenshot from 2024-08-29 14-40-50

What a test! :) Can you share a scene file?

Lypsolon:
> > im not sure what is happening but it dose not create the output i would have expected.
> > i belive its only extracting 1 of the explicit save paths
> > here is an image to what i see not sure how i would test or debug things here. Screenshot from 2024-08-29 14-40-50
> 
> What a test! :) Can you share a scene file?


here is the scene used for testing. 

but i will also say i dont think this is part of the current USD epic so im gonna postpone further testing for the next epic. 

[UB_ep2_layout_v002.zip](https://github.com/user-attachments/files/16812796/UB_ep2_layout_v002.zip)

MustafaJafar (commented):
Here you are some test runs.. 
I don't have the resolver yet, but I can at least validate if the PR works as expected. 
> No idea if the URI is correct or not. But, I'll assume it's correct.

So, I loaded few published products using different loaders. 
here you are the different URIs I got:
 - `ayon+entity://Robo/Assets/Character/moco&product=usdSphere&version=v006&representation=usd` 
image

- `ayon+entity://Robo/Assets/Character/moco&product=usdAsset_model&version=v005&representation=usd`
image

---
Another test, I used all the loaders to load the exact same asset: 
They all point to my `usdSphere` mentioned at the top. 
image
image
image
image
image

BigRoy:
@MustafaJafar your specific test run confirms the loading, but not this PRs saving of explicit layer save paths and embedding that in the published USD file.

---

@Lypsolon I've checked your scene files just now - and your scene actually behaves correctly. It saves only one explicit layer save path. This is due to the **MERGE** node.

See here:
image

At the merge, there's only one layer still in my Scene Graph Layers. This is due to the default being a "Simple merge".
When I change the merge style:

image

Then I get this:
image

That's when the publisher should pick that up.
image

Lypsolon:
ahh sorry rucky mistake i will test again. 
the merge node is getting me every now and then.

Lypsolon (approved):
after fixing the merge settings (Roocky mistake) 
it now works. 
Tested on Houdin 20.0 Almalinux 9 with 200Explicit Save Paths

Lypsolon:
One small note. 
something is printing the following
```
>>> We have preset for LOPLoadAssetLoader
  - setting `use_ayon_entity_uri`: `True`
>>> We have preset for LOPLoadShotLoader
  - setting `use_ayon_entity_uri`: `True`
>>> We have preset for USDSublayerLoader
  - setting `use_ayon_entity_uri`: `True`
>>> We have preset for USDReferenceLoader
  - setting `use_ayon_entity_uri`: `True`
>>> We have preset for SopUsdImportLoader
  - setting `use_ayon_entity_uri`: `True`
  ```

BigRoy:
> One small note. something is printing the following
> 
> ```
> >>> We have preset for LOPLoadAssetLoader
>   - setting `use_ayon_entity_uri`: `True`
> >>> We have preset for LOPLoadShotLoader
>   - setting `use_ayon_entity_uri`: `True`
> >>> We have preset for USDSublayerLoader
>   - setting `use_ayon_entity_uri`: `True`
> >>> We have preset for USDReferenceLoader
>   - setting `use_ayon_entity_uri`: `True`
> >>> We have preset for SopUsdImportLoader
>   - setting `use_ayon_entity_uri`: `True`
> ```

Correct - this is the discovery of the plugins in the Create Context and how it applies the project settings. It's verbose yes - but shouldn't be an issue for now.

PR #81, opened 2024-08-28
Collect current file ignore the default untitled scene when no save or load has happened yet

BigRoy:
## Changelog Description

Never allow the "new" file (the standard scene) to be collected as current file

## Additional info

This is a companion PR to https://github.com/ynput/ayon-core/pull/786
When used together a better error report is shown for unsaved files.

image

(Screenshots from Houdini)

## Testing notes:

1. Use this PR together with https://github.com/ynput/ayon-core/pull/786
2. Try publishing from an unsaved workfile
3. Restart houdini (allow it to open last workfile) - it should correctly detect that you have a scene file open now

### DIFF
--- a/client/ayon_houdini/plugins/publish/collect_current_file.py
+++ b/client/ayon_houdini/plugins/publish/collect_current_file.py
@@ -22,18 +22,8 @@ class CollectHoudiniCurrentFile(plugin.HoudiniContextPlugin):
             # By default, Houdini will even point a new scene to a path.
             # However if the file is not saved at all and does not exist,
             # we assume the user never set it.
+            self.log.warning("Houdini workfile is unsaved.")
             current_file = ""
-        else:
-            # Due to even a new file being called 'untitled.hip' we are unable
-            # to confirm the current scene was ever saved because the file
-            # could have existed already. We will allow it if the file exists,
-            # but show a warning for this edge case to clarify the potential
-            # false positive.
-            self.log.warning(
-                "Current file is 'untitled.hip' and we are "
-                "unable to detect whether the current scene is "
-                "saved correctly."
-            )
 
         context.data["currentFile"] = current_file
         self.log.info('Current workfile path: {}'.format(current_file))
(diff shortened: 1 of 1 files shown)


client/ayon_houdini/plugins/publish/collect_current_file.py
```diff
@@ -15,23 +15,15 @@ def process(self, context):
         """Inject the current working file"""
 
         current_file = hou.hipFile.path()
-        if not os.path.exists(current_file):
+        if (
+                hou.hipFile.isNewFile()
+                or not os.path.exists(current_file)
```
MustafaJafar, line 20: Should we add checking for unsaved changes ?
```suggestion
                or not os.path.exists(current_file)
                or hou.hipFile.hasUnsavedChanges()
```

MustafaJafar (approved):
This works on my side for new files.
It doesn't work if my scene has unsaved changes.

image

client/ayon_houdini/plugins/publish/collect_current_file.py
BigRoy: Sorry - I think you misunderstood, and I understand why - because of the log message.

It's **not** about having unsaved changes, that's totally up to you - as long as we know what your filepath should be for your current file. The "unsaved" I'm referring to is the case where no filepath was set at all currently (like a new scene) which Houdini by default returns as `untitled.hip` (but that's also customizable via an env var or something).

Does that change your suggestion - because it seems wrong? :)

BigRoy:
> It doesn't work if my scene has unsaved changes.

It doesn't work, because it fails/errors/crashes? Or just because it passes the validation.
If it passes validation, that's correct.
There hasn't been a requirement to have no unsaved changes.

MustafaJafar:
> If it passes validation, that's correct.

it passes the validations if the workfile has unsaved changes.

BigRoy:
That is the current correct behavior - _if_ we want to change that it would be a new 'feature'. I personally won't be a fan. E.g. Blender integration does that and I hate it.

Anyway, does that mean we can resolve [this conversation](https://github.com/ynput/ayon-houdini/pull/81#discussion_r1742818963)?

MustafaJafar:
> That is the current correct behavior - _if_ we want to change that it would be a new 'feature'. I personally won't be a fan. E.g. Blender integration does that and I hate it.
> 
> Anyway, does that mean we can resolve [this conversation](https://github.com/ynput/ayon-houdini/pull/81#discussion_r1742818963)?

yes please, go ahead.

PR #94, opened 2024-09-10
Update "Use task handles" instance attribute label

BigRoy:
## Changelog Description

Update instance attribute label "Use Asset Handles" to "Use Task Handles"

## Additional info

This makes it match in wording with the validation itself AND the repair action's label that [reads "Disable use Task Handles"](https://github.com/ynput/ayon-houdini/blob/2e32aafd2d240eb40cfa5e5fbfe4987a4c45ea33/client/ayon_houdini/plugins/publish/validate_frame_range.py#L12-L14)

It's basically cosmetics.

## Testing notes:

1. Publish an instance
2. Toggle on the instance should make sense.

### DIFF
--- a/client/ayon_houdini/plugins/publish/collect_task_handles.py
+++ b/client/ayon_houdini/plugins/publish/collect_task_handles.py
@@ -120,5 +120,5 @@ class CollectAssetHandles(plugin.HoudiniInstancePlugin,
                     " ignore start and end handles specified in the"
                     " task attributes for this publish instance",
                     default=cls.use_asset_handles,
-                    label="Use asset handles")
+                    label="Use task handles")
         ]


MustafaJafar (commented):
I wonder if should we refactor the plugin name. and fix it in settings too. 

Will that be compatible with existent instances ? 
Edit: surprisingly, it won't affect any current instances because we use `use_handles` publish attribute.
publish attributes: 
```json
    "CollectAssetHandles": {
        "use_handles": true
    },
```

MustafaJafar (approved):
This PR works fine since we don't refactor the plugin class name therefore the settings works as usual. 

But, if you would like to test my suggested changes, please find it in my previous comment. 
which also was tested and works fine on my side. so feel free to merge the PR at its current state or after merging my suggestions.

BigRoy:
> Edit: surprisingly, yes as it won't affect any current instances because we use `use_handles` publish attribute.
> publish attributes:

Actually - this shouldn't be backwards compatible since it's stored in `publish_attributes["CollectAssetHandles"]` before but `publish_attributes["CollectTaskHandles"]` after. I feel like it's not worth it to lose backwards compatibility over. I'll merge this PR since yours continues on it anyway and we can debate more if we want to break that backwards compatibility for the small cleanup on the class name and the settings name.

PR #107, opened 2024-09-23
Load filepath to node: Use <UDIM> token

BigRoy:
## Changelog Description

Load UDIM tokens as `<UDIM>` and frame tokens from template as `$F{padding}`

## Additional info

This now uses the formatted template from the representation to define the `frame` and `udim` tokens in the full path. Hence, this should technically now also be able to support a filepath with both UDIMs and frames (although I don't think we have anything yet that publishes that)

## Testing notes:

1. Load filepath to node should still work
2. Published UDIM sequences should load with `<UDIM>` token.

### DIFF
--- a/client/ayon_houdini/plugins/load/load_filepath.py
+++ b/client/ayon_houdini/plugins/load/load_filepath.py
@@ -43,10 +43,7 @@ class FilePathLoader(plugin.HoudiniLoader):
             node.destroy()
 
         # Add filepath attribute, set value as default value
-        filepath = self.format_path(
-            path=self.filepath_from_context(context),
-            representation=context["representation"]
-        )
+        filepath = self.filepath_from_context(context)
         parm_template_group = container.parmTemplateGroup()
         attr_folder = hou.FolderParmTemplate("attributes_folder", "Attributes")
         parm = hou.StringParmTemplate(name="filepath",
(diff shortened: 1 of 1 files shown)


client/ayon_houdini/plugins/load/load_filepath.py
```diff
@@ -119,12 +120,21 @@ def format_path(path: str, representation: dict) -> str:
...
+            template: str = representation["attrib"]["template"]
+            context: dict = representation["context"]
+            if udim is not None:
+                context["udim"] = "<UDIM>"
+            if frame is not None:
+                # Substitute frame number in sequence with $F with padding
+                context["frame"] = "$F{}".format(len(frame))   # e.g. $F4
+
+            project_name: str = context["project"]["name"]
+            anatomy = Anatomy(project_name)
```
iLLiCiTiT, line 136:
```suggestion
            anatomy = Anatomy(project_name, project_entity=context["project"])
```

moonyuet (approved):
Looks good in terms of testing. 
image

BigRoy:
@iLLiCiTiT could you do a code review to see if you're ok with the code. It's functional now. :)

iLLiCiTiT (approved):

iLLiCiTiT:
Looks good.

Issue #111, opened 2024-10-03
Update LOPs loaders upon context options changing (to allow multi-shot workflows)


BigRoy:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Please describe the feature you have in mind and explain what the current shortcomings are?

Currently whenever e.g. the folder path attribute of a Load Asset LOP or Load Shot LOP refers to a Houdini context option it doesn't trigger the callback to update the representation path whenever the context option changes inside Houdini. It requires a manual refresh and this disallows multi-shot workflows.

See relevant discussion [here](https://www.sidefx.com/forum/topic/98147/?page=1#post-431289).

### How would you imagine the implementation of the feature?

According to SideFX we should be able to have the update trigger if we use expressions instead of parameter callbacks instead.

https://www.sidefx.com/forum/topic/98147/?page=1#post-431289

Here's an example scene showing that working on a simple file:

[solaris_contextoptions_python_expressions.zip](https://github.com/user-attachments/files/17246232/solaris_contextoptions_python_expressions.zip)

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the enhancement request.

### Describe alternatives you've considered:

Building complex custom callbacks - but it wouldn't work in all cases, like e.g. render clones.

PR #112, opened 2024-10-03
Load LOPs: Support update via context options (replace parm callbacks with expressions)

BigRoy:
## Changelog Description

Replace the Load LOPs parm callbacks with expressions on the file parms to support Solaris Context Options.

## Additional info

With this PR, the LOP load nodes now LIVE update based on context options. Meaning, using context-options you can now easily do multi-shot lighting!

Fixes: https://github.com/ynput/ayon-houdini/issues/111

https://github.com/user-attachments/assets/a3807f41-eb24-4da9-94b9-1a88fca840db

_Credit: Audio snippet in the background at the beginning is a few seconds of the start from this [Youtube video](https://www.youtube.com/watch?v=DeXoACwOT1o)_

#### TODO

- [x] Fix the 'warning message' remaining on the node on path update because it doesn't trigger that callback anymore
- [x] Fix the thumbnails (when enabled) not updating on representation changes.
- [x] Add "refresh thumbnail" button.

## Testing notes:

1. [x] Load Asset LOP should work consistently
2. [x] Load Shot LOP should work consistently
3. [x] Both should NOT be super slow
4. [x] When using Solaris Context Options the loader should update along live. 
    - [x] even with Edit Context Options LOP updates should be live

### DIFF
--- a/client/ayon_houdini/api/hda_utils.py
+++ b/client/ayon_houdini/api/hda_utils.py
@@ -39,12 +39,6 @@ def get_session_cache() -> dict:
     return cache
 
 
-def clear_session_cache():
-    """Clear the current session `hou.session.ayon_cache` dict"""
-    if hasattr(hou.session, "ayon_cache"):
-        del hou.session.ayon_cache
-
-
 def is_valid_uuid(value) -> bool:
     """Return whether value is a valid UUID"""
     try:
--- a/client/ayon_houdini/api/hda_utils.py
+++ b/client/ayon_houdini/api/hda_utils.py
@@ -39,6 +39,12 @@ def get_session_cache() -> dict:
     return cache
 
 
+def clear_session_cache():
+    """Clear the current session `hou.session.ayon_cache` dict"""
+    if hasattr(hou.session, "ayon_cache"):
+        del hou.session.ayon_cache
+
+
 def is_valid_uuid(value) -> bool:
     """Return whether value is a valid UUID"""
     try:
--- a/client/ayon_houdini/api/hda_utils.py
+++ b/client/ayon_houdini/api/hda_utils.py
@@ -96,7 +96,7 @@ def get_available_versions(node):
     return version_names
 
 
-def update_info(node, context):
+def set_node_representation_from_context(node, context):
     """Update project, folder, product, version, representation name parms.
 
      Arguments:
@@ -125,11 +125,6 @@ def update_info(node, context):
     }
     parms = {key: value for key, value in parms.items()
              if node.evalParm(key) != value}
-    parms["load_message"] = ""  # clear any warnings/errors
-
-    # Note that these never trigger any parm callbacks since we do not
-    # trigger the `parm.pressButton` and programmatically setting values
-    # in Houdini does not trigger callbacks automatically
     node.setParms(parms)
 
 
@@ -179,6 +174,11 @@ def get_representation_path(
 
 
 def set_representation(node, representation_id: str):
+    # For now this only updates the thumbnail, but it may update more over time
+    _update_thumbnail(node, representation_id)
+
+
+def _update_thumbnail(node, representation_id):
     # TODO: Unused currently; add support again for thumbnail updates
     if not representation_id:
         set_node_thumbnail(node, None)
@@ -189,9 +189,6 @@ def set_representation(node, representation_id: str):
         or get_current_project_name()
     )
     repre_entity = get_representation_by_id(project_name, representation_id)
-    context = get_representation_context(project_name, repre_entity)
-    update_info(node, context)
-
     if node.evalParm("show_thumbnail"):
         # Update thumbnail
         # TODO: Cache thumbnail path as well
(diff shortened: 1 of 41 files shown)


MustafaJafar (approved):
This is cool and it's live!
Animation_7

Should we care about these warnings ? 
image

Also, should we have default context options and probably manage them on AYON context change?

BigRoy:
> Should we care about these warnings ?

Yes, we should! I thought I had fixed that already but maybe I forgot to push. Will continue with this.

Will also need to decide what to do with the thumbnails. Before this PR it was created to "auto-update" on context update (but especially now with those being expressions that makes even less sence.) Mark from SideFX says it best: 
> [You should really not be "doing things" as a result of evaluating expressions.](https://www.sidefx.com/forum/topic/98147/?page=1#post-431580)

So I guess for now the easiest way is to remove how it's "live" and then just add a "Update thumbnail" button. We can think of better fixes down the line.

MustafaJafar:
Isn't this cool!
I published three products `pointcacherock_1`, `pointcacherock_2` and `pointcacherock_3`
and loading them inside a loop works! (as loops are based on context options)

_Note: I don't know if my example here is favorite in production. tbh, I don't know what is the best way to publish multiple variants of my pointcache products. maybe it's the correct workflow since different variants of an AYON product are just different AYON products._

image

BigRoy:
@MustafaJafar could you give it one last test-drive?

MustafaJafar:
on my side, 
- when I click `Reload Files` then `Update Thumbnail`, I get the following warning.
image
- when I click `Clear Cache` then `Update Thumbnail`, I get the following error: 
```python
  File "ayon::Lop/lop_import::1.0/update_thumbnail", line 1, in <module>
  File "E:\Ynput\ayon-houdini\client\ayon_houdini\api\hda_utils.py", line 189, in update_thumbnail
    repre_entity = get_representation_by_id(project_name, representation_id)
  File "E:\Ynput\ayon-launcher\.venv\lib\site-packages\ayon_api\_api.py", line 3449, in get_representation_by_id
    return con.get_representation_by_id(*args, **kwargs)
  File "E:\Ynput\ayon-launcher\.venv\lib\site-packages\ayon_api\server_api.py", line 6533, in get_representation_by_id
    for representation in representations:
  File "E:\Ynput\ayon-launcher\.venv\lib\site-packages\ayon_api\server_api.py", line 6489, in get_representations
    for parsed_data in query.continuous_query(self):
  File "E:\Ynput\ayon-launcher\.venv\lib\site-packages\ayon_api\graphql.py", line 378, in continuous_query
    raise GraphQlQueryFailed(
ayon_api.exceptions.GraphQlQueryFailed: GraphQl query Failed: Invalid entity ID None on item 'project/representations' (Line 3 Column 5)
```

client/ayon_houdini/api/hda_utils.py
```diff
@@ -155,47 +172,21 @@ def _get_thumbnail(project_name: str, version_id: str, thumbnail_dir: str):
...
 
-def set_representation(node, representation_id: str):
-    file_parm = node.parm("file")
+def update_thumbnail(node):
+    if not node.evalParm("show_thumbnail"):
+        lib.remove_all_thumbnails(node)
+        return
+
+    representation_id = node.evalParm("representation")
     if not representation_id:
```
MustafaJafar, line 181: After digging, I don't know, why the `representation_id` is equal to the lateral `"None"` after clearing the cache. 
Anyways, this works.
```suggestion
    if not representation_id or representation_id == "None":
```

BigRoy: Thanks detective! This seems to hint to another issue that I'd rather fix so I'll try and reproduce and fix it.

BigRoy: Fixed with https://github.com/ynput/ayon-houdini/pull/112/commits/8d85e2638c48089e8fb11e9166f043dbcc5710ce

It took me way longer than I want to admit to find that :') haha..

I did learn that I can quickly update the `hda` code continuously so I don't need to re-run houdini each time. I put e.g. this in a Houdini shelf:
```python
import hou
from ayon_houdini.api import lib, hda_utils
import importlib

# Enforce reload of AYON houdini libs
importlib.reload(lib)
importlib.reload(hda_utils)

# Reload HDAs
hou.hda.reloadAllFiles(rescan=True)

# Ensure the HDA module uses the reloaded libs
for node in hou.selectedNodes():
    hou.hda.reloadHDAModule(node.hdaModule())
```

MustafaJafar (commented):
Here you are some test runs: 
I like how it can be used in a for loop.
Animation_23
I ca also use it in a custom HDA.
Animation_25
Update thumbnail is buggy, to make it reliable, I need to trigger reload every time before clicking update thumbnail.
Animation_24

BigRoy:
> Update thumbnail is buggy, to make it reliable, I need to trigger reload every time before clicking update thumbnail.

Correct, update thumbnail is intentionally not live - since expressions should not be allowed to edit the graph aside of the return value from the expression.

As such, @MustafaJafar I suppose everything works? :)

MustafaJafar:
> Correct, update thumbnail is intentionally not live - since expressions should not be allowed to edit the graph aside of the return value from the expression.

yes, but I've to click to button to actually update the thumbnail.

PR #116, opened 2024-10-08
update default value of houdini_vars

MustafaJafar:
## Changelog Description
Update default value of `$JOB` in  `Update Houdini Vars on context change`

## Testing notes:
1. Launch Houdini Via AYON
2. `$JOB` should be set to the same value as before this PR.

### DIFF
--- a/server/settings/general.py
+++ b/server/settings/general.py
@@ -41,7 +41,7 @@ DEFAULT_GENERAL_SETTINGS = {
         "houdini_vars": [
             {
                 "var": "JOB",
-                "value": "{root[work]}/{project[name]}/{hierarchy}/{asset}/work/{task[name]}",  # noqa
+                "value": "{root[work]}/{project[name]}/{hierarchy}/{folder[name]}/work/{task[name]}",  # noqa
                 "is_directory": True
             }
         ]


server/settings/general.py
```diff
@@ -41,7 +41,7 @@ class GeneralSettingsModel(BaseSettingsModel):
         "houdini_vars": [
             {
                 "var": "JOB",
-                "value": "{root[work]}/{project[name]}/{hierarchy}/{asset}/work/{task[name]}",  # noqa
+                "value": "{root[work]}/{project[name]}/{hierarchy}/{folder[name]}/work/{task[name]}",  # noqa
```
BigRoy, line 44: This is not backwards compatible, because it seems that the old `{asset}` resolved to {folder[path]}`? Or didn't it?

```suggestion
                "value": "{root[work]}/{project[name]}/{hierarchy}/{folder[path]}/work/{task[name]}",  # noqa
```

Also, note that `{folder[path]}` is prefixed with a `/` so technically the `/` in front of this doubles it. So it should be:

```suggestion
                "value": "{root[work]}/{project[name]}/{hierarchy}{folder[path]}/work/{task[name]}",  # noqa
```

MustafaJafar: The folder path is equivalent for hierarchy/folder name. 
I've raised this question on Discord [here](https://discord.com/channels/517362899170230292/773127690453516298/1293236898680274995).

BigRoy: You are absolutely correct - I missed `{hierarchy}` there.
Hmm - bit confused why there was a double forward slash in the logs then.

BigRoy: Can't reproduce the slash issue myself.

BigRoy (approved):
Works for me.

Issue #13, opened 2024-06-14
Houdini: Load products in active panel instead of a container group


MustafaJafar:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Please describe the feature you have in mind and explain what the current shortcomings are?

Loading AYON products in Houdini is not much intuitive. As loaders put the loaded data in the container group instead of the current active panel.

Also, let me quote BigRoy's [comment](https://github.com/ynput/ayon-core/pull/456#discussion_r1637951239): 
```
makes me wonder a bit whether it's still worth it to put it inside that container group instead of trying to put it in an active matching panel or alike.
``` 

### How would you imagine the implementation of the feature?

we can use something like this. 
```
pane = stateutils.activePane(kwargs)
if isinstance(pane, hou.NetworkEditor):
    pwd = pane.pwd()
    pwd .createNode(node_type, node_name)

    # Imprint it manually
        data = {
            "schema": "openpype:container-2.0",
            "id": AVALON_CONTAINER_ID,
            "name": node_name,
            "namespace": namespace,
            "loader": str(self.__class__.__name__),
            "representation": context["representation"]["id"],
        }

        # todo: add folder="Avalon"
        lib.imprint(container, data)
```

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the enhancement request.

antirotor:
This needs some more concrete plan

BigRoy:
The dedicated `Load Asset LOP` has been doing this already.

Using the loader tries to find the "active LOP network", see code [here](https://github.com/ynput/ayon-houdini/blob/1046ee30dc08d94e15ba0a8f478d5b7f4934fad5/client/ayon_houdini/plugins/load/load_asset_lop.py#L24-L29).
But of course, it being a dedicated HDA already helps tons, because you can just create the "loaded container" using TAB menu and configure it fully on the HDA.

> This needs some more concrete plan

Could you elabore what part is lacking for you to plan this?

Lypsolon:
i think ondra is not wrong there a simple screenshot or mock up would be great. 
i think i know what this is about from a talk a few days back but given the description i thing in a month time i would not be able to find it out without asking around.

BigRoy:
@Lypsolon updated the description. Is it clear now?

MustafaJafar:
I think this issue should be closed as https://github.com/ynput/ayon-houdini/pull/121 serves the end goal of this issue. 
What do you think @BigRoy @antirotor ?
[...]

PR #121, opened 2024-10-14
Generic Loader OTLs

### LINKED ISSUE #118: Create generic "load" HDA that can load any product type

Similar to Load Asset LOP and Load Shot LOP implement a 'generic' HDA/OTL that exposes the same parameters to define the representation, like the Folder path and the folder path picker, etc. but then support it in SOPs.

Preferably (if even possible) it's `null`-like node that can be created in ALL contexts and would solely be used to reference the resulting filepath parm on. This way it can be used easily on ANY node in Houdini as an input filepath.

This would also make an amazing building block for studio HDAs, they can then contain this in their HDAs - benefit from all the same features of 'loading' ANY product whatsoever, etc.

Create a generic load HDA that is valid for all contexts.

_No response_

### Additional context:

_No response_

MustafaJafar:
## Changelog Description
Hopefully, resolve #118 
This PR mixes our solutions for [LOPLoadAssetLoader](https://github.com/ynput/ayon-core/pull/294) and [FilePathLoader](https://github.com/ynput/ayon-core/pull/222) resulting in the new generic loader otls.

The new otls provide the same parameters UI to load/get the file path. 
Houdini artists can then reference the relevant Houdini parm to use the exact value.

The new load otl is added to multiple locations/contexts/levels in Houdini.
- [x] Object
- [x] SOPs
- [ ] LOPs
- [ ] ...

image

## Additional Info
When generic loader is used in the loader. the generic loader will detect the active network and create the node there.
image

## Testing notes:

1. Launch Houdini via AYON Launcher
2. Load different products using the new generic loader.


MustafaJafar:
> ## HDA maintenance
> I really dislike the fact that we now need to maintain multiple separate HDA definitions for something that is (or should essentially be) a node that has no uniqueness in behavior across the different contexts. It's just a node that takes input parameters to define the result on another parameter.
> 
> It's quite cumbersome to update the HDA definitions if we need to do this per context - not impossible, but very annoying. This is just two contexts - but there may be way more.

Yeah, I think our case should be similar to the subnet nodes. since HDAs are built on top of subnets.
In asset manager, I can find multiple definitions of subnet, one for each Houdini context. 
I hope there's a solution for this.

image

MustafaJafar:
> ## UDIM path
> Currently it'll return the resolved filepath - but there are some cases where we need it to resolve to `<UDIM>` token. See [here](https://github.com/ynput/ayon-houdini/pull/107). Not required for MVP for this - but definitely high priority after.
> 
> So just noting that it may be on the todo list.

I've ported the logic from the filepath loader. we may clean it later/move the logic to lib so that both file loader and hda_utils can share it.

MustafaJafar:
With the current state, I created an HDA with the loader node inside.
I copied the parameters from the generic loader and I needed to do the following:
1. remove callbacks.
2. change the menu script in version and representation parameters to import the functions instead of using `hou.phm().`

image

MustafaJafar:
I was able to use the generic loader in a for loop to load multiple products with the same name and different postfix number (although, I had to use a python node to change the product name and trigger the callback). 
I don't know how to make it easy like https://github.com/ynput/ayon-houdini/pull/112#issuecomment-2420768027

image

BigRoy:
Thanks - I see why this commit https://github.com/ynput/ayon-houdini/pull/121/commits/bccc305d952993e55e69176d613a50cb28540181 would fix that. However, I'd love actually for somehow the menu script to stay attached to the 'inner' node in an HDA instead of also promoted.

Because, that way - one wouldn't need to promote ALL relevant attributes to the parent HDA but could even (for whatever reason) promote e.g. only "Version" or whatever. The question becomes - how do you do that?

In this case, how did you promote the attributes to the parent HDA? Did you copy/paste the parms? Or did you promote them upwards (which would then actually be referencing the inner parms I'd imagine?)

MustafaJafar:
> In this case, how did you promote the attributes to the parent HDA? Did you copy/paste the parms? Or did you promote them upwards (which would then actually be referencing the inner parms I'd imagine?)

Yes, I copied/pasted these parms to the parent node and then copied them by reference to the inner node. 
Cool thing that the parameter actions work fine. 
e.g. I copied these parms to a null node and removed their callback actions.
 
Animation_9

BigRoy:
> This commit [bccc305](https://github.com/ynput/ayon-houdini/commit/bccc305d952993e55e69176d613a50cb28540181) allows me to copy `version` and `representation` parms without errors.

Correct - but it basically means the logic runs on that node. Which is what we'd like to avoid. We want it to run on the inner node preferably so that it's still contained all as logic there. Because now you need ALL those attributes exposed on the parent HDA. I'd prefer it it could also work if e.g. we ONLY expose e.g. the `representations` attribute. 

The question is - how can we easily make it do that without needing to be super-specific when making HDAs to remap those callbacks.

MustafaJafar:
Here's a demo for my last two commits. (I wish if I were able to avoid the noise in my street :" )

https://github.com/user-attachments/assets/01980fc2-4219-428f-a5c9-017a296db8e1

MustafaJafar:
> Because now you need ALL those attributes exposed on the parent HDA. I'd prefer it it could also work if e.g. we ONLY expose e.g. the `representations` attribute.

Good catch. 
my custom HDA still works if only the representation is exposed. 
But the menu button doesn't work anymore. 
image
image

BigRoy:
> I made another test. my custom HDA has its own representation menu with hard corded values `abc`, `usd` and `exr`.
> My custom HDA still works.
> tbh, I'm not yet sure how to expose/mirror the generic loader parameters on the parent HDA while relying on the generic loader to run the logic. should it be the responsibility of the parent HDA creator? I'll keep digging around finding ideas to make it easier to expose these parameters.

The issue here is with **how** you created the HDA and exposed the parameters the subnet.

This is how it works absolutely fine:

1. Create a subnet
2. Add the Generic Loader inside
3. Go to Edit Parameters of the subnet
4. Go top left to the "From nodes" tab
5. Drag 'n' drop the parms onto your parameter list.

So drag 'n' drop e.g. from here to root:
parm1

Then the menu script becomes:
parm2

Which just means - use the inner node's menu.

So the issue was solely due to you 'copy pasting' the parameters instead of how you should be exposing the inner node's parameters upstream.

MustafaJafar:
> So basically we shouldn't need to change anything, we can keep the `phm()` calls etc.

I've reverted the changes in https://github.com/ynput/ayon-houdini/pull/121/commits/f7d719c5830aa3f54694ca396d0b12a201f9b9ef

MustafaJafar:
This commit https://github.com/ynput/ayon-houdini/pull/121/commits/baa2082c89555054186a91f31d593b26cbd129b0 replaces the list of operator paths with operator list. 

- list of operators: where I used an invisible parameter to control/modify the list of operator paths.
image
- operator list.
image


Although, I like the operator list less but it complies with mark tucker's comment on the side fx forums: 
> [You should really not be "doing things" as a result of evaluating expressions.](https://www.sidefx.com/forum/topic/98147/?page=1#post-431580)

moonyuet:
I have tested the generic loader. It's really nice as you can update and switch the context easily with the loader itself, and it shows up in the scene inventory. The updating and switching function works as expected and the file parameter changes to the path with the correct version. Overall, it is a starting point for all integration being "generic".
One thing I would think is it would be more artist-friendly if the file parameter being optional to be exposed on the generic loaders? (Like we did in nodes) Or we can have a setting in AYON (mainly for HDA) to allow users to expose the parameters, like we did in Nuke?

BigRoy:
> One thing I would think is it would be more artist-friendly if the file parameter being optional to be exposed on the generic loaders? (Like we did in nodes) Or we can have a setting in AYON (mainly for HDA) to allow users to expose the parameters, like we did in Nuke?
> Exposed parameter would be like the screenshot below.

Sorry, could you elaborate on this? I don't understand. If you're saying you want to expose additional parameters on the node from AYON settings I'm not sure I understand why we'd want that? What attributes?

My gut feeling is saying "no" we should not do that. The building block is supposed to be generic. It should preferably have the least amount of stuff that generates a pleasant user experience to pick the product to load whilst maintaining the right level of access from a Houdini perspective (e.g. allowing certain choices to be based of off references or other expressions). If someone wants the node to do more and expose more - that starts to sound like something studio specific which then should become their own HDA that contains the generic loader; not something that's driven by settings to configure the Load node itself. But I may be mistaken what you're describing to do.

moonyuet:
> > One thing I would think is it would be more artist-friendly if the file parameter being optional to be exposed on the generic loaders? (Like we did in nodes) Or we can have a setting in AYON (mainly for HDA) to allow users to expose the parameters, like we did in Nuke?
> > Exposed parameter would be like the screenshot below.
> 
> Sorry, could you elaborate on this? I don't understand. If you're saying you want to expose additional parameters on the node from AYON settings I'm not sure I understand why we'd want that? What attributes?
> 
> My gut feeling is saying "no" we should not do that. The building block is supposed to be generic. It should preferably have the least amount of stuff that generates a pleasant user experience to pick the product to load whilst maintaining the right level of access from a Houdini perspective (e.g. allowing certain choices to be based of off references or other expressions). If someone wants the node to do more and expose more - that starts to sound like something studio specific which then should become their own HDA that contains the generic loader; not something that's driven by settings to configure the Load node itself. But I may be mistaken what you're describing to do.

you didn't misunderstand me. I agree that it is more studio specific. The reason I come up with this contributes to the possibility of the HDAs containing some user-defined parameters. With the option of exposing the parameters, the users no need to always dive into the loader and reach to HDA to set all the parameters.

MustafaJafar:
The next TODO is to try creating custom HDAs that include generic loader.

MustafaJafar:
Based on our last discussion, it'd be better to open the parameters panel for the created generic loader. 
Also, it'd be cool to use better names for the generic loader, so we came up with `{filenode_node}_{fileparm_name}_loader`
I borrowed some code from #140  and here's how it looks like:

image

BigRoy:
> Based on our last discussion, it'd be better to open the parameters panel for the created generic loader. Also, it'd be cool to use better names for the generic loader, so we came up with `{filenode_node}_{fileparm_name}_loader` I borrowed some code from #140 and here's how it looks like:
> 
> image

@MustafaJafar I've made some commits to your branch - in particular it now 'replaces' the "Load filepath to node" loader.

Aside of that I can't seem to get it to work that it shows me the popup panel?

MustafaJafar:
> in particular it now 'replaces' the "Load filepath to node" loader.

Thank you.

> Aside of that I can't seem to get it to work that it shows me the popup panel?

we only need to add: 
```python
show_generic_loader_parmpanel(node)
```

MustafaJafar:
## Promoting the Generic Loader parameters
When building a custom HDA that uses the generic loader, the recommended way to promote the generic loader parameters is by   creating the parameters from nodes.
image

But, you may hit an issue like this [one](https://github.com/ynput/ayon-houdini/pull/121#issuecomment-2423317312) if you promote a parameter with action, e.g. `product` parameter. 

This issue is much similar when promoting some parameters (that have action callback) of the group sop node.
image

The issue is discussed in [SOP subnet parameter promotion issue | SideFx Forums](https://www.sidefx.com/forum/topic/85884/)
and let me quote the easiest suggested solution: 
```
The simplest is usually to promote the other parameters referenced by any such scripts as well,
making them invisible if they serve no other purpose on the promoted node. 
This is the easiest way to ensure the script is running similarly to how it does on the original node.
```

MustafaJafar:
Show the parameter editor, set it to the node path without selecting the node.
Animation_21
Animation_22

client/ayon_houdini/api/lib.py
```diff
@@ -1283,7 +1283,7 @@ def is_attached_to_node(image):
         return attached_images
 
 
-def find_active_network(category, default):
+def find_active_network(category=None, default="/obj"):
```
BigRoy, line 1286: Why was this change needed for this PR. I'd rather keep these explicitly required instead of 'falliing back' to some arbitrary default.

client/ayon_houdini/api/lib.py
```diff
@@ -1420,3 +1420,61 @@ def start_workfile_template_builder():
...
+
+    # Check if there's a floating parmater editor pane with its node set to the specified node.
+    for t in hou.ui.paneTabs():
+        if (
+            t.type() == hou.paneTabType.Parm
+            and t.isFloating()
+            and t.currentNode() == node
+        ):
+            t.setIsCurrentTab()
+            return
```
BigRoy, line 1440:
```suggestion
    # Check if there's a floating parameter editor pane with its node set to the specified node.
    for tab in hou.ui.paneTabs():
        if (
            tab.type() == hou.paneTabType.Parm
            and tab.isFloating()
            and tab.currentNode() == node
        ):
            tab.setIsCurrentTab()
            return
```

client/ayon_houdini/api/lib.py
```diff
@@ -1420,3 +1420,61 @@ def start_workfile_template_builder():
...
+
+    # We are using the hscript to create and set the network path of the pane
+    #   becasue hscript can set the node path without selecting the node.
```
BigRoy, line 1443:
```suggestion
    # because hscript can set the node path without selecting the node.
```

client/ayon_houdini/plugins/load/load_filepath.py
```diff
@@ -50,64 +25,54 @@ class FilePathLoader(plugin.HoudiniLoader):
...
+                                      node_name=node_name)
+        node.moveToGoodPosition()
+
+        # Set representation id
+        parm = node.parm("representation")
+        parm.set(context["representation"]["id"])
+        parm.pressButton()  # trigger callbacks
+
+        # Show Parameter Editor for the Node
+        show_node_parmeditor(node)
```
BigRoy, line 44: I'd say - remove this. We want this dialog only on the right click context menu from a parameter. Why?
Because if you do it via the loader - the product you wanted to load is already preconfigured.

You might say "But if I load it, it's in the avalon containers group and hard to find, etc.". Yes. But that's a separate issue not needing to be solved here: https://github.com/ynput/ayon-houdini/issues/13
```suggestion
```

client/ayon_houdini/api/lib.py
```diff
@@ -1420,3 +1420,61 @@ def start_workfile_template_builder():
...
+
+    # Create a generic loader node and reference its file parm
+    main_container = get_or_create_avalon_container()
+    
+    node_name = f"{file_parm.node().name()}_{file_parm.name()}_loader"
+    node = main_container.createNode("ayon::generic_loader", node_name=node_name)
+    node.moveToGoodPosition()
+    # Set relative reference via hscript. this way avoids the issues of `setExpression` e.g. having a keyframe.
+    hou.hscript(
+        f"""opparm -r  {file_parm.node().path()} {file_parm.name()} \`chs\(\\"`oprelativepath("{file_parm.node().path()}", "{node.path()}")`/file\\"\)\`"""
```
BigRoy, line 1478: Can we make this more readable?

BigRoy (approved):
Looks good to me. Nice work, I'd say - ready to merge!

MustafaJafar:
After using expressions instead of callbacks, This unlocked some powerful features.

e.g. I've these 3 products with different numbers
image

I can load them in a for loop. and it works either or not I put my nodes inside a custom HDA.
image
image

---

Also, 
here's the a demo for updating the version, the representation.
Animation_26
here's the a demo for updating the thumbnail
Animation_27

PR #122, opened 2024-10-14
Procedural Publishing: AYON Publish ROP

### LINKED ISSUE #11: Houdini: Combination of two extensions for loaders in pointcache product type

### Current Behavior:

In Houdini, we are publishing the pointcache in either bgeo or abc format. We have two loaders(bgeo and abc) to load the cache in different formats respectively. These cause the issues of updating or switching assets if we dont have both `bgeo` or `abc` format published. The issues also block publishing assets with the asset, which is not at the latest version, where the validator (Validate Outdated Container) would check on the outdated container to ensure the version is at the latest.

### Expected Behavior:

We expect to have just one loader plugin for pointcache product type which loads bgeo or abc format for the scene inventory tools to be concise on which format they are going to update/switch in pointcache product type

### Version

1.0.0

Windows

### Steps To Reproduce:

1. Publish pointcache(bgeo) wihout pointcache(abc)
2. Publish pointcache(abc) without pointcache(bgeo)
3. Have a new scene
4. Load the pointache product type
5. Go to Manage
6. Update or Change Versions of the assets
7. You will get the error below
```
  File "D:\ayon-core\client\ayon_core\tools\sceneinventory\view.py", line 250, in <lambda>
    lambda: self._update_containers(items, version=-1)
  File "D:\ayon-core\client\ayon_core\tools\sceneinventory\view.py", line 822, in _update_containers
    update_container(item, item_version)
  File "D:\ayon-core\client\ayon_core\pipeline\load\utils.py", line 508, in update_container
    repre_name
ValueError: Representation 'abc' wasn't found on requested version
```
or 
```
  File "D:\ayon-core\client\ayon_core\tools\sceneinventory\view.py", line 250, in <lambda>
    lambda: self._update_containers(items, version=-1)
  File "D:\ayon-core\client\ayon_core\tools\sceneinventory\view.py", line 822, in _update_containers
    update_container(item, item_version)
  File "D:\ayon-core\client\ayon_core\pipeline\load\utils.py", line 508, in update_container
    repre_name
ValueError: Representation 'bgeo' wasn't found on requested version
```

### Relevant log output:

_No response_

### Additional context:

_No response_

[cuID:[AY-4794](https://app.clickup.com/t/86by88z95)]

### LINKED ISSUE #25: Universal caching node

AYON Cache node in compliance with Houdini's vanilla filecache node.

There are some ideas: We can create 
- our own AYON filecache node
- Wrap the existing filecache nodes inside some AYON HDA
- Extend filecache node defintion and add AYON functionality

_No response_

### Additional context:

Related to `Node based houdini integration` Epic.
It was also mentioned in [Houdini's Future plan](https://community.ynput.io/t/ayon-houdini-future/812#ayon-cache-node-14) and it was discussed in [Houdini WG](https://github.com/ynput/ayon-workgroups/blob/4c04ff6b3c46fe4b2385c9ff501166eb68e4ebbe/houdini-workgroup/reports/240513-report-2.md?plain=1#L14-L34)

### LINKED ISSUE #119: Create AYON Publish ROP node

Add a generalized "AYON Publish ROP" for `/out` context that can submit itself as dependent job to Deadline with the input ROPs as dependencies and their files.

This may seem like it overlaps with #2 - but it's different in that this is about exposing our own ROP node that just "publishes upstream inputs" instead of being imprinted/embedded as logic onto ROP nodes themselves.

So, we want a ROP node in the `/out` context which takes the inputs and "publishes" those into AYON (as 'existing files').
This way, you can attach the ROP to any upstream dependencies and publish their outputs trivially.

The default deadline submitter (outside of AYON) can already take ROPs in /out and submit them each as individual dependent jobs as needed - so that logic already lives there and is not the primary issue to solve here (we can essentially copy the majority of that logic). It being a ROP in `/out` should already allows to generate a dependency graph inside of deadline to execute work as a tree. 

1. Create a custom AYON Publish ROP HDA
2. Allow it to publish the input dependencies.

(Each upstream node could become its own representation; if you need multiple products you can then position multiple AYON publish nodes in the graph for example)

---

Lyon stated:
> we want to publish a scene. 
> - first we need to generate and export all the assets. 
> - then we need to generate and export the scene. 
> - then we need to start export the render. 
> 
> this could look like this. 
> image
> 
> and other use case would be an game asset generation. 
> image
> 
> in all those cases we need to build a dependency graph recursively upwards from our Ayon-Deadline node. 

It's crucial to understand that we're not here to make this a Deadline submitter node. We want to make it an AYON Publish node - that knows _nothing_ of Deadline; and the goal is that hopefully you can then throw any other submitter to it and have it 'work' out of the box (like using Deadline's own "Deadline Submission ROP").

Create an AYON publish ROP that can publish its input streams.

1. Create an AYON Publish ROP for `/out` context
2. Allow it to publish just the first input (MVP)
3. Allow multiple upstream dependencies (nice-to-have).
4. Allow multiple AYON publish ROPs in a stream to 'break up' what files end up in which product, etc.

_No response_

### Additional context:

_No response_

Lypsolon:
## Changelog Description

Implements a special AYON Publish ROP node that publish a product from any of the files of its input ROPs. This way, any ROP can be easily published into a product type of the user's choice and it can act as a generic building block to allow publishing from within custom studio HDAs.

This PR implements a General Publish node into the `/out` context. The idea is that you will be capable to publish everything you want.

#### The approach 

A new AYON Publish ROP will publish any of its direct inputs as representations to a new product defined by the AYON Publish ROP's attributes.

With systems like deadline's own submitter from Houdini you can then publish complex graphs to Deadline without requiring AYON logic to do the farm submission itself.

#### Implementation.

- A DriverRop node is used with pre render scripts 
- We avoid problems with the SOHO system by having an single SOHO py file with a simple pass 
- We use the Pre-Render script to call a function on the HDA python module. This function only exists to call the `ayon_publish()` function in the ayon_houdini module to keep most code in the AYON codebase instead of the HDA.
- The `ayon_publish` function calls publish with a selection of pyblish plugins
    - TODO: We may want to avoid doing this too much.
- When publishing from the GUI we will call the `.render()` function on the given `ayon_publish` node. In case we publish from a node we will set a variable to bypass the `.render()` call because the nodes will call it themselves. 

#### Known issues

- Connecting multiple nodes that publish the same file extension this will be registered as separate representations and thereby error because the representations would overlap/overwrite each other.
- When you publish via AYON Publish nodes this will not increment your scene file because if we increment the `$HIP` variable gets out of sync because upstream nodes will export with the different file version as downstream nodes. 
- When using the OpenGl node you will need to set an output image (`ip` or `md` does not work because it writes no files) 
- Deadline submissions allows you to run every node on its set node ranged instead of respecting the deadline node range. (this is different than Houdini's behavior)
    - - in Houdini if the last node is on single frame all upstream nodes will be set to export only a single frame 
    - our system will find out what the maximum is your node can export then it will check whats on disk and publish this  
        - This avoids publishing "extra" files that should not be there  
        - In case your node right now only publishes 1 frame but did publish 10 frames a min ago and the range is still set to 10 frames this system will end up publishing the extra 9 frames even if they don't belong there. 

### Local publish

1. Launch Houdini via AYON.
2. Create the AYON Publish ROP
3. Connect some unique input file types (different extensions)
4. Each input would become a representation of the published product
5. Publish should work locally

### Farm publish

It should work also completely on the farm when customizing some logic in Deadline's submitter, [see the changes here](https://community.ynput.io/t/load-ayon-plugin-in-deadline-when-submitting-from-houdini-manually/1864/7?u=bigroy).

### Complex dependency graphs

A complex dependency graph in `/out` should also work fine. See the example files shared [here](https://github.com/ynput/ayon-houdini/pull/122#issuecomment-2447122892).
 
---

# TODO 

- [x] ensure that the hda has all the attributes 
- [x] Clean up the parameter layout for the ROP.
    - [x] Scripts TAB should be at the bottom like the other ROPs
       - [x] can they just be hidden ? 
    - [x] The SOHO parms should be invisible parms.
    - [ ] When rendering the ROP and an error occurred it would be nice to display the error on the ROP itself, like it turns red, etc. likely by raising an error explicitly in the code or by calling `node.addError`? (Do we need to have our own flag error event for this as we need to also checked on the error during cooking too.)
    - [ ] Make it very easy to select product types on the ROP or the publisher UI
    - [ ] Make sure publishing via publisher UI does not error on `model` validations if publish model product type (this may required a separate PR to differentiate the relevant validators for certain ROPs to target the ROPs instead once https://github.com/ynput/ayon-core/pull/994 is merged.
    - [x] Do not store `AYON_productName` on the instance, but compute it at runtime from creator + variant, etc.
    - [ ] Allow to get the JSON publish report from the 'headless' publish (unfortunately this requires a feature in ayon-core first: https://github.com/ynput/ayon-core/issues/946

### Use cases

    - [ ] Showcase/demo how this could be used with e.g. a `FileCache` type node, including custom ones by a studio (we can test with the one from SideFX LABs)


client/ayon_houdini/standalone/ayonpub.py
BigRoy, line 1: This should likely just end up in `ayon_houdini.api.***` somewhere there. Let's not define new custom locations.

client/ayon_houdini/hooks/set_paths.py
```diff
@@ -15,4 +17,6 @@ def execute(self):
             self.log.warning("BUG: Workdir is not filled.")
             return
 
+        addon_dir = os.path.dirname(os.path.dirname(__file__))
```
BigRoy, line 20: Unrelated change. Can be removed? Please revert changes to this file?

client/ayon_houdini/hooks/set_paths.py
```diff
@@ -6,6 +6,7 @@ class SetPath(PreLaunchHook):
 
     Hook `GlobalHostDataHook` must be executed before this hook.
     """
+
```
BigRoy, line 9: I know, cosmetics - but can you revert this too. So that less files are changed.

client/ayon_houdini/standalone/ayonpub.py
```diff
@@ -0,0 +1,45 @@
...
+
+def get_upstream_nodes(node: hou.Node):
+    # Retruns the upstream nodes to a given node
+    return [i_node.inputNode() for i_node in node.inputConnections()]
+
+
+from ayon_houdini.standalone import ayonpub
+
+
+def get_us_node_grapth(node: hou.Node, root_node: hou.Node = None):
```
BigRoy, line 12: Typo, `grapth` -> `graph`

client/ayon_houdini/standalone/ayonpub.py
```diff
@@ -0,0 +1,45 @@
...
+    return node_info
+
+
+def print_grapth(graph: dict, depth: int = 0):
+    for k, v in graph.items():
+        print("-" * depth + k.path())
+        print_grapth(v, depth=depth + 1)
+
+
+def ayon_pub_command():
```
BigRoy, line 41: Would be good not to abbreviate this - just call it `ayon_publish` instead of `ayon_pub`

client/ayon_houdini/startup/otls/AyonPub.1.1.hda/Driver_1AyonPub_8_81.1/CreateScript
```diff
@@ -0,0 +1,15 @@
+# Automatically generated script
+\set noalias = 1
+#
+#  Creation script for AyonPub::1.1 operator
```
BigRoy, line 4: Can we update the full HDA to be just `AYONPublish` or `ayon_publish` so that it matches the naming conventions of the other HDAs in the `otls/` folder.

client/ayon_houdini/startup/otls/AyonPub.1.1.hda/Driver_1AyonPub_8_81.1/DialogScript
```diff
@@ -0,0 +1,690 @@
...
+	""
+    }
+
+    inputlabel	1	"Input #1"
+    inputlabel	2	"Input #2"
+    inputlabel	3	"Input #3"
+    inputlabel	4	"Input #4"
+    inputlabel	5	"Input #5"
+    inputlabel	6	"Input #6"
+    inputlabel	7	"Input #7"
```
BigRoy, line 18: Oh lord. Can we clean this up? Or do we need this for all the inputs? :') (Seems odd?)

client/ayon_houdini/startup/otls/AyonPub.1.1.hda/Driver_1AyonPub_8_81.1/DialogScript
```diff
@@ -0,0 +1,690 @@
...
+    inputlabel	497	"Input #497"
+    inputlabel	498	"Input #498"
+    inputlabel	499	"Input #499"
+    inputlabel	500	"Input #500"
+
+    parm {
+        name    "soho_program"
+        label   "soho_program"
+        type    string
+        default { "ayon_pub.py" }
```
BigRoy, line 517: Let's also rename this `ayon_publish`?

client/ayon_houdini/standalone/ayonpub.py
BigRoy, line 1: Let's move this from `ayon_houdini.standalone` into `ayon_houdini.api` like most of the other code. Feel free for now to keep it in e.g. a `ayon_houdini.api.ayon_publish` if that's easier for now. Or in `ayon_houdini.api.hda_ayon_publish`.

BigRoy (commented):
Some mostly cosmetic notes - will test out functionality tomorrow.

client/ayon_houdini/startup/otls/AyonPub.1.1.hda/Driver_1AyonPub_8_81.1/DialogScript
Lypsolon: i did it the way that the deadline node dose it because if you dont set it to 9999 inputs you "limit" the amount of nodes that can be connected to the connection block. 
the block atop of a node is actually just the inputs in a group representation. 

that's how i understand it.

client/ayon_houdini/standalone/ayonpub.py
Lypsolon: okay then i clean it up and put it into api.

Lypsolon: [For Reference]
the problem right now is that when changing the code in the hda foulder the hda complains about code that was changed (it looks like its loading the old code needs checking tomorrow)

Lypsolon: done

client/ayon_houdini/hooks/set_paths.py
Lypsolon: done by copy pasting the original file back in from develop.

client/ayon_houdini/startup/otls/AyonPub.1.1.hda/Driver_1AyonPub_8_81.1/CreateScript
Lypsolon: updated it to ayon_publish to be like the others

client/ayon_houdini/startup/otls/AyonPub.1.1.hda/Driver_1AyonPub_8_81.1/DialogScript
Lypsolon: image
this happens when you do 1 input. 
so if we want to allow the connection of multiple upstream nodes like the other rpos do we will need them like this. 

i guess we could reduce the amount but tbh i dont think this is gonna win us anything. 

just to mention it i believe multiple upstream nodes connected like a merge node is a nice convenience feature for users otherwise they would probably just plug a merge node into our node.

client/ayon_houdini/standalone/ayonpub.py
Lypsolon: i put it into api.ayon_publish for now.

client/ayon_houdini/standalone/ayonpub.py
Lypsolon: done

client/ayon_houdini/startup/otls/AyonPub.1.1.hda/Driver_1AyonPub_8_81.1/CreateScript
Lypsolon: done

client/ayon_houdini/standalone/ayonpub.py
Lypsolon: done in external commit.

BigRoy:
@Lypsolon I checked out your PR.. decent start. I did some tweaking as I tested it: https://github.com/ynput/ayon-houdini/pull/127

We'll need to keep in mind some other goals to clean things up down the line:

- [ ] Clean up the parameter layout for the ROP.
    - [ ] Scripts TAB should be at the bottom like the other ROPs
    - [ ] The SOHO parms should be invisible parms.
    - [ ] When rendering the ROP and an error occurred it would be nice to display the error on the ROP itself, like it turns red, etc. likely by raising an error explicitly in the code or by calling `node.addError`?

client/ayon_houdini/startup/otls/AyonPub.1.1.hda/Driver_1AyonPub_8_81.1/DialogScript
BigRoy: Stupid Houdini I suppose - why can't we just set num inputs to 1000 but skip the labels. Well, fine with me 🤷

Lypsolon:
soo a bit later we got a little system that exports an advanced character via deadline. 
i will append a scene that i used for experimentation. 

its important to know that this system dose not support Ayon submit to farm right now as the submit to farm dose not generate the dependent jobs. 

also the system will run Jobs for all the nodes in the Graph (as Deadline usually dose) this is nice because upstream nodes are executed and if we would do things like Run node A -> load the output -> export the loaded data this should work. 
but i have the feeling that things are run multiple times (i have no prove for this but that's what i think) 

image


[UT_Roy_Test_Generic_v028.zip](https://github.com/user-attachments/files/17471946/UT_Roy_Test_Generic_v028.zip)

MustafaJafar (commented):
Hello, 
I've some cosmetic notes to be consistent with the current AYON HDAs if you don't mind. 
1. This HDA file `ayon-houdini\dell\ynput.dev.ayon_publish.1.6.hda`, I think you can remove it from this PR. You can keep it on your side if you are using it, but can just exclude it  when using `git add` 
    ```
    git add .
    git restore --staged dell/ynput.dev.ayon_publish.1.6.hda
    ```
2. when saving/creating an HDA for ayon-houdini, we 
- set the menu Entry to `AYON`
- set the Author to `ayon`
- set version to `1.0`. currently, we don't change the version unless we have made a non backward compatible changes. (actual version control is done explicitly with git)
- don't use branches or adding them to the HDA name.
- add the Houdini context to the node. so, the name should be `ayon_rop_publish`
image

Lypsolon:
>     1. This HDA file `ayon-houdini\dell\ynput.dev.ayon_publish.1.6.hda`, I think you can remove it from this PR. You can keep it on your side if you are using it, but can just exclude it  when using `git add`
>        ```
>        git add .
>        git restore --staged dell/ynput.dev.ayon_publish.1.6.hda
>        ```

forgot to remove it. did.

>     2. when saving/creating an HDA for ayon-houdini, we
> 
> 
>     * set the menu Entry to `AYON`
> 
>     * set the Author to `ayon`
> 
>     * set version to `1.0`. currently, we don't change the version unless we have made a non backward compatible changes. (actual version control is done explicitly with git)
> 
>     * don't use branches or adding them to the HDA name.
> 
>     * add the Houdini context to the node. so, the name should be `ayon_rop_publish`

done.

client/ayon_houdini/startup/otls/ayon_rop_publish.hda/ynput_8_8dev_8_8Driver_1ayon__publish_8_81.7/Tools.shelf
```diff
@@ -0,0 +1,20 @@
...
+ a single file. -->
+
+  <tool name="$HDA_DEFAULT_TOOL" label="$HDA_LABEL" icon="$HDA_ICON">
+    <toolMenuContext name="viewer">
+      <contextNetType>ROP</contextNetType>
+    </toolMenuContext>
+    <toolMenuContext name="network">
+      <contextOpType>$HDA_TABLE_AND_NAME</contextOpType>
+    </toolMenuContext>
+    <toolSubmenu>Digital Assets</toolSubmenu>
```
MustafaJafar, line 15:
```suggestion
    <toolSubmenu>AYON</toolSubmenu>
```

MustafaJafar (commented):
Here's a better reference to make the HDA naming more consistent. 
(I don't have any strong opinion about it. I'm just writing about the current naming convention in other OTLs.)
image

MustafaJafar:
Hey, could you add some additional info to the PR description for future refence?  
e.g. 
- References like [SOHO | SideFx Docs](https://www.sidefx.com/docs/houdini/render/soho.html). 
- Also, mention the benefits of using soho.

MustafaJafar:
I gave it a simple test, this one works. (but it only publishes the first frame)
image

and this one raises an error.
```
ayon_core.pipeline.publish.publish_plugins.KnownPublishError: Transfer to destination is already in queue: \\storage\work\ayon\RnD\Ideas\work\Generic_Publisher\geo\rnd_Ideas_Generic_Publisher_v005.Cube.1001.bgeo.sc -> \\storage\work\ayon\RnD\Ideas\publish\pub\pubTestCube\v007\rnd_Ideas_pubTestCube_v007.bgeo.sc. It's not allowed to be replaced by a new transfer from \\storage\work\ayon\RnD\Ideas\work\Generic_Publisher\geo\rnd_Ideas_Generic_Publisher_v005.Sphere.1001.bgeo.sc
```
image

so, a question, do multiple inputs considered as different products ? or should they considered as one product with multiple representation ? 

The current status of this PR: we have one product: `product_type`-> `pub` and `product_name` -> `{product_type}_{variant_name}`.

Lypsolon:
im gonna start with this one. 


> I gave it a simple test, this one works. (but it only publishes the first frame) image

this only publishes the first frame because the default behavior of the /out context is to use what ever is set in the last node. aka if you AyonPub node is set to single frame everything above will be single frame. (you can test this behavior with multiple geo publishes in a row set the upper ones to a frame range and the last one to a single frame) 

this behavior dose not apply to null nodes or nodes that dont have a frame range to set. 
but those nodes also dont have a pre render script. 

its also important to know that when you use the deadline default submitter it allows you to export the frame ranges from the different nodes. we currently support this by checking the outer bounds of what a node can publish and then checking how much of this is actually on disk. 
this runs the risk that if something did not export correctly our node will ignore instead of erroring, but we cant get the correct range from the node because this can be changed at runtime. 


> and this one raises an error.
> 
> ```
> ayon_core.pipeline.publish.publish_plugins.KnownPublishError: Transfer to destination is already in queue: \\storage\work\ayon\RnD\Ideas\work\Generic_Publisher\geo\rnd_Ideas_Generic_Publisher_v005.Cube.1001.bgeo.sc -> \\storage\work\ayon\RnD\Ideas\publish\pub\pubTestCube\v007\rnd_Ideas_pubTestCube_v007.bgeo.sc. It's not allowed to be replaced by a new transfer from \\storage\work\ayon\RnD\Ideas\work\Generic_Publisher\geo\rnd_Ideas_Generic_Publisher_v005.Sphere.1001.bgeo.sc
> ```
> 
> image
> 
> so, a question, do multiple inputs considered as different products ? or should they considered as one product with multiple representation ?
> 
> The current status of this PR: we have one product: `product_type`-> `pub` and `product_name` -> `{product_type}_{variant_name}`.



yes this errors because every publish node you place is its own product. and the inputs get to be representations. 

this is just the way this feature was planed because it makes control over the node graph easier.

Lypsolon:
> Hey, could you add some additional info to the PR description for future refence? e.g.
> 
>     * References like [SOHO | Houdini Docs](https://www.sidefx.com/docs/houdini/render/soho.html), [SOHO | HDK Docs](https://www.sidefx.com/docs/hdk/_h_d_k__s_o_h_o.html).
> 
>     * Also, mention the benefits of using soho.
> 
> 
> Please, correct me.
> ## Additional Info: About SOHO
> 
> SOHO describes **What Happens When The Render Button Is Hit?** SOHO is invoked through a SOHO output driver. A SOHO output driver is an OTL with some specific parameters created to control the behavior of the output driver.
> 
> SOHO comes with standard callback functions, or _hooks_, `pre-render`, `pre-frame`, `post-render` and `post-frame`.
> 
> For this prototype, we are using SOHO to leverage the power of these parameters: Currently, we trigger/run the `ayon publish logic` (`hou.phm().run()`) using `pre-render` script. _Yes, I've re-arranged the parameters to take this screenshot._
> 
> For more info, please refer to [SOHO | Houdini Docs](https://www.sidefx.com/docs/houdini/render/soho.html) and [SOHO | HDK Docs](https://www.sidefx.com/docs/hdk/_h_d_k__s_o_h_o.html)

so having a lot of documentation about soho is a bit over the top if you ask me. 

hmm we can add a bit if its needed sure but i use the description for too-do and keeping it short makes my live easier so lets keep it in a comment. 

also you talk a lot about SOHO but we only need it because we have a output driver. 
the content of the soho script is just `pass` and we use the pre-render callback to call a python function in ayon-houdini.


> I wonder if could we have a `Driver` tab in compliance with mantra node. which can be used to perform headless publishes? I'm not sure if it's a good idea, but just sharing food for thought. 


hmm i dont know if we need this tbh. 
the main target that roy gave me is to work with the deadline driver and this thing already publishes in the background. 

also we need the node graph to be intact for our operation. aka a .hou file needs to be some where so we can get the current state of the node graph, while the mantra node only needs the geo scene and can just load it into memory. 

with that in mind i will place it late onto the wanted list ?

Lypsolon:
i and @BigRoy talked last week about the question if depended network publishes might be useful. i said yes he said show here we go. 

lets say you have a pipeline like this. 
you have Houdini artists that work mainly in sops (because that's nice) but your pipeline for shading lighting etc is in USD. 
now you have a simulation lets say sand or stones on the ground, and you have a second artist that will scatter plants on this sim (or simulate the impact of the moving stones on those plants) for this its great not to deal with USD because your usd file will probably use an instance to store only points with the sim data on-top and de-referencing usd in sops is heavy if you have an instancer. 

with that in mind we dont want to simulate 2 times so having the sim triggered before we trigger the usd export is efficient. we could probably use a file Cache but this setup is nice and small. 

image
image
image

MustafaJafar:
> hmm we can add a bit if its needed sure but i use the description for too-do and keeping it short makes my live easier so lets keep it in a comment.

Personally, I'm fine as long as the info is reachable.
Maybe, we can adjust the PR later to be consistent with other PRs.. where we have some sections like: changelog description. how the feature works. demo, testing note.. this is important for future reference as we may need in the future.
 
e.g. this falls under how the feature works
```
also you talk a lot about SOHO but we only need it because we have a output driver.
the content of the soho script is just pass and we use the pre-render callback to call a python function in ayon-houdini.
```

Lypsolon:
> 
> Personally, I'm fine as long as the info is reachable. Maybe, we can adjust the PR later to be consistent with other PRs.. where we have some sections like: changelog description. how the feature works. demo, testing note.. this is important for future reference as we may need in the future.
> 
> e.g. this falls under how the feature works
> 
> ```
> also you talk a lot about SOHO but we only need it because we have a output driver.
> the content of the soho script is just pass and we use the pre-render callback to call a python function in ayon-houdini.
> ```

yes we should add it up there but not while its in Draft currently this PR is just a work space. 
lets stop treating it as something to be for reference lets treat it like a branch whit a node pad. 

lets not get caught up in the idea that this pr is gonna get testing nodes or a change log right now. 
we dont know what we will change this is a Draft not a PR. 

also i dont treat a Draft PR like a Post or Documentation draft. a draft PR for me is a place to work on a feature that will be as dirty as it needs to be and when the feature is cleaned up its decided this PR is gonna go Live then and only then i will clean up the Description and add Testing notes and a change log. 
if i would do this right now i would need to add to the change log every time i push, and with every idea and feature that we change the testing notes would need to change. 
this would take for ever.

MustafaJafar:
> yes this errors because every publish node you place is its own product. and the inputs get to be representations.

like this ?
image

Lypsolon:
there was a debate about how to handle temp data. 

this is an image where both sides first need to export some temp data and then publish a variation of that. 

the question is what side is nicer to use. 

left side: 
- only publishes the first input it has. 

right side:
- publishes everything that it has as an input until it hits the breaker node. 

image

BigRoy:
> the question is what side is nicer to use.

There is an issue in your explanation of the left side because it reads like it only ever uses one input, but instead it publishes **all** direct inputs (not just the first direct input - it just doesn't recurse up the graph further and only considers direct input).

I think the right is visually nicer - but it also adds more complexity in us needing to 'manage' the dependency tree more. Plus I'm worried about the cases where a publisher breaker is in there, but a connection accidentally also 'passes' it around in complex graphs. As soon as we get to more complex graphs I think it may be visually nicer to just see whatever inputs are to the publish ROP itself - without not needing to care about anything else. It makes our ROP more stupid and only care about its direct input which I think from a coding perspective is actually a good thing too. (Plus requires less custom "layer break" nodes and whatnot)

@antirotor @MustafaJafar on those two graphs, thoughts?

MustafaJafar:
> It makes our ROP more stupid and only care about its direct input which I think from a coding perspective is actually a good thing too. (Plus requires less custom "layer break" nodes and whatnot)

I agree with you so the one the left should be simpler. 
But, I still have some problems understanding the intended flow. e.g. how to interpret the graph? 

**As far as I understood,** publisher node resembles a product, its input resemble representations. 
e.g. this should be my `testCube` product, its type is `pointcache`.  it has two representations: `bgeo` and `abc`. also, it has an additional review representation. 
image

In this context, how can I interpret the graph on the left? 
knowing that the graph is equivalent to: 
image

So, How can we tell the difference between: 
- Rop that I want to run at a particular time with in the node graph.
- Representations to publish.
- Auxiliary files that should be included to a representation as they are referenced by that particular representation. (but, they are generated by a separate Rop)

Lypsolon:
> So, How can we tell the difference between:
> 
>     * Rop that I want to run at a particular time with in the node graph.
> 
>     * Representations to publish.
> 
>     * Auxiliary files that should be included to a representation as they are referenced by that particular representation. (but, they are generated by a separate Rop)

let me try to answer this. 
the order off compute will allays be handled by Houdini. 
aka in my left example only the direct inputs are exported but because the Usd export is depended on the Geo export the geo export will be run first and not in parallel. 
so the answer to Q1 is Houdini dose it for us if we trigger .render() on a downstream node. 

Q2
the representations to publish is decided by the file extension. aka you will get a publish product (that currently has no validation etc) and every node gets to be an representation and if you did 2 nodes with the same file type you get an error. 
i believe this is in line with the design of ayon. 

ps i dont think there is a good way to say if 2 nodes have the same file type we add it to the same representation, i imagine this to be very confusing.  

Q3 with this one im unsure what you mean tbh so i cant fully answer it right now sorry.

Lypsolon:
> Also, when computing the dependency graph, I'd like to stick the default Houdini behavior. without adding any additional logical relations to it. So, these two graph should give the same results.
> 
> image


the thing is those 2 graphs dont give you the same result because the calculation is ordered differently. 

on the left:
the usd publish can run in parallel with the geo publish (in deadline it would be, in Houdini it depends on how parallel things get handled) 

on the right: 
the usd publish can only run after the geo publish. 

this is because Houdini dose not see the graph from the last node but from a dependency standpoint. 
aka on the right usd rop is depended on geo rop so it gets blocked until geo rop is done. on the left its not so it can run when it wants. 
this is a problem if your usd export imports the geo export.

Lypsolon:
> There is an issue in your explanation of the left side because it reads like it only ever uses one input, but instead it publishes **all** direct inputs (not just the first direct input - it just doesn't recurse up the graph further and only considers direct input).

your right my bad. 


> I think the right is visually nicer - but it also adds more complexity in us needing to 'manage' the dependency tree more. Plus I'm worried about the cases where a publisher breaker is in there, but a connection accidentally also 'passes' it around in complex graphs. As soon as we get to more complex graphs I think it may be visually nicer to just see whatever inputs are to the publish ROP itself - without not needing to care about anything else. It makes our ROP more stupid and only care about its direct input which I think from a coding perspective is actually a good thing too. (Plus requires less custom "layer break" nodes and whatnot)
> 
> TL;DR - my vote goes for left.

honestly after playing around a bit i think left is a bit simpler not having an extra node to handle is a good thing. 
but i also think it might be more complex in explanations? but this is just a feeling i find the right one simpler to understand (but this might be because i implemented it) 

with that in mind if the left one actually covered all grounds and we don't forget something i will vote for it too. 
it will make the code a lot simpler (but we will need to rewrite / reduce a good few functions)

BigRoy:
> Auxiliary files that should be included to a representation as they are referenced by that particular representation. (but, they are generated by a separate Rop)

This is a good question. But:
1. We not have a means to detect these from _any ROP_ regardless so would mean more knowledge on those files regardless. So how we specify those would automatically become a separate problem too. I wouldn't bother yet.
2. I believe those 'additional files' on a publish are not stored per representation but are currently shared across all representations in an AYON version - since a representation does not have the notion of auxiliary files. As such, even if we would detect it per representation I don't think there's a 1:1 notion of that in AYON itself currently.

Lypsolon:
soo a new commit has been done and now we only accept direct inputs. 
i removed everything we dont need (maybe that is) 

image

Lypsolon:
A follow up Comment with 2 Test Sceenes. 
1 For the more "Complex" test roy gave me 
image

and 1 for the 3 simple ones. 
image


[Hou_Rop_Pub_oct_30A.zip](https://github.com/user-attachments/files/17573160/Hou_Rop_Pub_oct_30A.zip)

Lypsolon:
Regarding Raising errors on the node. 

in Houdini if you want to Raise an error on a node you do `raise hou.NodeError()` this works only if we are in an execution that is on the node. 
but with piblish we are not on the node anymore. further more errors are Shadowed/Handled by piblish so they never get to the node. 
i assume we would need some kind of a callback or similar to get back into the nodes unit of execution to raise the error in the correct place to achieve this.

client/ayon_houdini/api/ayon_publish.py
```diff
@@ -0,0 +1,133 @@
...
+        if "IncrementCurrentFile" not in str(plugin)
+    ]
+
+    for result in pyblish.util.publish_iter(pyblish_context, filtered_plugins):
+        log.debug(result)
+        for record in result["records"]:
+            log.debug("{}: {}".format(result["plugin"].label, record.msg))
+
+        # Exit as soon as any error occurs.
+        if result["error"]:
```
moonyuet, line 57: Wouldn't it raise any KeyError if there is no `error` inside the `result` dict?

client/ayon_houdini/plugins/publish/collect_ayon_publish_outputs.py
```diff
@@ -0,0 +1,97 @@
...
+                    "we can not detect its output files."
+                )
+                continue
+
+            # Split extension, but allow for multi-dot extensions
+            ext = lib.splitext(
+                filepaths[0],
+                allowed_multidot_extensions=self.multidot_extensions,
+            )[-1]
+            ext_no_dot = ext[1:]
```
moonyuet, line 82:
```suggestion
            ext= ext.strip(".")
```
or
```suggestion
           ext = lib.splitext(
                filepaths[0],
                allowed_multidot_extensions=self.multidot_extensions,
            )[-1].strip(".")
```

client/ayon_houdini/plugins/publish/collect_ayon_publish_outputs.py
```diff
@@ -0,0 +1,97 @@
...
+
+            # representation files must be single `str` if single file
+            # and also must be the filename, not full path
+            filenames = [os.path.basename(file) for file in filepaths]
+            if len(filenames) == 1:
+                filenames = filenames[0]
+
+            representation = {
+                "name": ext_no_dot,
+                "ext": ext_no_dot,
```
moonyuet, line 92: If you make change on this https://github.com/ynput/ayon-houdini/pull/122/files#r1900672175
```suggestion
                "name": ext,
                "ext": ext,
```

moonyuet:
I have tested with pubish node <- really cool first of all. We hit some errors for getting product_type from the instance data.
We should set `product_type` as `instance.data` in the creator or use instance.data["productType"] inside the collect_output_node.py
image

PR #124, opened 2024-10-16
Enable use selection by default on creators

BigRoy:
## Changelog Description

Enable "use selection" by default on creators (pre create attribute)

## Additional info

Why?

1. New users expect this behavior.
2. It matches most other host integrations, like blender, maya, cinema4d, aftereffects, substance painter, resolve

Do note that `ayon-nuke` also has it disabled by default.
Also, the legacy creator (long time ago) also had it enabled by default.

## Testing notes:

1. Open Publisher UI on create tab...
2. Creators should default with "Use Selection" checkbox enabled.


MustafaJafar (approved):
It works and I like it!  
Indeed it's the expect typical behavior.

PR #125, opened 2024-10-17
Create Creator via TAB with null, keep null selected

BigRoy:
## Changelog Description

When creating a ROP using a Creator from the TAB menu in the node network it'd select the ROP node after placing down the Null. This was quite confusing because it'd take you out of your working area UNLESS you had your current panel "pinned" explicitly.

With this change, we instead keep the Null selected (since it'll be preconfigured on the ROP anyway) so that you can still continue to work where you were and deal with the ROPs later.

## Additional info

This may be nice to expose in settings to 'toggle' the behavior. But it may actually be more of a user preference than studio preference?

## Testing notes:

1. Go to /stage
2. Press TAB
3. Use `Create USD`

image

4. Place the null.
5. The null should remain selected.

### DIFF
--- a/client/ayon_houdini/api/creator_node_shelves.py
+++ b/client/ayon_houdini/api/creator_node_shelves.py
@@ -89,6 +89,7 @@ def create_interactive(creator_identifier, **kwargs):
     #   to a method on the Creators for which this could be the default
     #   implementation.
     pane = stateutils.activePane(kwargs)
+    is_null_created = False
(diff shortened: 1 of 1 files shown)


MustafaJafar (approved):
I'm keep forgetting that I've already reviewed this PR.
I like this PR and it works. we may add the flash messages in another PR. 
maybe we can add them to the `create_interactive` function or add them to the `HoudiniCreator` so it works by default (maybe we can add a setting for in the Generic Settings section).

PR #126, opened 2024-10-20
Add a render resolution validator for USD Renders

BigRoy:
## Changelog Description

Add a render resolution validator for USD renders.

## Additional info

image

The Jump To Editor Node action behaves similar to the "Editor Nodes" Houdini is able to show as metadata on the primitives or attributes in the Scene Graph Details (or Geometry Spreadsheet in Hou 20+) in Solaris.

image

## Testing notes:

1. Set up USD render publish from Houdini with differing resolution and pixel aspect ratio than current context.
2. It should be invalidated.
3. The validator is optional, so should be possible to be ignored.
4. If the resolution is correct - it should of course come through without errors.

### DIFF
--- a/.github/workflows/release_trigger.yml
+++ b/.github/workflows/release_trigger.yml
@@ -2,10 +2,23 @@ name: 🚀 Release Trigger
 
 on:
   workflow_dispatch:
+    inputs:
+      draft:
+        type: boolean
+        description: "Create Release Draft"
+        required: false
+        default: false
+      release_overwrite:
+        type: string
+        description: "Set Version Release Tag"
+        required: false
 
 jobs:
   call-release-trigger:
     uses: ynput/ops-repo-automation/.github/workflows/release_trigger.yml@main
+    with:
+      draft: ${{ inputs.draft }}
+      release_overwrite: ${{ inputs.release_overwrite }}
     secrets:
       token: ${{ secrets.YNPUT_BOT_TOKEN }}
       email: ${{ secrets.CI_EMAIL }}
--- a/client/ayon_houdini/api/creator_node_shelves.py
+++ b/client/ayon_houdini/api/creator_node_shelves.py
@@ -89,6 +89,7 @@ def create_interactive(creator_identifier, **kwargs):
     #   to a method on the Creators for which this could be the default
     #   implementation.
     pane = stateutils.activePane(kwargs)
+    is_null_created = False
     if isinstance(pane, hou.NetworkEditor):
         pwd = pane.pwd()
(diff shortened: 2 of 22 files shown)


moonyuet (commented):
When the render does not match the resolution it would error out as expected. And if I fixed with the correct resolution, it would pass the validation
image
But I do have one question on the editor node(a very dumb one).
When I click on the editor node, it jumps to the nodes of the karma render setting. Is it intended?
image

BigRoy:
> But I do have one question on the editor node(a very dumb one).
> When I click on the editor node, it jumps to the nodes of the karma render setting. Is it intended?

Perfect - thanks for asking, not dumb at all. Technically it's correct; that is actually the node responsible inside that HDA for "writing" that opinion in the USD graph in Solaris. Very nice to see that works as nicely as it does in Houdini.

Anyway, we may want to - if the node happens to be inside a locked HDA - instead find the first 'editable' parent node that is not INSIDE a locked HDA.

I'll look into it! :)

BigRoy:
@moonyuet Thanks again - it should now correctly select the editable node - in this case the Karma Render Settings LOP node.

moonyuet (approved):
The validator blocks from publishing as expected when the resolution does not align with the db. 
When users click on `jump to editor node`, it would guide the users to the karma rendersetting as expected so that they can reset resolution setting there. Once they reset the resolution accordingly to the db, it would pass the validation.
image

MustafaJafar (approved):
It works.
image

BigRoy:
Thanks - I just want to still add a setting for this validator so it could be disabled if a studio hates it. 🗡️

moonyuet:
> Thanks - I just want to still add a setting for this validator so it could be disabled if a studio hates it. 🗡️

Would you mind also adding repair action for fixing the resolution too?

BigRoy:
> Would you mind also adding repair action for fixing the resolution too?

Unfortunately I'd say this is non-trivial and in the majority of cases also not 'repairable' in the way a user would like it. I could technically add a edit attribute into the graph below the "Editor Node" but usually you don't want to clutter your solaris graph with that - but intead want to specify the resolution on the render settings node. The issue there is that that node is not a standard one. It may be set on:
- Karma Render settings LOP
- Arnold Render Settings LOP
- Redshift Render Settings LOP
- Render Settings LOP
- Render Product LOP

It may even have been set by an Attribute Wrangle LOP or Python LOP or even not set at all in the current stage but loaded from a USD file on disk already containing render settings.
Adding a repair I'd say would be way too magical to implement I'd say.

PR #139, opened 2024-10-24
Review: Don't add ftrack family

iLLiCiTiT:
## Changelog Description
Do not add ftrack family to review instance automatically.

## Additional info
It is job of ftrack addon to care about it.

## Testing notes:
1. Create package, upload to server, add to bundle (with ftrack addon).
2. Review should go to ftrack.


BigRoy (approved):
If this is indeed handled by Ftrack plug-in I'm fine with this, so the change seems ok. I can't test the Ftrack side of it.

The only worry could be that the instance does not have `instance.data["families"]` set anymore; however I've rarely seen that and most instances should have that by default. So code looks good.

PR #140, opened 2024-10-24
Add python panel for AYON Generic Loaders

MustafaJafar:
## Changelog Description

> [!Note]
> This PR is for trying Houdini python panels for controlling AYON Generic Loaders. 
> It's mainly for sharing code and showing a demo.

- The python panel provide a filter to show/select only AYON generic loaders.
Animation_14

- When using `Load with AYON` action, the pypanel pops up pinned to the generic loader node.  
_Currently, the panel doesn't update according to the reflect the values of the selected node._
Animation_16

## Testing notes:
1. follow the gif above.

### DIFF
--- a/client/ayon_houdini/startup/python_panels/ayon_generic_loader.pypanel
+++ b/client/ayon_houdini/startup/python_panels/ayon_generic_loader.pypanel
@@ -8,7 +8,7 @@
   <interface name="ayon_panel" label="AYON Generic Load Panel" icon="MISC_python" showNetworkNavigationBar="false" help_url="">
     <script><![CDATA[
 import hou
-from PySide2 import QtCore, QtWidgets
+from qtpy import QtCore, QtWidgets
 from ayon_houdini.api.ayon_load_menu import SelectAYONProductDialog
 
 
--- a/client/ayon_houdini/startup/python_panels/ayon_generic_loader.pypanel
+++ b/client/ayon_houdini/startup/python_panels/ayon_generic_loader.pypanel
@@ -1,0 +1,86 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<pythonPanelDocument>
+  <!-- This file contains definitions of Python interfaces and the
+ interfaces menu.  It should not be hand-edited when it is being
+ used by the application.  Note, that two definitions of the
+ same interface or of the interfaces menu are not allowed
+ in a single file. -->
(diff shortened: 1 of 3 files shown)


client/ayon_houdini/startup/python_panels/ayon_generic_loader.pypanel
```diff
@@ -0,0 +1,86 @@
...
+<pythonPanelDocument>
+  <!-- This file contains definitions of Python interfaces and the
+ interfaces menu.  It should not be hand-edited when it is being
+ used by the application.  Note, that two definitions of the
+ same interface or of the interfaces menu are not allowed
+ in a single file. -->
+  <interface name="ayon_panel" label="AYON Generic Load Panel" icon="MISC_python" showNetworkNavigationBar="false" help_url="">
+    <script><![CDATA[
+import hou
+from PySide2 import QtCore, QtWidgets
```
moonyuet, line 11: just curious: can we use qtpy instead of PySide2? I am asking this for more innovative development as I remember we used to have done similar import with PySide2 in Substance Painter but the API uses PySide6 after its update.
But it's okay if there is no way to do that.

BigRoy: Yes - should be `qtpy`. Future versions of Houdini will also ship PySide6.

BigRoy:
Nice! Very interesting prototype. 

I do want to state that I'd like to avoid maintaining more separate Qt UIs where we can - I'd rather focus as much as possible on staying closer to Houdini itself. For example, I'd rather have it be a panel of a regular "node parms" that auto switches. That way, you also still have full control over putting in expressions there, etc. Note that with e.g. context options in LOPs these types of attributes may also be dynamically set from the node graph itself. As soon as we start moving to a custom UI or dialog we'll suddenly be writing Houdini's parameter panel ;) but then ourselves, in Qt. I'd like to avoid that.

Aside of that - this is very nice experiment and could potentially be used as a drawing board / inspiration of other approaches we could take. So let's definitely keep this open for now to think about - but I wouldn't prioritize it unless it offers us features that we really need and would be impossible in native Houdini.

@dee-ynput @antirotor thoughts?

MustafaJafar:
> Currently it's not updating the filepath when using python panel to switch the asset representation type.

I happens because I don't trigger the callbacks after assigning the parameters.
It should be fixed when using expressions instead of callbacks.

MustafaJafar:
closing this in favor of creating [opening Parameter floating panel](https://www.sidefx.com/forum/topic/43083/).

PR #145, opened 2024-10-27
Support Houdini context options from the ROP node for the USD collectors + validations

BigRoy:
## Changelog Description

Support Houdini context options from the ROP node for the USD collectors + validations

## Additional info

- Collect USD stage and layers with the right context options
- Collect the USD Stage and Sdf Layers once and re-use down the line

With this change you can create a Solaris stage in LOPs that is dependent on context options, e.g. for multi-shot lighting workflows and publishing each of them will now correctly retrieve the expected render products for e.g. USD rendering for that Solaris context and/or whatever is different in the LOPs graph for that particular context.

This also changes the behavior that each validator recollects the stage and layers each time from the LOP. Instead, we query them once, keep in memory and re-use them in validators down the line. Whether this will be entirely correct, safe to do and remain without crashes we'll have to investigate in some good testing of publishing some decently complex graphs of models, looks and render scenes. A bigger test render scene with multi-shot workflow that I had seemed to run fine.

How to best keep the unique cloned layers in memory I wasn't sure about and took me a while to figure out something that didn't crash Houdini or was immensely slow. This now seems to behave consistent - but I have an open USD Alliance forum post [here](https://forum.aousd.org/t/copy-in-memory-stage-and-its-layers-to-be-unique-copies/1946/2) hoping to get even better approaches.

## Testing notes:

1. Set up a scene with context options
2. On the USD ROP set the "Context Options" you want for a particular publish instance.
3. Publish both at the same time.
4. It should've correctly applied the validations and the publish.

Note: The generated USD product being correct isn't much up to us - that's still a regular Houdini ROP render so we would mostly be validating whether our collectors and validators are correctly picking up the context options. Like e.g. having different render product output paths per context, or different textures loaded per context (and have those correctly be published along with a look).

### DIFF
--- a/client/ayon_houdini/api/lib.py
+++ b/client/ayon_houdini/api/lib.py
@@ -109,6 +109,107 @@ def get_output_parameter(node):
     raise TypeError("Node type '%s' not supported" % node_type)
 
 
+def get_lops_rop_context_options(
+        ropnode: hou.RopNode) -> "dict[str, str | float]":
+    """Return the Context Options that a LOP ROP node uses."""
+    rop_context_options: "dict[str, str | float]" = {}
+
+    # Always set @ropname and @roppath
+    # See: https://www.sidefx.com/docs/houdini/hom/hou/isAutoContextOption.html
+    rop_context_options["ropname"] = ropnode.name()
+    rop_context_options["roppath"] = ropnode.path()
+
+    # Set @ropcook, @ropstart, @ropend and @ropinc if setropcook is enabled
+    setropcook_parm = ropnode.parm("setropcook")
+    if setropcook_parm:
+        setropcook = setropcook_parm.eval()
+        if setropcook:
+            # TODO: Support "Render Frame Range from Stage" correctly
+            # TODO: Support passing in the start, end, and increment values
+            #  for the cases where this may need to consider overridden
+            #  frame ranges for `RopNode.render()` calls.
+            trange = ropnode.evalParm("trange")
+            if trange == 0:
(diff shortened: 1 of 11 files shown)


MustafaJafar:
_I'm not sure if I'm testing it correctly._
I made this simple scene with some context vars used in my stage as well as my Rop to publish something to my `shot_01` and `shot_02`. 

The Rop works perfectly but the stage doesn't. [RB_shot_01_fx_v001.zip](https://github.com/user-attachments/files/17543574/RB_shot_01_fx_v001.zip)

image

BigRoy:
> _I'm not sure if I'm testing it correctly._ I made this simple scene with some context vars used in my stage as well as my Rop to publish something to my `shot_01` and `shot_02`.

Did you specify the context options on the ROP to override it?

For example:

image

That way you can have two ROPs pointing to the same graph but outputting something else.
The `usdTest` ROP you have there doesn't seem to specify them on the ROP?

Here is a stupidly simple example that just changes the name of a cube prim based on the context options:
[simple_usd_rop_context_options.zip](https://github.com/user-attachments/files/17543912/simple_usd_rop_context_options.zip)

MustafaJafar (approved):
This works on my side. 
So, I needed to add the context vars on my Rop node like this: 
image

I made two tests: 
1. publish usd, I've also changed the output file path manually so that they don't override each other.
image
2. publish usd render, Also, I've changed the path in karma settings where I used my `@shot` that's defined in the Rop node.
image
image

Issue #33, opened 2023-11-24
Support for dependancies for farm caching in Houdini


moonyuet:
Hello, I'm sorry if this comes out as a bit harsh but I think the approach this PR is taking to support caching in the farm is wrong and over engineered.

First of all, caching in the farm (and rendering or any other Houdini processes) are already supported by third-party toolsets (Deadline, HQueue, Tractor...) and in WAY more powerful ways that this PR tries to accomplish and the OP plugin framework can manage. This is duplicating all of that logic in OP and adding 1,398 more lines to the already super complex code base!! Most Houdini TDs are already familiarized with those vanilla workflows and having them learn this other "black box" approach through OP is backwards and doesn't add any benefit in my opinion. You can see an example of a very normal submission to the farm here https://github.com/ynput/OpenPype/pull/5621#issuecomment-1732166830

OpenPype shouldn't try to orchestrate the extract/render dependencies of the Houdini node graph, that's already done by these schedulers/submitters, we just need means to be able to run OP publish tasks of the generated outputs, but without doing any gimmicks, just taking a path, a family, a few other data inputs and registering it to the DB so it runs the other integrate plugins of OP like publishing that to SG/ftrack as well (and ideally the API for doing that in OP should be super straightforward to call it from anywhere! the current json file input isn't the best access point to that funcionality). If we wanted to help the existing vanilla submission OP could provide a wrapper of the vanilla submitters so it sets some reasonable defaults and we could intersect the submission to run some pre-validations on the graph... set some parms that might be driven by the OP settings or create utility HDAs to facilitate the creation of the submitted graph so frame dependencies are set correctly and chunk sizes for simulations... but that's it, we don't need to reinvent the wheel by interpreting how the graph needs to be evaluated.

[...]

PR #149, opened 2024-10-29
Add publish families to cache creators

MustafaJafar:
## Changelog Description
This PR 
- adds  publish families to cache creators.
- removes unnecessary collector.
- implements callbacks to cachesize collector to make its attribute visible if `farm`  is enabled

> [!Note]
> This PR is mainly made for sharing code and show a use case  for https://github.com/ynput/ayon-core/pull/994
> It should be used along https://github.com/ynput/ayon-deadline/pull/49 and https://github.com/ynput/ayon-deadline/pull/54
> Here's what to expect when using the three PRs togther
> Animation_17

## Testing notes:
1. Launch Houdini via AYON
2. Create a cache instance (vdb, model, pointcache, ass, rsproxy)
3. Switch farm toggle in the publisher UI, click save,  click refresh and watch the attributes appear and disappear.


client/ayon_houdini/plugins/publish/collect_chunk_size.py
BigRoy, line 1: Honestly this plug-in should never have been in ayon-houdini to begin with. It's related to farm setting and this should really be in `ayon-deadline` as a setting most likely for e.g. a `farm` family.

client/ayon_houdini/plugins/publish/collect_instances_type.py
BigRoy, line 1: You got to love some good cleanup. ❤️ 🥳

client/ayon_houdini/plugins/publish/collect_chunk_size.py
MustafaJafar: yeah, totally agreed.

MustafaJafar: I'm going to mention it in the houdini deadline refactoring issue https://github.com/ynput/ayon-houdini/issues/33

moonyuet (commented):
I did simple test on this, it is submitted to the farm successfully
image
Are we gonna wait for the merge of this PR https://github.com/ynput/ayon-houdini/pull/171 ? Or there should be something needed to be working on as I see this PR is still a draft?

MustafaJafar:
> Are we gonna wait for the merge of this PR #171 ?

Both are separate PRs. they don't depend on each other.

MustafaJafar:
I can still submit to deadline when I'm on this PR but having `publish.hou` family in `get_publish_families` has a downside, that deadline plugins will still be triggered even if I don't have farm enabled. 

This plugin shouldn't be triggered when disabling submit to farm. 
image

MustafaJafar:
This PR shouldn't add the intermediate publish family `publish.hou` used for submitting to deadline.
as it will always assume they are farm instances.

BigRoy:
Actually @MustafaJafar - I think it is up to the creators to do so. This PR wasn't too far off, it should just be more dynamic.

As soon as you toggle (on the creator) to submit it to the farm or not - then the relevant families should be added if set to farm.
Then on the deadline plug-in it should respond to changes on `instance["families"]` so that if the `farm.publish` (or "publish.hou" if you prefer that naming) job is not set that it's not active.

That way the attributes are still completely aligned with the instance's family.

MustafaJafar:
Is that achieved by implementing a callback that listens to the value change and assigns the family ?

BigRoy:
> Is that achieved by implementing a callback that listens to the value change and assigns the family ?

Correct - value change callback on the creator which adds/removes the farm related family based on the enum's state.

One could approach it the opposite so that all this logic is in the Deadline plug-in to watch for creator attribute changes, however that means it's up to that plug-in to detect whether "farm is enabled" on the attribute which makes it harder to maintain over time if we want to expose the toggle in the creator a different way. Or like add a different additional toggle or whatever.

PR #150, opened 2024-10-30
Look: Validate material allow bindings from any purpose and configure in settings

BigRoy:
## Changelog Description

Validate material allow bindings from any purpose, not just the `UsdShade.Tokens.allPurpose`.

This also exposes the plug-in to settings to allow the default state or disable completely, etc.

n/a

## Testing notes:

1. Assign a material with a binding that is not "All" but also allow preview or full render purposes.

image


Lypsolon (approved):
Tested on Almalinux 9 Houdini 20.5 

- created a model and published it 
- loaded the model into a separate sceeen create 2 materials (preview and render) 
- assigned the materials using assign material / layer break mat library 
- published (all green) 
- loaded the contributed model and all worked

PR #152, opened 2024-11-01
Do not enforce "Match Maya Mask" aperture expression on Load Camera (abc)

### LINKED ISSUE #106: AY-6773_wrong camera aperture
### Please describe the issue you have acountered?

When loading it the aperture is wrong and it changes the framing a lot. If I import the publish manually, then the aperture is right.

_No response_

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1286355929683464253/1286357058102497384)

_This issue was automatically created from Clickup ticket [AY-6773](https://app.clickup.com/t/86c0cue0e)_

BigRoy:
## Changelog Description

Do not enforce "Match Maya Mask" aperture expression on Load Camera (abc).

Instead, allow an option for the user to choose with a setting in settings to pick a default at `ayon+settings://houdini/load/CameraLoader/camera_aperture_expression`

This does now default to the Houdini default behavior instead of the "Match Maya Render Mask" because it seems more aligned with what users seem to want. To keep the old default behavior on load that setting would have to be set to `match_maya`.

image

Fixes #106

## Testing notes:

1. Load camera should default to regular Houdini expression on Alembic camera load
2. Unless the option was chosen to use the Match Maya Render Mask expression.
3. On update the chosen option should persist - so that if on load maya render mask was chosen, it would be preserved on updating.
4. It should be backwards compatible on updates for already loaded cameras in the scene (preserving the match maya render mask expression).

### DIFF
--- a/client/ayon_houdini/plugins/load/load_camera.py
+++ b/client/ayon_houdini/plugins/load/load_camera.py
@@ -1,4 +1,7 @@
+from typing import Optional
 import hou
+
+from ayon_core.lib import EnumDef
 from ayon_core.pipeline import get_representation_path
 
 from ayon_houdini.api import (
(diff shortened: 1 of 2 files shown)


MustafaJafar (approved):
It works. 

--- 
I wonder should artists be able to choose whether to load with `Houdini Default` or `Match Maya render mask`?
Also, should we implement loader actions to convert between them?

MustafaJafar:
> They can - there should be an option box on the action in the Loader?

This is cool, I didn't we can do that. 
Animation_30

Issue #37, opened 2024-07-11
Changing the order of BGEO  cache while creating


krishnaavril:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Please describe the feature you have in mind and explain what the current shortcomings are?

![image](https://github.com/ynput/ayon-houdini/assets/61413765/6d868666-c07c-4346-967b-76c71c28c235)
Can we have BGEO.SC as a default,This eliminates little bit of manual processof changing it everytime.

BGEO.SC is better compression format that BGEO(which is default currently)

### How would you imagine the implementation of the feature?

Change the order when creating the BGEO cache

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the enhancement request.

PR #153, opened 2024-11-02
Changing the order of BGEO cache format while creating

krishnaavril:
## Changelog Description
Changing the order of BGEO cache while creating the point cache to bgeo.sc instead bgeo
since bgeo.sc in a compressed format and widely used

## Testing notes:
1. when creating the point cache, by default it should be bgeo.sc
2. everything should work as expected


krishnaavril:
closes #37

BigRoy (approved):
@krishnaavril I've changed the code, reverting the order but instead setting the default value. Could you confirm whether this is ok with you?

krishnaavril:
Make sense 🫡 It works!

Thank you

PR #154, opened 2024-11-02
Publish-BGEO-cache-using-the-existing-frames

krishnaavril:
## Changelog Description
Publish using the existing frames in houdini BGEO

## Testing notes:
1. When creating added boolean parameter for using the existing frames
2. when publishing if the node is turned on-> it should skip the render and use the existing frames
3. If the node is turned off-> It should render without any issues

### DIFF
--- a/client/ayon_houdini/plugins/create/create_bgeo.py
+++ b/client/ayon_houdini/plugins/create/create_bgeo.py
@@ -13,13 +13,18 @@ class CreateBGEO(plugin.HoudiniCreator):
     product_type = "pointcache"
     icon = "gears"
 
+    # Default render target
+    render_target = "local"
+
     def create(self, product_name, instance_data, pre_create_data):
 
         instance_data.update({"node_type": "geometry"})
         creator_attributes = instance_data.setdefault(
             "creator_attributes", dict())
-        creator_attributes["farm"] = pre_create_data["farm"]
-        creator_attributes["local_no_render"] = pre_create_data["local_no_render"]
+        
+        for key in ["render_target"]:
+            if key in pre_create_data:
+                creator_attributes[key] = pre_create_data[key]
 
         instance = super(CreateBGEO, self).create(
             product_name,
@@ -61,13 +66,17 @@ class CreateBGEO(plugin.HoudiniCreator):
         instance_node.setParms(parms)
 
     def get_instance_attr_defs(self):
+        render_target_items = {
+            "local": "Local machine rendering",
+            "local_no_render": "Use existing frames (local)",
+            "farm": "Farm Rendering",
+            "farm_no_render": "Use existing frames (farm)"
+        }
         return [
-            BoolDef("farm",
-                    label="Submitting to Farm",
-                    default=False),
-            BoolDef("local_no_render",
-                    label="Use existing frames (local)",
-                    default=False)
+            EnumDef("render_target",
+                    items=render_target_items,
+                    label="Render target",
+                    default=self.render_target)
         ]
 
     def get_pre_create_attr_defs(self):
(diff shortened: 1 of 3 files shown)


BigRoy (changes requested):
Unfortunately this is the wrong approach - we should unify it with the other render options of other instances so they behave the same - instead of making this separate unique behavior.

As such, this should instead become an `EnumOption` that allows you to pick from:
- Local render
- Local use existing frames
- Farm render
- Farm use existing frames
- etc.

krishnaavril:
Hello!

After Roy's feedback, I changed the method and approach similar to the other logic, It is working fine for the below 3 methods:

1. Local render
2. Local use existing frames
3. Farm render

Please validate the code @BigRoy  @MustafaJafar 

Can you guys hint where we need to tweak and skip the caching for the "Farm use existing frames" working method?
Appreciate your time!

client/ayon_houdini/plugins/publish/extract_rop.py
```diff
@@ -31,7 +31,12 @@ def process(self, instance: pyblish.api.Instance):
         )
         ext = ext.lstrip(".")
 
-        self.render_rop(instance)
+        creator_attributes = (instance.data["creator_attributes"])
+        if creator_attributes.get("render_target") in {"local"}:
+            self.render_rop(instance)
+        else:
+            print("Skipping Houdini local render...")
```
BigRoy, line 38:
```suggestion
        creator_attributes = instance.data["creator_attributes"]
        if creator_attributes.get("render_target") == "local_no_render":
            self.log.debug(
                "Skipping ROP render to use existing frames"
                f" for instance: {instance}")
        else:
            self.render_rop(instance)
```

client/ayon_houdini/plugins/publish/collect_cache_farm.py
```diff
@@ -18,7 +18,9 @@ class CollectDataforCache(plugin.HoudiniInstancePlugin):
 
     def process(self, instance):
         creator_attribute = instance.data["creator_attributes"]
-        farm_enabled = creator_attribute["farm"]
+        farm_enabled = creator_attribute["render_target"]
+        # Determine settings based on render_target value
+        farm_enabled = False if farm_enabled in ["local", "local_no_render"] else farm_enabled
```
BigRoy, line 23:
```suggestion
        render_target: str = creator_attribute.get("render_target")
        # Determine settings based on render_target value
        farm_enabled = render_target in {"farm", "farm_no_render"}
```

client/ayon_houdini/plugins/create/create_bgeo.py
```diff
@@ -60,10 +66,17 @@ def create(self, product_name, instance_data, pre_create_data):
         instance_node.setParms(parms)
 
     def get_instance_attr_defs(self):
+        render_target_items = {
+            "local": "Local machine rendering",
+            "local_no_render": "Use existing frames (local)",
+            "farm": "Farm Rendering",
+            "farm_no_render": "Use existing frames (farm)"
```
BigRoy, line 73: I'm pretty sure we'll need matching PRs on `ayon-deadline` to support this. @MustafaJafar thoughts?

I do like the feature though.

krishnaavril:
The same technique must also be used in the vdb cache as well

MustafaJafar:
Hello @krishnaavril,
Many thank for your contribution. and in turn I've created another PR https://github.com/ynput/ayon-houdini/pull/180 to support publishing Bgeos with different render targets. 
It reuses some of the logic of render instances. 

IMO, supporting more different render targets like `"farm_no_render": "Use existing frames (farm)"` should have a separate issue.

PR #156, opened 2024-11-04
Fix textures not being collected and published along for `look` products

BigRoy:
## Changelog Description

Fix textures not being collected and published along for `look` products.

This PR https://github.com/ynput/ayon-houdini/pull/145 made it so that the USD stage and layers is collected once globally for an instance and then accessed downstream by other plug-ins. However, the Collect Look Assets plug-in, using this data runs at the same collector order and hence depending on "chance" that plug-in may load before or after the this one making it unreliable.

## Testing notes:

1. Publish "look" for an asset with textures
2. The textures should be published along and remapped accordingly.

### DIFF
--- a/client/ayon_houdini/plugins/publish/collect_usd_rop_layer_and_stage.py
+++ b/client/ayon_houdini/plugins/publish/collect_usd_rop_layer_and_stage.py
@@ -68,7 +68,9 @@ class CollectUsdRenderLayerAndStage(plugin.HoudiniInstancePlugin):
 
     label = "Collect ROP Sdf Layers and USD Stage"
     # Run after Collect Output Node
-    order = pyblish.api.CollectorOrder
+    # Run just before regular CollectorOrder to have stage accessible
+    # to default collector orders
+    order = pyblish.api.CollectorOrder - 0.01
     hosts = ["houdini"]
     families = ["usdrender", "usdrop"]
 


Lypsolon (approved):
tested on almalinux 9 houdini 20.5

image file is in the expected place 
image

im not sure how to to best test the "unreliable" aspect tho.

BigRoy:
> im not sure how to to best test the "unreliable" aspect tho.

Basically - loading the USD layer should be using the remapped local resources. If so, I'd say it's reliably doing what we expect it to. @Lypsolon

Lypsolon:
> > im not sure how to to best test the "unreliable" aspect tho.
> 
> Basically - loading the USD layer should be using the remapped local resources. If so, I'd say it's reliably doing what we expect it to. @Lypsolon

ohhh man i was writing a long text but just realised i read your text wrong. 
soo yes its using the right path `/home/ynput/Documents/Ayon_
Server_Mount/Usd_Testing/Roy_Tests/Houdini/Usd_Pub/https___github_com_ynput_ayon
_houdini_pull_156/publish/look/lookMain/v003/resources/m4.jpg` (i sourced the image from `/home/ynput/pictures`) 
but it might be interesting to know that even with the usd remap relative paths on its an abs path. 

if this is intended then i say its good to go.

PR #157, opened 2024-11-04
add staging dir to render pub

Lypsolon:
## Changelog Description
Brief description of the solution this PR is implementing. This should be only a few sentences and it will be publicly visible in the changelog. You can highlight importance of the change and add links to further information or supporting documentation

## Testing notes:
1. start with this step
2. follow this step

### DIFF
--- a/client/ayon_houdini/plugins/publish/collect_render_products.py
+++ b/client/ayon_houdini/plugins/publish/collect_render_products.py
@@ -7,9 +7,7 @@ import pxr.UsdRender
 import pyblish.api
 
 from ayon_houdini.api import plugin
-from ayon_houdini.api.usd import (
-    get_usd_render_rop_rendersettings
-)
+from ayon_houdini.api.usd import get_usd_render_rop_rendersettings
 
 
 class CollectRenderProducts(plugin.HoudiniInstancePlugin):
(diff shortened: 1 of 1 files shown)


BigRoy:
Also I know why this is implemented, just so that in the `ayon-usd` repo we can rely on it. But I'd argue that it seems we've suddenly introduced a dependency on this attribute which apparently was not needed before. As such - we're maybe misrelying on this data being present for the 'farm' scenario? I'm mostly wondering why we suddenly need this now.

Also if you expected that pinning file to live next to the USD file - then this `stagingDir` is not it.. it only defines where the renders will live, not where the scenes' USD file will live.

Lypsolon:
> It would be very nice if these PRs could exclude the style changes - because now the single line change is like a bulk of changes. Seems you may have an aggressive linter at play here?

its default black and this repo has no config file.

Lypsolon:
> Also I know why this is implemented, just so that in the `ayon-usd` repo we can rely on it. But I'd argue that it seems we've suddenly introduced a dependency on this attribute which apparently was not needed before. As such - we're maybe misrelying on this data being present for the 'farm' scenario? I'm mostly wondering why we suddenly need this now.
> 
> Also if you expected that pinning file to live next to the USD file - then this `stagingDir` is not it.. it only defines where the renders will live, not where the scenes' USD file will live.

good to know. 
then i need to find out what ondra intended there because the system dose rely on some staging dir and i dont know where it should come from.

BigRoy:
I think @antirotor tried to rely on the staging dir of the USD publish - not the staging dir of the rendered images.

BigRoy:
> its default black and this repo has no config file.

I see - be wary that I think default black has a different line length than what we use for AYON dev if I'm not mistaken.
Anyway, @antirotor how should we solve the linter discrepancy here?

Lypsolon:
> > its default black and this repo has no config file.
> 
> I see - be wary that I think default black has a different line length than what we use for AYON dev if I'm not mistaken. Anyway, @antirotor how should we solve the linter discrepancy here?

well if we dont use default then we are missing the config file.

Lypsolon:
> I think @antirotor tried to rely on the staging dir of the USD publish - not the staging dir of the rendered images.

okay i guess that explains somme thing maybe.

PR #158, opened 2024-11-04
Fix Load Asset LOP and Load Shot LOP

BigRoy:
## Changelog Description

Fix Load Asset LOP and Load Shot LOP loading and updating via loader and scene inventory.

PR #112 changed the parameter callbacks to expressions that worked the other way around. Hence just setting `representation` id on the node and triggering a callback (which does not exist anymore) didn't actually work breaking the loaders for these LOPs.

## Testing notes:

1. Load a Load Asset LOP and Load Shot LOP through loader UI
2. The nodes should be loaded correctly.
3. Update the nodes via the Scene Inventory - update should work.


MustafaJafar (approved):
It works. 
Thank you ^^

PR #159, opened 2024-11-04
Remove the :SDF_FORMAT_ARGS: from the collected filenames for Explicit Layer Save paths

BigRoy:
## Changelog Description

Remove the :SDF_FORMAT_ARGS: from the collected filenames for Explicit Layer Save paths

This may fix the issue mentioned here: https://community.ynput.io/t/houdini-usd-publish-errors/1944?u=bigroy

## Testing notes:

1. Create asset with component builder
2. Publish as USD

### DIFF
--- a/client/ayon_houdini/plugins/publish/collect_usd_layers.py
+++ b/client/ayon_houdini/plugins/publish/collect_usd_layers.py
@@ -8,6 +8,7 @@ from ayon_core.pipeline.create import get_product_name
 from ayon_houdini.api import plugin
 import ayon_houdini.api.usd as usdlib
 
+from pxr import Sdf
 import hou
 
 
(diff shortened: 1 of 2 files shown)


MustafaJafar (approved):
This actually works. 

image

I was checking the USD rop output (it's much easier to check it than checking the published files) here it's [testCube.zip](https://github.com/user-attachments/files/17624524/testCube.zip). 

Do we need to have usdAsset file that sublayers the asset created by the component builder? 
is it beneficial for having a consistent contribution workflow?
(what about the default prim validator. can we bypass it?)

---
Also, what about having one more preset (use default prim from the asset) ?

BigRoy:
> Do we need to have usdAsset file that sublayers the asset created by the component builder?
> is it beneficial for having a consistent contribution workflow?
> (what about the default prim validator. can we bypass it?)

In this case it's more of a bugfix then a recommendation on the workflow. Preferably the user is capable of publishing this WITHOUT the need of the extra USD layer the USD ROP writes - and instead is capable of just publishing the Component Builder output directly (which would be the ROP inside the component builder.)

However, that'd be a separate issue to investigate.

> Also, what about having one more preset (use default prim from the asset) ?

That wouldn't be valid with the USD contribution workflow unfortunately - since it just requires the default prim to be your current folder name. Not much freedom there.

MustafaJafar:
Also, loading works. 
I'm not how to test the re-pathing thing in https://github.com/ynput/ayon-houdini/pull/159/commits/d504edb3ea4352e507eddff64eefbd296bc52e27. 
Animation_29

BigRoy:
> Also, loading works.
> I'm not how to test the re-pathing thing in [d504edb](https://github.com/ynput/ayon-houdini/commit/d504edb3ea4352e507eddff64eefbd296bc52e27).

In the Houdini Solaris > Scene Graph Layers panel you should be able to investigate the actual layers it loaded for that stage. The layers from the published file should refer to the published paths, not the paths in your work area. That's it.

PR #160, opened 2024-11-05
Create correct Hdas from Subnetworks, keeping Parms, Outputs and Inputs.

### LINKED ISSUE #99: AY-6728_Keep additional parameters as inputs on HDA publish

Currently the Subnetwork loses any parameters added to it when creating a Houdini Digital Asset Type publish. We'd like it to be able to retain those. As we might add controls for shaders/fx in there
Currently it also removes any inputs. We've found it useful to have at least 1 input so that we can either connect transforms, or inputs for nodes inside

_No response_

_No response_

### Additional context:
()

_This issue was automatically created from Clickup ticket [AY-6728](https://app.clickup.com/t/86c09vge1)_

Lypsolon:
## Changelog Description
This PR automatically detects how many outputs are present in an subnet and how many inputs are connected to any node. it also moves all the parms from a subnet over to the generated hda. 
the Parms, inputs and outputs are kept intact when generating an HDA from an Subnet by copy pasting them over. 

- [ ] test if some expressions might get broken by copy pasting them.

Old Behavior
- when creating an HDA from an Subnet the outputs of the hda will be set to 1 and the inputs to 0. 
- when creating an HDA from a Subnet the parms will be deleted. 
- if you change the inputs of the HDA from the type definition the inputs will survive into the published hda 

New Behavior
- check the indirect inputs of the selected subnet and check if those inputs have a node connected to them. we will then count the subnet inputs that have a node connected (inside the subnet not tpo layer) and set the max inputs of the HDA. 
- we then collect the Output nodes and check if there index is unique (Houdini dose not allow you to create multiple output nodes with the same index but you could create them and then create a HDA this would then internally error.) from this list we then set the HDA outps (we need to set the HDA outputs in there HDA definition after the creation because Houindi dose not offer a parameter in the .CreateHda() function from hou.Node class 
- we copy the parameter template from the source subnet into a tempVar before we create the hda from it. then we simply add the template onto the hda definition. 

## Testing notes:
1. Create a subnet add a few output nodes and connect what ever you want to the input nodes of the subnet(internally not the external ones)
2. Ayon -> Create -> Create HDA 
3. check if all your connections still function correctly and if the right inputs outputs and prams exist. also check if relative parm references and parm expressions still work as expected. 

extra points: 
- create an hda from a selection of nodes and check if this works.

### DIFF
--- a/client/ayon_houdini/plugins/create/create_hda.py
+++ b/client/ayon_houdini/plugins/create/create_hda.py
@@ -218,9 +218,7 @@ class CreateHDA(plugin.HoudiniCreator):
             hda_node.layoutChildren()
         elif self._check_existing(folder_path, node_name):
             raise CreatorError(
-                (
-                    "product {} is already published with different HDA" "definition."
-                ).format(node_name)
+                f"product {node_name} is already published with different HDA definition."
             )
         else:
             hda_node = to_hda
--- a/client/ayon_houdini/plugins/create/create_hda.py
+++ b/client/ayon_houdini/plugins/create/create_hda.py
@@ -236,7 +236,7 @@ class CreateHDA(plugin.HoudiniCreator):
             )
 
             hda_def = hda_node.type().definition()
-            hda_def.setMaxNumOutputs(len(output_index_list))
+            hda_def.setMaxNumOutputs(max(len(output_index_list), 1))
 
             hda_parm_template_group = hda_def.parmTemplateGroup()
 
--- a/client/ayon_houdini/plugins/create/create_hda.py
+++ b/client/ayon_houdini/plugins/create/create_hda.py
@@ -210,9 +210,6 @@ class CreateHDA(plugin.HoudiniCreator):
                 node_name=node_name,
             )
 
-            intput_names = to_hda.inputs()
-            input_count = len(intput_names)
-
             source_parm_template_group = to_hda.parmTemplateGroup()
 
             output_index_list = set()
@@ -224,10 +221,10 @@ class CreateHDA(plugin.HoudiniCreator):
                 output_index = node.parm("outputidx").eval()
                 output_index_list.add(output_index)
 
-            raise RuntimeError(
-                output_index_list,
-                "find out what inputs are connected to outputs and then use this info to fill in the max inputs ",
-            )
+            subnet_inputs_with_connections = set()
+            for subnet_input in to_hda.indirectInputs():
+                if subnet_input.outputs():
+                    subnet_inputs_with_connections.add(subnet_input)
 
             hda_node = to_hda.createDigitalAsset(
                 name=type_name,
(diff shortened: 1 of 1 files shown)


BigRoy (commented):
Is it correct that this draft PR is ONLY code style changes currently - or what am I missing?

Lypsolon:
> Is it correct that this draft PR is ONLY code style changes currently - or what am I missing?

i always create a PR right away, this is how GitLab works and i believe its simply better in all ways. 

also i forgot to push the changes yesterday.

client/ayon_houdini/plugins/create/create_hda.py
```diff
@@ -201,34 +194,63 @@ def create_instance_node(
...
+                for node in to_hda.allSubChildren()
+                if node.type().name() == "output"
+            ]:
+                output_index = node.parm("outputidx").eval()
+                output_index_list.add(output_index)
+
+            subnet_inputs_with_connections = set()
+            for subnet_input in to_hda.indirectInputs():
+                if subnet_input.outputs():
+                    subnet_inputs_with_connections.add(subnet_input)
```
BigRoy, line 227: Could you add a comment above this as to what it is trying to solve?

Why do we care about the children output nodes?
Also, doesn't `allSubChildren()` recurse into potential other subnetworks and their children, etc. which doesn't sound like "outputs" we care about?

Lypsolon: > Why do we care about the children output nodes? Also, doesn't `allSubChildren()` recurse into potential other subnetworks and their children, etc. which doesn't sound like "outputs" we care about?

the recursive thought is a good one i will need to check this case. 

and we care about the child output nodes because that defines how many outputs your subnet will have (i mean there index dose but you cant have 2 output nodes with the same index) i use this to determine how many ouputs your hda should have.

Lypsolon: > Shouldn't we also consider existing outputs (e.g. connections to outside of my current selection) even if I don't have a `output` node for them already?
> 
> It may be worth attaching example hip files you're testing with.

if i understood the code right this should just trigger for subnet types not for node selections ? 
and in an subnet the inputs are determined by the input nodes that have a connection and the outputs by the output nodes

BigRoy: > if i understood the code right this should just trigger for subnet types not for node selections ?
> and in an subnet the inputs are determined by the input nodes that have a connection and the outputs by the output nodes

Ah you may be right. :)

Lypsolon: > Could you add a comment above this as to what it is trying to solve?
> 
> Why do we care about the children output nodes? Also, doesn't `allSubChildren()` recurse into potential other subnetworks and their children, etc. which doesn't sound like "outputs" we care about?

yes should have been .children() not allSubChildren(). 
i fixed this in the last PR

MustafaJafar:
Hello, 
I've created another PR https://github.com/ynput/ayon-houdini/pull/172 based on the changes on this PR.

MustafaJafar:
closing this PR in favor of https://github.com/ynput/ayon-houdini/pull/172

PR #161, opened 2024-11-05
Fix Load Asset LOP and Load Shot LOP (but better?) 

BigRoy:
## Changelog Description

Fix old LOP loaders missing the file/representation parm expressions due to having explicit attribute value overrides

Before the loaders used explicit parm callbacks to set the filepath and representation id on the nodes. As such, these attributes had "overrides" from their default values - lacking the default expressions. This PR, ensures those parameter expressions are 'reverted' on update.

This will fix through the scene inventory, if e.g. the user broke it actively and then trigger UPDATE once via the loader itself to get the support for changing versions on the nodes directly again.

However, it also fixes existing loaders on scene open.

## Testing notes:
Either:
1. Set up Load Asset LOP or Load Shot LOP
2. Explicitly break the `file` or `representation` parm expressions.

Or:
1.  Open a legacy scene with loaders from before the change to expressions PR

Then:
1. Update via scene inventory

The node should be updated.

### DIFF
--- a/client/ayon_houdini/api/hda_utils.py
+++ b/client/ayon_houdini/api/hda_utils.py
@@ -136,23 +136,52 @@ def set_node_representation_from_context(
     node.setParms(parms)
 
     if ensure_expression_defaults:
-        # The filepath and representation id parms are updated through
-        # expressions, however in older versions they were explicitly set
-        # so we ensure that the current value is set to the default value
-        # with the expression - otherwise the value will be set to the
-        # previously explicitly set overridden value.
-        for parm_name in ["representation", "file"]:
-            parm = node.parm(parm_name)
-            if parm is None:
-                continue
+        ensure_loader_expression_parm_defaults(node)
 
-            if parm.isAtDefault(compare_expressions=True):
-                continue
 
-            locked = parm.isLocked()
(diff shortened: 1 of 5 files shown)


MustafaJafar (approved):
It works when restarting my houdini session. (I can open the workfile tool and reopen the same opened file)
also, when updating (e.g. set version) from scene manager. 
it's worth mentioning that scene manager thing won't work when using the loader to load products from other projects. 
image
I was going to suggest having an inventory action, but reopening the scene is much easier for me than triggering an inventory action.
<details><summary>inventory action</summary>

```python
from ayon_core.pipeline import InventoryAction
from ayon_houdini.api.hda_utils import ensure_loader_expression_parm_defaults


class SetDefaultParms(InventoryAction):

    label = "Set Default Parms"
    icon = "wrench"
    color = "orange"

    @staticmethod
    def is_compatible(container):
        return (
            container.get("loader") in {
                "LOPLoadAssetLoader",
                "FilePathLoader",
                "LOPLoadShotLoader"
            }
        )

    def process(self, containers):
        for container in containers:
            node = container["node"]
            ensure_loader_expression_parm_defaults(node)
```
</details>

image

PR #162, opened 2024-11-05
Add `stagingDir` to `usdrender` product

MustafaJafar:
## Changelog Description
needed for `ExtractSkeletonPinningJSON` to work as it expects `usdrender` instances to have `stagingDir`, check it [here](https://github.com/ynput/ayon-usd/blob/4f52045f564bb5c877aee15ff02fa456ea53f1dd/client/ayon_usd/plugins/publish/extract_skeleton_pinning_json.py#L57-L65).

## Additional Info
This PR is exactly the same as https://github.com/ynput/ayon-houdini/pull/157 excluding the style changes. 

## Testing notes:
1. Checkout this PR and https://github.com/ynput/ayon-usd/pull/76
2. publishing `usdrender` should be okay.

### DIFF
--- a/client/ayon_houdini/plugins/publish/collect_render_products.py
+++ b/client/ayon_houdini/plugins/publish/collect_render_products.py
@@ -68,6 +68,8 @@ class CollectRenderProducts(plugin.HoudiniInstancePlugin):
             dirname = os.path.dirname(name)
             basename = os.path.basename(name)
 
+            instance.data["stagingDir"] = dirname
+
             dollarf_regex = r"(\$F([0-9]?))"
             if re.match(dollarf_regex, basename):
                 # TODO: Confirm this actually is allowed USD stages and HUSK


MustafaJafar:
I checked the discussion on https://github.com/ynput/ayon-houdini/pull/157 about the value of `stagingDir`. 
The bottom line of the discussion is: should the pinning file live next to the scenes' USD file or next to the render? 

`stagingDir` in this this PR is the render directory. 

As far as I know, the USD file location is specified in the `usdrender`'s `USD Output Directory` parameter.
So, we may want to us this location instead of the `stagingDir`. (which requires to do some changes in both `ayon-houdini` (to collect that value) and `ayon-usd` (to use the new key instead of `stagingDir`) repos. )
image



Also, for reference, here's what I found out about computing the render meta data path:
it depends on the [outputDir](https://github.com/ynput/ayon-core/blob/9fbc08dd5e296409f4d735593c5595bafb696daf/client/ayon_core/pipeline/farm/pyblish_functions.py#L1279-L1280)

and, regards Houdini it's set to `instance.data["files"][0]` check it [here](https://github.com/ynput/ayon-deadline/blob/a04203add8a6692280566347326d010c15d4ff2f/client/ayon_deadline/plugins/publish/houdini/submit_houdini_cache_deadline.py#L137) and [here](https://github.com/ynput/ayon-deadline/blob/a04203add8a6692280566347326d010c15d4ff2f/client/ayon_deadline/plugins/publish/houdini/submit_houdini_render_deadline.py#L384).
which is roughly the render directory (`stagingDir`)

BigRoy:
I'd say this is not the correct fix.

1. @Lypsolon actually created the exact same fix before here https://github.com/ynput/ayon-houdini/pull/157 but was also closed because not the correct fix.
2. The pinning files is not required to live next to the USD file - we just need to know where the USD file is and where the pinning file will be.
3. I had a meeting with Ondrej and Lyon today actually - out of that we concluded that there wasn't necessarily a reason to only allow the pinning file generation on an existing USD file on disk so instead we now considered instead doing it on the in-memory stage (from the ROP) in Houdini and write that out locally before submitting to the farm. It'd allow a working scenario for all the different render cases + fix some other issues too. So @Lypsolon will be giving that a go.

PR #163, opened 2024-11-05
Publish component builder LOP usd while maintaining hierarchical structure

BigRoy:
## Changelog Description

Implements a prototype creator that can publish the output USD of a Component Builder LOPs network.

Example graph:

image

Publisher:

image

The published asset is contained within a single publish with all its relative paths and localized files:

image

Note that this screenshot does not contain the textures - but that's only because I didn't have them on my disk in this case. That's then logged forgivingly as a warning during publish. For example:

image

Selecting the Creator in the publisher UI and pressing the question mark on the top right of the publisher UI will try and guide you about what it does and how to use it.

image

## Testing notes:

1. Create component builder network in LOPs
2. Select a component output node in the graph
3. Create the component builder LOPs instance
4. (Optional) Disable the USD contribution workflow on the instance. (Since the component builder basically 'defines' the full USD asset it makes little sense to contribute it as part of a parent asset using a contribution workflow.)
5. Publish


client/ayon_houdini/plugins/create/create_usd_componentbuilder.py
```diff
@@ -0,0 +1,75 @@
...
+            this creator. It will generate an instance for each selected
+            Component Output LOP.
+        """)
+
+    def create(self, product_name, instance_data, pre_create_data):
+
+        nodes = hou.selectedNodes()
+
+        builders = [
+            node for node in nodes if node.type().name() == "componentoutput"
```
BigRoy, line 34: There must be a better way to ensure I'm exactly finding the right node type instead of only matching by name?

client/ayon_houdini/plugins/create/create_usd_componentbuilder.py
```diff
@@ -0,0 +1,69 @@
...
+            To use it, select a Component Output LOP and click "Create" for
+            this creator. It will generate an instance for each selected
+            Component Output LOP.
+        """)
+
+    def create(self, product_name, instance_data, pre_create_data):
+        nodes = hou.selectedNodes()
+        builders = [
+            node for node in nodes if node.type().name() == "componentoutput"
+        ]
```
MustafaJafar, line 33:
```suggestion
        builders = hou.lopNodeTypeCategory().nodeType("componentoutput").instances()
```

client/ayon_houdini/plugins/create/create_usd_componentbuilder.py
MustafaJafar: I'm not sure I get it.
what are your the concerns about finding node by type definition name?

let me quote from Houdini [hou.NodeType class](https://www.sidefx.com/docs/houdini/hom/hou/NodeType.html)  docs: 
> All the node types in Houdini are organized into categories, and a node type is uniquely identified by its category and node type name.

BigRoy: It's because the node type name itself can be used in other categories too. So to be entirely exact I **should** be checking against the category and the node type name. ;) So basically the current check isn't entirely explicit because it may pick up the wrong node types.

client/ayon_houdini/plugins/create/create_usd_componentbuilder.py
BigRoy: Unfortunately this would return ALL nodes of them - and notes the ones in the `nodes` input list which is the user's selection. ;)

But likely doing a check like:

```python
if node.type() == hou.lopNodeTypeCategory().nodeType("componentoutput"):
```
Or something along those lines may make sense and is pretty readable/explicit.
Or maybe even easier, we just do `node.type().nameWithCategory()` and compare with the category included directly - I think I'll do that instead.

MustafaJafar:
```suggestion
        builders = [
            node for node in nodes if node.type().nameWithCategory() == "Lop/componentoutput"
        ]
```

MustafaJafar:
using the following publish template : 
1. in `ayon+anatomy://RnD/templates/publish`
    ```json
    {
      "name": "HoudiniComponentBuilder",
      "directory": "{root[work]}/{project[name]}/{hierarchy}/{folder[name]}/publish/{product[type]}/{product[name]}/{@version}",
      "file": "{originalBasename}.{ext}"
    }
    ```
2. in `ayon+settings://core/tools/publish/template_name_profiles` 
  image

allows me to replicate have the same output with the same names in the  publish folder. 
**but keep in mind this publish template will also applied to all USD products!**
image

MustafaJafar:
I tried publishing the output of component builder as an assembly variant in usd contribution. 
I may be missing a step. 
Anyways, do you think it's a good thing/valuable to publish the component builder a an assembly product in a `usdAsset`? 

---

Side question: What tasks types are most suitable for component builders? since component builders include look and model products.

BigRoy:
> I tried publishing the output of component builder as an assembly variant in usd contribution.
> I may be missing a step.
> Anyways, do you think it's a good thing/valuable to publish the component builder a an assembly product in a `usdAsset`?

I'd say component builder makes little sense with the contribution workflow - yet at the same time, there may be an edge case where one may want to do that and instead have it _be_ the full usd asset itself.

> Side question: What tasks types are most suitable for component builders? since component builders include look and model products.

Generic task type maybe. But yes, there isn't a clear 'default' task type in AYON that is the combination of those two.

MustafaJafar:
I've created https://github.com/ynput/ayon-usd/issues/78 to track the issue of using publish templates and contribution workflow. 

Gave it another test with the default publish template. 
When using a used rop (which publish each layer of the component builder as a separate layer + one additional layer by the usd rop itself), the contribution workflow works as usual.

When using the new component builder creator (with contribution enabled and variant disabled), the contribution workflow breaks because the default prim was not set properly. (making these changes manually made it work)
image
 
When using the new component builder creator (with contribution enabled and variant enabled), it works.

and Here's an example where I published both the component builder directly and using a USD ROP along with contribution workflow. 
Animation_40

client/ayon_houdini/plugins/publish/collect_componentbuilder_lop.py
```diff
@@ -0,0 +1,84 @@
...
+        publish_root = instance.data["publishDir"]
+        for src in paths:
+
+            if filepath == Path(src):
+                continue
+
+            relative_path = os.path.relpath(src, start=src_root_dir)
+            self.log.debug(f"Collected dependency: {relative_path}")
+            dest = os.path.normpath(os.path.join(publish_root, relative_path))
+            transfers.append((src, dest))
```
MustafaJafar, line 84: Maybe we can use the thumbnail as source thumbnail for the publish. 

```suggestion
            transfers.append((src, dest))

            # Use the thumbnail created by the component builder
            # if the user didn't explicitly take a snapeshot using the publisher.
            if (
                relative_path.startswith("thumbnail") 
                and 
                not instance.data.get("thumbnailSource")
            ):
                instance.data["thumbnailSource"] = src
```

client/ayon_houdini/plugins/create/create_usd_componentbuilder.py
```diff
@@ -0,0 +1,70 @@
...
+            instance = CreatedInstance(
+                self.product_type,
+                product_name,
+                instance_data,
+                self)
+            self._add_instance_to_context(instance)
+            self.imprint(instance_node, instance.data_to_store())
+        except hou.Error as er:
+            raise CreatorError("Creator error: {}".format(er)) from er
+
```
MustafaJafar, line 55: This is essential for contribution workflow to work properly.
```suggestion

        instance_node.setParms({
            # Set the 'default prim' by default to the folder name being
            # published to
            "rootprim": '/`strsplit(chs("folderPath"), "/", -1)`',
        })
```

MustafaJafar (commented):
This PR works pretty well on my side. 
Some few notes.

client/ayon_houdini/plugins/create/create_usd_componentbuilder.py
BigRoy: Let's not do this. The component builder is more likely to be used without contribution workflow.

client/ayon_houdini/plugins/publish/collect_componentbuilder_lop.py
BigRoy: I wouldn't do this, the component builder has explicit thumbnail generation logic. Let's rely on that solely. This is all about the component builder, not about is ingesting something else out of the blue.

client/ayon_houdini/plugins/publish/collect_componentbuilder_lop.py
```diff
@@ -0,0 +1,84 @@
...
+    def process(self, instance):
+
+        node = hou.node(instance.data["instance_node"])
+
+        # Render the component builder LOPs
+        # TODO: Do we want this? or use existing frames? Usually a Collector
+        #  should not 'extract' but in this case we need the resulting USD
+        #  file.
+        node.cook(force=True)  # required to clear existing errors
+        node.parm("execute").pressButton()
```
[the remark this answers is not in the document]
BigRoy: So this makes sense, however... At the same time I wonder why this isn't triggering on the regular 'render'. There must be a reason that SideFX doesn't do this *always* but has kept it separate. I'm worried that we are too confidently enforcing thumbnail updates where as user may have not wanted this to happen.

Also, wasn't there a toggle on the component builder to make it itself auto generate/save the thumbnail on render without having to trigger it separately?

I feel this is also one of those things that we are deciding for the user more than one may want. However at the same time I personally can't think of a reason if why you wouldn't want the thumbnail updated too (unless the generation of it can be relatively slow or the quality of the default generation sucks?). Anyway, this might be best checked with the client. I am fine with committing this if there is reason to do so, but also would be fine with ignoring it.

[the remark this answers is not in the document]
MustafaJafar: For reference, this fixes the bug mentioned in https://github.com/ynput/ayon-houdini/pull/163#pullrequestreview-2427741417 where the thumbnail is always expected.
Let me try it once again.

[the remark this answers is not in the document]
MustafaJafar: > Also, wasn't there a toggle on the component builder to make it itself auto generate/save the thumbnail on render without having to trigger it separately?

Yes, found it. let me quote from [Component Builder Docs](https://www.sidefx.com/docs/houdini/solaris/component_builder.html)
```
If you set Mode to Render or Houdini GL , you can turn on Auto-generate on export to always regenerate the thumbnail when you write out the USD files.
```

But it's not enabled by default.

client/ayon_houdini/plugins/create/create_usd_componentbuilder.py
MustafaJafar: What about considering the component builder `Validate USD Asset Contribution Default Prim` so that users can get a hint about wha is going wrong? maybe we can add an issue for it for reference?

client/ayon_houdini/plugins/publish/collect_componentbuilder_lop.py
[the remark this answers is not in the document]
MustafaJafar: I'm not sure how to tackle this. 
But, I have two ideas: 
1. Exclude the thumbnail from the publish if it doesn't exist. (although the existent thumbnail can be wrong)
2. Force generating the thumbnail, this 👇  might be a better code snippet, but this may be overwhelming as the thumbnail can be heavy as the node supports using Karma.
    ```suggestion
            node.parm("execute").pressButton()
            autothumbnail_parm = node.parm("autothumbnail")
            if not(autothumbnail_parm.isVisible() and autothumbnail_parm.eval()): 
                node.parm("executeviewport").pressButton()
    ```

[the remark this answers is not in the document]
BigRoy: Thanks. I still want to avoid forcing the thumbnail. It's absolutely fine for it to not exist.

In my test runs it ran absolutely fine without thumbnail. It would log a warning but if I recall correctly it published fine. If for whatever reason it's failing to publish then we should fix that. I thought the path would get skipped if it didn't exist, I recall writing that code and logging warnings for that.

client/ayon_houdini/plugins/create/create_usd_componentbuilder.py
MustafaJafar: > I believe the validator you describe is already in place?

Yes, I was mainly talking about [ValidateUSDAssetContributionDefaultPrim](https://github.com/ynput/ayon-houdini/blob/develop/client/ayon_houdini/plugins/publish/validate_usd_asset_contribution_default_prim.py)

So, I did a small experiment to reuse it with component builder.
Here's it

https://github.com/BigRoy/ayon-houdini/compare/enhancement/147-ay-7047_publishing-component-builder-usd-while-maintaining-hierarchical-structure...ynput:ayon-houdini:enhancement/147-ay-7047_publishing-component-builder-usd-while-maintaining-hierarchical-structure-add-rootprim-validation

---
I've found a potential bug here, if the default prim was modified manually and the user forgot to add `/` at the beginning. this case will pass the following condition 

https://github.com/ynput/ayon-houdini/blob/fb03bd0b91bbf7c9ef701b940eae40884ec56c2a/client/ayon_houdini/plugins/publish/validate_usd_asset_contribution_default_prim.py#L63-L63

client/ayon_houdini/plugins/publish/collect_componentbuilder_lop.py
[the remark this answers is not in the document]
MustafaJafar: omg 😅 
I can't reproduce it.

MustafaJafar (approved):
It works!

client/ayon_houdini/plugins/create/create_usd_componentbuilder.py
BigRoy: I see. I think it does look fine the way you implemented it - I just would recommend not doing it at all, because it's not a production use case at all as far as I know. It could be worth leaving this code somewhere memorable, e.g. a draft PR that you close so that it's easier found back later if we do want to implement it for new reasons.

PR #164, opened 2024-11-05
Bugfix: Farm cache publish support newer ayon-deadline versions

### LINKED ISSUE #138: Collect cache farm instances and Deadline plugins order

### Current Behavior:

This [collector](https://github.com/ynput/ayon-deadline/blob/develop/client/ayon_deadline/plugins/publish/global/collect_deadline_job_env_vars.py#L19) won't work with houdini's farm cache submission although it has `publish.hou`. 
this happens because its order is less than the [houdini collector](https://github.com/ynput/ayon-houdini/blob/90cc39fbb700fa671f8affdf3441a86758804611/client/ayon_houdini/plugins/publish/collect_cache_farm.py#L14) that adds  `publish.hou` family to farm cache instances.

### Expected Behavior:

I should be able to see the Deadline plugin in the publisher.

### Version

1.0.0

Windows

### Steps To Reproduce:

1. Create a cache publish instance.
2. Enable `Submitting to Farm`
3. Publish
4. You won't be able to see `Deadline Farm Environment Variables` plugin.

### Relevant log output:

_No response_

### Additional context:

_No response_

BigRoy:
## Changelog Description

Set the `publish.hou` family much earlier to avoid issues with `ayon-deadline` plug-ins that were ordered earlier than how we set it before this PR in `ayon-houdini`.

Fixes #138

## Testing notes:

1. Use latest ayon-deadline
2. Publish farm cache
3. Should work

### DIFF
--- a/client/ayon_houdini/plugins/publish/collect_cache_farm.py
+++ b/client/ayon_houdini/plugins/publish/collect_cache_farm.py
@@ -46,25 +46,23 @@ class CollectDataforCache(plugin.HoudiniInstancePlugin):
         ropnode = hou.node(instance.data["instance_node"])
         output_parm = lib.get_output_parameter(ropnode)
         expected_filepath = output_parm.eval()
-        instance.data.setdefault("files", list())
-        instance.data.setdefault("expectedFiles", list())
 
+        files = instance.data.setdefault("files", list())
         frames = instance.data.get("frames", "")
         if isinstance(frames, str):
             # single file
-            instance.data["files"].append(expected_filepath)
+            files.append(expected_filepath)
         else:
             # list of files
             staging_dir, _ = os.path.split(expected_filepath)
-            instance.data["files"].extend(
-                ["{}/{}".format(staging_dir, f) for f in frames]
-            )
+            files.extend("{}/{}".format(staging_dir, f) for f in frames)
 
-        cache_files = {"cache": instance.data["files"]}
+        expected_files = instance.data.setdefault("expectedFiles", list())
+        expected_files.append({"cache": files})
+        self.log.debug(f"Caching on farm expected files: {expected_files}")
 
         instance.data.update({
+             # used in HoudiniCacheSubmitDeadline in ayon-deadline
             "plugin": "Houdini",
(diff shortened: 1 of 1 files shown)


MustafaJafar (approved):
It works. Thank you.

BigRoy:
> it was wrong.

Actually - it was right. @MustafaJafar both create from context and this one being at `-0.5` made it ambiguous which would run first. This could be `-0.499` sure, but it MUST run after that one because that is what creates the instances from the Create Context.

MustafaJafar:
> > it was wrong.
> 
> Actually - it was right. @MustafaJafar both create from context and this one being at `-0.5` made it ambiguous which would run first. This could be `-0.499` sure, but it MUST run after that one because that is what creates the instances from the Create Context.

So, the `-0.5` is valid but it conflicts with [CollectFromCreateContext](https://github.com/ynput/ayon-core/blob/develop/client/ayon_core/plugins/publish/collect_from_create_context.py).

MustafaJafar (approved):
It works.

BigRoy:
> I wonder if CollectFarmCacheFamily plugin can be replaced by using the get_publish_families method.

Sort of. But it shouldn't be there ALL the time, but only if `farm` is enabled on the instance. So we'd need to make that dynamic (which is possible now with value changed callbacks) however then the Deadline plug-ins wouldn't "dynamically" respond to that by default and the deadline plug-ins would need to actively listen for changes on the families to dynamically update as soon as the instance's families changes through such a toggle in the UI.

BigRoy:
Merging this - because it fixes a bug. We can turn it into changed callback in separate PR.

MustafaJafar:
> But, do we need to make this plugin run at earlier time? the `CollectJobInfo` has a late value `pyblish.api.CollectorOrder + 0.420`

 I forgot about that context plugin `CollectDeadlineJobEnvVars` that has `pyblish.api.CollectorOrder` order. and this PR fixed the issue where the submitted jobs didn't have the proper env variables to run. 

so it does fix a bug thank you.

Issue #23, opened 2023-07-13
Houdini: Unable to review/Render using Lopimportcamera


krishnaavril:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Please describe the feature you have in mind and explain what the current shortcomings are?

<img width="1529" alt="image" src="https://github.com/ynput/OpenPype/assets/61413765/60aa286d-06f2-4d50-a7e7-bfe3f266821e">

Currently OP is publishing without errors using SOP level import camera. Its thowing error when loading the LOP camera

### How would you imagine the implementation of the feature?

The OP should successfully publish, using Lopimportcamera SOP node as well

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the enhancement request.

[cuID:[OP-6365](https://app.clickup.com/t/863h6y90b)]

MustafaJafar:
Related but doesn't solve the issue. 
Publishing render supports marking as review.
![image](https://github.com/ynput/ayon-houdini/assets/20871534/39ad2f18-d914-4e81-8502-deba7b36a2d7)

PR #167, opened 2024-11-08
Support new flipbook node for reviews

### LINKED ISSUE #85: AY-6687_Houdini 20.5: Support new "Flipbook" node for reviews

[Houdini 20.5 introduces a new ROP node "Flipbook"](https://www.sidefx.com/docs/houdini//nodes/out/flipbook.html) which acts like a replacement to the the OpenGL ROP but instead uses the Vulkan viewport renderer.

It would be great if we can add support for it to use that for "review" renders instead of the OpenGL ROP. The node's parameters pretty much seems like a 1:1 match of the OpenGL node.

Support the new Flipbook ROP node natively as a review instance.

Client mentioned that for them it made sense to set a default in settings somewhere so that e.g. a review defaults to either the OpenGL node OR the Flipbook node. I'm not convinced that's the best approach yet - I think having a separate creator may be safest, that is _also_ generating `review` product type should be just fine. Then in settings the studio can enable/disable EITHER creator so the old remains as is, and the new becomes available. Since we may never know how much the two nodes will remain aligning.

### Additional context:

_No response_

MustafaJafar:
## Changelog Description
resolve #85 

This PR enhances the current review plugins to add support for the new flipbook node.
image

## Additional Info
Flipbook node behaves a lot similar to opengl node. 
one interesting change is the `Color Correction` setting which can choose  (`color space`) or (`display` and `view`).
which may change a bit the color space validation and collection to properly support the flipbook node.

## Testing notes:
1. Create review instance using `Review`  creator
2. Try different node types
3. Publish Reviews


client/ayon_houdini/plugins/create/create_review.py
```diff
@@ -150,3 +151,10 @@ def get_pre_create_attr_defs(self):
                       minimum=0.0001,
                       decimals=3)
         ]
+
+
+class CreateFlipbookReview(CreateReview):
+    
+    identifier = "io.ayon.creators.houdini.review.flipbook"
+    label = "Review (Flipbook)"
+    node_type = "flipbook"
```
MustafaJafar, line 160: > Client mentioned that for them it made sense to set a default in settings somewhere so that e.g. a review defaults to either the OpenGL node OR the Flipbook node. I'm not convinced that's the best approach yet - I think having a separate creator may be safest, that is _also_ generating `review` product type should be just fine. Then in settings the studio can enable/disable EITHER creator so the old remains as is, and the new becomes available. Since we may never know how much the two nodes will remain aligning.

Quoted from the linked issue. 
Now, should we have a separate creator for the flipbook node? or using the same creator while exposing the node type?

BigRoy:
> Should this PR adds support for publishing reviews from LOPs?
> Adding this for reference [ynput/OpenPype#4890](https://github.com/ynput/OpenPype/pull/4890)

I'd say that yes, we should fix that... **but** it doesn't need to be part of this PR. Instead I see that irrelevant to this because it's already relevant to the OpenGL ROP too. I thought we had an issue for that, but I can't find it 🤦

client/ayon_houdini/plugins/create/create_review.py
BigRoy: Whatever makes most sense from development side of things I'd say. It may appear cleaner if it's one Creator to avoid clutter in the Create list?

MustafaJafar:
> I thought we had an issue for that, but I can't find it 🤦 

We have this one https://github.com/ynput/ayon-houdini/issues/23

BigRoy:
> > I thought we had an issue for that, but I can't find it 🤦
> 
> We have this one #23

Yup, saw that one - but unfortunately it's not about supporting cameras IN LOPS - but that is requesting to support `lopimportcamera` which is still at geometry level.

BigRoy:
Let's see if we can get this one https://github.com/ynput/ayon-houdini/pull/148 merged first - so that this PR is slightly simplified and works with the changes made there.

moonyuet:
I slightly tested on the PR and find out that it's not working for the elder version of Houdini. Shall we introduce the condition so that users would know it only supports for Houdini 20.5? Or we can do it with some scripts in elder version?
I find a simple script from this website for creating flipbook animation for the elder version of Houdini: https://www.sidefx.com/docs/houdini/anim/flipbook.html

BigRoy:
> I slightly tested on the PR and find out that it's not working for the elder version of Houdini. Shall we introduce the condition so that users would know it only supports for Houdini 20.5? 

We could enforce the `enabled` state of the Creator to `False` in an older version of Houdini than 20.5?

> Or we can do it with some scripts in elder version?
> I find a simple script from this website for creating flipbook animation for the elder version of Houdini: https://www.sidefx.com/docs/houdini/anim/flipbook.html

No please. That flipbook would just be the `OpenGL review` we already have.

client/ayon_houdini/plugins/create/create_review.py
MustafaJafar: https://github.com/ynput/ayon-houdini/pull/167/commits/5f8d3254c8b611c627e81f2865608637aa605fbb

server/settings/create.py
```diff
@@ -9,6 +9,22 @@ class CreatorModel(BaseSettingsModel):
...
+    ]
+
+class CreateReviewModel(BaseSettingsModel):
+    enabled: bool = SettingsField(title="Enabled")
+    default_variants: list[str] = SettingsField(
+        title="Default Products",
+        default_factory=list,
+    )
+    node_type: str = SettingsField(
+        Title="Default Node Type",
```
MustafaJafar, line 25: No idea why the title doesn't show the word `Default`
image

client/ayon_houdini/plugins/create/create_review.py
```diff
@@ -29,8 +30,15 @@ def apply_settings(self, project_settings):
...
     def create(self, product_name, instance_data, pre_create_data):
-
-        instance_data.update({"node_type": "opengl"})
+        
+        self.node_type = pre_create_data.get("node_type")
+        hou_version = hou.applicationVersionString()
+        if self.node_type == "flipbook" and not hou_version.startswith("20.5"):
+            self.log.debug("The Flipbook node type isn't supported in Houdini versions below 20.5."
+                           " Switching to the OpenGL node type instead.")
+            self.node_type = "opengl"
```
BigRoy, line 39: Let's remove this here - and just keep the `flipbook` entry out of the `Enum`? if `houdini < 20.5`

client/ayon_houdini/plugins/create/create_review.py
```diff
@@ -123,8 +131,15 @@ def get_pre_create_attr_defs(self):
             "bmp", "cin", "exr", "jpg", "pic", "pic.gz", "png",
             "rad", "rat", "rta", "sgi", "tga", "tif",
         ]
+        node_type_enum = [
+            "opengl", "flipbook"
+        ]
```
BigRoy, line 136:
```suggestion
        node_type_enum = ["opengl"]
        if hou.applicationVersion() >= (20, 5, 0):
            node_type_enum.append("flipbook")
```

client/ayon_houdini/plugins/create/create_review.py
BigRoy: Also, we should not be updating the `self.node_type` value - it makes no sense to change the class attribute from here.

client/ayon_houdini/plugins/publish/extract_rop.py
```diff
@@ -78,13 +78,32 @@ def update_representation_data(self,
...
 
-class ExtractOpenGL(ExtractROP,
-                    publish.ColormanagedPyblishPluginMixin):
+class ExtractOpenGLAndFlipbook(ExtractROP,
+                               publish.ColormanagedPyblishPluginMixin):
 
     order = pyblish.api.ExtractorOrder - 0.01
-    label = "Extract OpenGL"
+    label = "Extract Review (OpenGL & Flipbook)"
     families = ["rop.opengl"]
```
BigRoy, line 86: I'm not too sure about both these node types going through `rop.opengl` family - they should each get their own family so that we could potentially still target just one of the two if we get to any specific validations down the line. Not critical for this PR - but worth a low prio backlog issue.

client/ayon_houdini/plugins/publish/extract_rop.py
```diff
@@ -78,13 +78,32 @@ def update_representation_data(self,
...
+            self.log.debug("Skipping instance without instance node.")
+            return
+
+        rop_node = hou.node(instance_node)
+        if rop_node.type().name() not in {"opengl", "flipbook"}:
+            self.log.debug("Skipping OpenGl extraction. Rop node {} "
+                           "is not an OpenGl node.".format(rop_node.path()))
+            return
+
+        super(ExtractOpenGLAndFlipbook, self).process(instance)
```
BigRoy, line 105:
```suggestion
        super().process(instance)
```

client/ayon_houdini/plugins/publish/validate_review_colorspace.py
```diff
@@ -57,6 +57,18 @@ def apply_settings(cls, project_settings):
...
+
+        # This plugin is triggered when marking render as reviewable.
+        # Therefore, this plugin will run on over wrong instances.
+        # TODO: Don't run this plugin on wrong instances.
+        # This plugin should run only on review product type
+        # with instance node of opengl type.
+        if rop_node.type().name() not in {"opengl", "flipbook"}:
+            self.log.debug("Skipping Validation. Rop node {} "
+                           "is not an `opengl` or `flipbook` node.".format(rop_node.path()))
+            return
```
BigRoy, line 70: Why is this needed? This is not the case anymore right? Now it triggers on `rop.opengl` family?

server/settings/create.py
BigRoy: It's the capital `T` ;)
```suggestion
        title="Default Node Type",
```

client/ayon_houdini/plugins/create/create_review.py
MustafaJafar: https://github.com/ynput/ayon-houdini/pull/167/commits/9bb792ead39c988edc78f19342dcfeba8c62fa94

client/ayon_houdini/plugins/publish/validate_review_colorspace.py
MustafaJafar: https://github.com/ynput/ayon-houdini/pull/167/commits/39637766f65e6723d2a1179b4906ab0ce908f424

client/ayon_houdini/plugins/publish/extract_rop.py
MustafaJafar: https://github.com/ynput/ayon-houdini/pull/167/commits/39637766f65e6723d2a1179b4906ab0ce908f424

client/ayon_houdini/plugins/publish/extract_rop.py
MustafaJafar: Added a todo here https://github.com/ynput/ayon-houdini/pull/167/commits/979452c4dee0a1769be2ed418b1897467ed10524
and also here's an issue https://github.com/ynput/ayon-houdini/issues/176

client/ayon_houdini/api/lib.py
```diff
@@ -896,40 +896,51 @@ def get_current_context_template_data_with_entity_attrs():
...
-    opengl_node.setParms(
+    node.setParms(
         {"ociocolorspace": review_color_space}
     )
 
     log.debug(
         "'OCIO Colorspace' parm on '{}' has been set to "
         "the view color space '{}'"
-        .format(opengl_node, review_color_space)
+        .format(node.path(), review_color_space)
```
moonyuet, line 943: just wondering why you don't use `node` (hou.Node) here as the previous code used `opengl_node` which happened to be hou.Node.

moonyuet:
When testing in 19.5, the flipbook option does not show up as expected.
image

client/ayon_houdini/api/lib.py
MustafaJafar: I thought it was already solved.

MustafaJafar: for consistency, I turned this one into `node.path()` instead of just `node`
other log messages are using `node.path()`

moonyuet (changes requested):
I tested with 20.5 and find out that the repair action is not repairing the colorspace param as expected and the validator keeps blocking me to publish.
The video would show you why and it is related to the regex form of the ocio colorspace in houdini i believe. i.e. `'<USE_DISPLAY_NAME>'`

https://github.com/user-attachments/assets/ec209ca0-df58-40bd-b793-1c3b064666ec

MustafaJafar:
> The video would show you why and it is related to the regex form of the ocio colorspace in houdini i believe. i.e. `'<USE_DISPLAY_NAME>'`

I remember Roy mentioned it somewhere! 
Unfortunately, I don't know how to replicate it. I can only find the token [here](https://github.com/ynput/ayon-core/blob/a181fc897d16db9563fca73473eddee590cdd427/client/ayon_core/pipeline/colorspace.py#L1486) in the code.

moonyuet (approved):
Publish successfully in Houdini 20.5
image

MustafaJafar:
> I re-test again and it seems that I haven't turned on the color configuration in Ayon setting, meaning that It only happens when the OCIO color management is not enabled and using the default Houdini ocio configuration. @BigRoy, just want to hear your thought on this. We can create an issue to resolve this later(I dont think it should be solved here but instead the separate PR.)


I was able to reproduce it in develop branch with opengl node. 
image

Actually, I can pass the validation when choosing something manually from the drop down menu. (even if I disable the color management toggle in Houdini addon settings)
image

moonyuet:
> > I re-test again and it seems that I haven't turned on the color configuration in Ayon setting, meaning that It only happens when the OCIO color management is not enabled and using the default Houdini ocio configuration. @BigRoy, just want to hear your thought on this. We can create an issue to resolve this later(I dont think it should be solved here but instead the separate PR.)
> 
> I was able to reproduce it in develop branch with opengl node. image
> 
> Actually, I can pass the validation when choosing something manually from the drop down menu. (even if I disable the color management toggle in Houdini addon settings) image

But the users may expect that the repair action would fix it. We can actually inform the users somewhere in the docs to talk about this limitation, and we need to figure out what to do to fix it eventually if there is somebody who runs on this issue.

MustafaJafar:
I've created an issue for it.
https://github.com/ynput/ayon-houdini/issues/179

Issue #12, opened 2024-04-30
Houdini: Switch HDA version manually


MustafaJafar:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior:

Houdini users can change filepath in HDA definition. which allows them to swap between different versions. 
While the inventory tool doesn't detect such change. 

![Animation_52](https://github.com/ynput/ayon-core/assets/20871534/48989b58-8f7f-401e-aaa2-dddb612c58e0)

### Expected Behavior:

As a user I expected some live updates. where Inventory tool can show the real version based on my selection.

### Version

1.0.0

### What platform you are running on?

Windows

### Steps To Reproduce:

1. publish HDA multiple times from Houdini.
2. change path manually from asset definition. 

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the bug report.

[cuID:[AY-5263](https://app.clickup.com/t/86bypmhc8)]

BigRoy:
This _could_ be solved by ON publish and saving of the HDA secretly imprint it with the parms needed for the containerization (like give it the representation id, etc.) so that if you just create it - it has those attributes.

As such, switching versions would also switch these vaues?

MustafaJafar:
This issue happens because we are adding the parameters after node creation (spare parameters) and spare parameters are not updated when changing the HDA definition manually.

I can think of two solutions: 
1. As @BigRoy mentioned earlier, we can add these parameters on publishing so that they be part of the definition.
2. We can add a callback on publishing that takes care of updating the `Extra`  parameters when changing the asset definition.

PR #172, opened 2024-11-15
propagate the number of inputs of the subnetwork to the HDA and support publishing HDAs with spare parameters

### LINKED ISSUE #99: AY-6728_Keep additional parameters as inputs on HDA publish

MustafaJafar:
## Changelog Description
This PR: resolve #99 
- [x] Propagate the number of inputs of the subnetwork to the HDA creation
- [x] Keep user spare parameters when publishing
- [x] Keep the HDA and its parameters/spare parameters in sync when changing the HDA version.

> [!Note]
> This PR is inspired by the code changes in https://github.com/ynput/ayon-houdini/pull/160

## Additional Info
This PR exposed an issue where adding a ramp spare parameter breaks collecting instances for the publisher tool.
https://github.com/ynput/ayon-houdini/issues/173

## Demo

https://github.com/user-attachments/assets/e5850692-811a-415f-9614-5b75f6b13bc3

https://github.com/user-attachments/assets/d6245ecd-eb91-48b1-9c92-38cd9d54c6d3

## Testing notes:
- Case 1: 
  - test: Create subnetwork and connect nodes to its inputs and create different output nodes inside it.
  - expected result: HDA should be created without losing the connections.


MustafaJafar:
I've one question: 
Which of the following situations means that the problem is with retaining the additional parameters?
- The parameters were added to the subnet node and they disappear when creating an HDA from the subnet node.
  Note, Houdini by default add these additional parameters to the HDA as shown in this demo.

    https://github.com/user-attachments/assets/9f38da77-4976-496a-9703-23b89ddd553c

- The additional parameters disappears when loading the HDA or creating it from tab menu.

    https://github.com/user-attachments/assets/decf3e9a-4022-4724-bef0-8cf2f524e4d7

    https://github.com/user-attachments/assets/ec79eb94-4232-43f8-93b7-914a1e5750bb


I'll assume it's the second one. I thought we already have an issue for it other than the linked issue #99 but I was wrong.

MustafaJafar:
I checked the issue. so the parameters will disappear when using the parameter interface to add them. 
However, using the type properties to add those parameters, keeps them on the node. 

Maybe the parameter interfaces is something related to the node instance itself not the node the type. 
and the type properties changes the actual type.


https://github.com/user-attachments/assets/b27900c9-ec88-4b52-a78f-251c5fc46012


So, enabling `Save Spare Parameters` also saves the extra parameters added by the AYON HDA creator plugin. 
image
image

Maybe, I can destroy AYON extra parameters before extraction and re-add them at the end of extraction? (while enabling the save spare parameters of course).

BigRoy:
> Maybe the parameter interfaces is something related to the node instance itself not the node the type.
> and the type properties changes the actual type.

Yes - that is correct.

moonyuet:
I tested by publishing one version with custom ordered menu and the other with the menu and some float attributes, It seems that the parameter is still not updated as expected. (It could be that I am using 19.5 for test)

https://github.com/user-attachments/assets/f8710e2c-ef45-4912-9a32-4311eca912cf

MustafaJafar:
> I tested by publishing one version with custom ordered menu and the other with the menu and some float attributes, It seems that the parameter is still not updated as expected. (It could be that I am using 19.5 for test)
> 
>  updated_parameter_not_shown_in_Houdini.mp4

This issue is shown in my demo (the second video in the PR description).

moonyuet:
> > I tested by publishing one version with custom ordered menu and the other with the menu and some float attributes, It seems that the parameter is still not updated as expected. (It could be that I am using 19.5 for test)
> > updated_parameter_not_shown_in_Houdini.mp4
> 
> This issue is shown in my demo (the second video in the PR description).

I also did some retest on that, finding it actually loads the updated parameters for even elder version. I think that makes sense if houdini doesn't really have callback to update parameters by version(show or hide the parameters).

client/ayon_houdini/plugins/publish/extract_hda.py
```diff
@@ -13,17 +12,33 @@ class ExtractHDA(plugin.HoudiniExtractorPlugin):
     families = ["hda"]
 
     def process(self, instance):
-        self.log.info(pformat(instance.data))
+
         hda_node = hou.node(instance.data.get("instance_node"))
         hda_def = hda_node.type().definition()
         hda_options = hda_def.options()
         hda_options.setSaveInitialParmsAndContents(True)
+        hda_options.setSaveSpareParms(True)
```
moonyuet, line 20:
```suggestion
        hda_options.setLockContents(True)
        hda_options.setSaveSpareParms(True)
```
Maybe trying to lock the contents might help to resolve the updating issue in the HDA.

BigRoy:
So hear me out. 

The only reason **why** we're imprinting these nodes with these attributes is so that we know what source representation they are, etc. Nothing requires us to store it on the nodes as attributes - we just happen to be doing so because other implementations in Houdini did so too. And hence we could rely on the pre-existing "get_containers" logic that just looked for those attributes.

However, nothing is keeping us from adding an additional loop there that does something like (psuedocode):
```
for hda in hda_definitions:
    container = get_hda_container_data(hda)
    if container:
        yield container
```

Or if we do want to start from the node get the representation data from the HDA, but from the node - etc.
Anyway, the question then becomes: why couldn't we store these attributes directly in the HDA definition on publish (maybe not even as attributes)

moonyuet:
> So hear me out.
> 
> The only reason **why** we're imprinting these nodes with these attributes is so that we know what source representation they are, etc. Nothing requires us to store it on the nodes as attributes - we just happen to be doing so because other implementations in Houdini did so too. And hence we could rely on the pre-existing "get_containers" logic that just looked for those attributes.
> 
> However, nothing is keeping us from adding an additional loop there that does something like (psuedocode):
> 
> ```
> for hda in hda_definitions:
>     container = get_hda_container_data(hda)
>     if container:
>         yield container
> ```
> 
> Or if we do want to start from the node get the representation data from the HDA, but from the node - etc. Anyway, the question then becomes: why couldn't we store these attributes directly in the HDA definition on publish (maybe not even as attributes)

I tried to manually export the hda node with ayon instance data and, it seems that it gets the parameter data right when I imported.. I tried to use the python to export hda, the parameter data is loaded correctly with the contents locked before exporting. 
Speaking of the HDA definition for publish, I do see other issues when we add another folder group other than just `Extra` folder template, and the extractor only gets the parameters from `Folder` group published. But this can be a future talk.
image

MustafaJafar:
Let me add a quick summary. 

First of all this is the standard Houdini behavior, it is not relevant to our HDA plugins (to create, publish and load HDA product). 
However, using our HDA plugins makes these issues occurs more often.

So, there are couple of things: 

### Parameters UI and spare parameters
When the HDA is loaded, The Extra parameters folder might change the parameters UI. 
Actually, it depends on how the parameters were arranged the first place.
e.g. I got these two results (and they are not so ugly)

| No folder for spare parameters | use folder for spare parameters |
|--|--|
| image | image |

### Update spare parameters on definition change

_This one is what really bothers me_

Changing the HDA asset definition either from the definition dropdown menu or from the loader. doesn't update the spare parameters accordingly. 
It most likely happen because HDAs by default doesn't have a callback to update the spare parameters. 
**BUT, they will show up correctly if the node is removed and re-created.** 

The best way to avoid this issue is by adding the parameters using the type properties.

### Mixing parameters (type properties) and spare parameters

This one can result in weird lookin parameters UI especially when keep changing the asset definition. 
This also is avoided by deleting and re-creating the node.


image

Here's how it looks like when removing and re-creating. 
image

BigRoy:
> **Also, keep in mind, that when changing the definition, the Extra folder is not updated accordingly.** mentioned here #12

This is only the case because:

1. The attributes are not part of the HDA definition
2. The loader on update updates the `representation` id value - which in turn becomes a parameter override (so even if it were part of HDA definition it would still now have an override).

If the attributes were part of the HDA definition (Type Properties of the HDA) then we'd never need to change the attributes at all from the loader - we would just need to switch the HDA version itself.

---

However, I wanted to clarify - what I meant with my earlier post was: "Can't we avoid attributes altogether?" - Like, can't we as part of the HDA defintion itself (not in Type Properties, but in metadata elsewhere) store this representation id, etc. _on publish_ (so it's a static part of the HDA definition itself). That way, loading the HDA - in what form whatsoever, even if loaded e.g. manually without AYON it would still have the imprinted containerisation and could be detected as such?

If that's completely impossible, I also want to stress that by no means is it necessary that these attributes are in a folder, and/or are visible. As such, again on publish of the HDA definition, we can add to its type properties (as we "save the definition" and publish it) the relevant metadata we need as hidden parameters (or even one 'blob data' parameter if we prefer that) where we store what we need and later look back for to find the loaded containers.

---

I'm completely confused by this issue described here @MustafaJafar - the parameters should not change, at all. What's up with that 'juggling of parameters'? The HDA definition should just have a set of attributes (Type Attributes) and that's it. And on load/update we preferably do as LITTLE as possible so we can rely as much as possible on Houdini. So, at best, all we need to 'detect' the representation is completely defined in the HDA definition - and not something we containerize on load.

MustafaJafar:
As far as I know, to have the `Extra` parameters part of the HDA definition, I can think of two solutions: 
- we will need to know the value of these parameters (e.g representation id) before saving the file. (and we will add it and delete it afterwards)
- maybe after the publishing, we can open the published file and add these parameters/set their values.

Honestly, I can't think of a clean way to do it.

Also, for reference, I'm removing the `Extra` parameters before saving the file and re-adding them after saving the HDA file.
Maybe at this point I can add more parameters or convert spare parameters to be parameters. 

----

> I'm completely confused by this issue described here @MustafaJafar - the parameters should not change, at all. What's up with that 'juggling of parameters'? The HDA definition should just have a set of attributes (Type Attributes) and that's it. And on load/update we preferably do as LITTLE as possible so we can rely as much as possible on Houdini. So, at best, all we need to 'detect' the representation is completely defined in the HDA definition - and not something we containerize on load.

This happens automatically by Houdini. If my HDA has parameters and spare parameters. 
and the parameters differs per different version, Houdini will do its default behavior where it moves the spare parameters to a folder called `spare` and put the parameters in a folder called `standard`. 

to trigger this effect, you can simply publish HDA with different parameters and use manage to change the version or you can just change it from the asset definition toolbar. 
Animation_45

BigRoy:
> This happens automatically by Houdini. If my HDA has parameters and spare parameters.
> and the parameters differs per different version, Houdini will do its default behavior where it moves the spare parameters to a folder called `spare` and put the parameters in a folder called `standard`.

So technically, this has nothing to do with us - but is an _issue_(?) on the Houdini side itself. Or is there reason this is happening to us specifically?

MustafaJafar:
> > So technically, this has nothing to do with us - but is an _issue_(?) on the Houdini side itself. Or is there reason this is happening to us specifically?
> 
> Because we are making it easier to publish HDAs with spare parameters supported, this behavior will show up more often.

Also for reference, [From Develop Branch] I can publish HDA with parameters (that were add using type properties) and the parameters will exist when loading the HDA!

moonyuet:
> > This happens automatically by Houdini. If my HDA has parameters and spare parameters.
> > and the parameters differs per different version, Houdini will do its default behavior where it moves the spare parameters to a folder called `spare` and put the parameters in a folder called `standard`.
> 
> So technically, this has nothing to do with us - but is an _issue_(?) on the Houdini side itself. Or is there reason this is happening to us specifically?

Not sure if it is really Houdini side of thing as we can actually export the parameters when using python shell to do some simple export. The best solution is to reach out the support :).

BigRoy:
In a call with Mustafa we went through the PR and changes - we ended up going down this route:

- We are **not** enforcing the "Save Spare Parameters" on the HDA definition - but with this PR do allow a houdini power user to enable it manually inside the HDA definition and make sure the custom "Extra" publish attributes from AYON itself (from the Creator) will then be force excluded from the publish.
- We added support to "Promote spare parameters" to type properties on initial HDA creation, because **that's** what houdini does by default and most likely the main source of confusion for most users - maybe even @krishnaavril too? Once it's a HDA definition, the user can still choose to either add type properties or spare parameters, etc. The promotion of the spare parameters on Create can be disabled by a new pre create attribute.

image

- We fixed some creation logic where potentially if a subnet was the first in your selection (when having multiple nodes selected) it would incorrectly make an HDA of just that subnet instead of collapsing all together.
- We changed/fixed logic that if the user selected one node and it was already an HDA that it would still be collapsed - now it will make just that HDA publishable.

moonyuet (approved):
Publish successfully
image
One with spare parameter
image
One without spare parameter
image
The error is not related to the parameter itself, it's just that the input is mandatory for the loaded HDA with the spare parameter. 
image

MustafaJafar:
It works on my side too.
One minor cosmetic thing.. on updating HDA, should we remove the extra folder and re-append it so it always at the end?
image

PR #180, opened 2024-11-20
Support render targets for cache instances

### LINKED ISSUE #19: publish existing caches

Render Targets is supported with render but not cache. 
image

Render targets should be supported with cache

_No response_

### Additional context:

_No response_

MustafaJafar:
## Changelog Description
resolve #19 
Support render targets for the following products: (I'll add them by creators)
- Alembic camera
- Arnold ass 
- Pointcache Bgeo
- Model
- Pointcache Abc
- Redshift proxy
- Review
- Static Mesh
- USD (it works with contribution workflow. users should wait the publish before loading the usd asset)
- USD Look (it works with contribution workflow, users should wait the publish before loading the usd asset))
- VDB Cache

> [!Note]
> When supporting render targets for Houdini `CompositeSequence` (Cop Node), composite instances triggers `Submit Image Publishing job to Deadline` along `Submit cache jobs to Deadline` because the instance has both `imagesequence` and `publish.hou` families.
> Here's an issue for it https://github.com/ynput/ayon-houdini/issues/183

This means that supporting Cop Nodes will need to go through the render submitting logic instead of cache submitting logic.
image

## Testing notes:
1. Try publishing different cache instances with different render targets

### DIFF
--- a/client/ayon_houdini/plugins/publish/collect_cache_farm.py
+++ b/client/ayon_houdini/plugins/publish/collect_cache_farm.py
@@ -10,15 +10,13 @@ from ayon_houdini.api import (
 class CollectFarmCacheFamily(plugin.HoudiniInstancePlugin):
     """Collect publish.hou family for caching on farm as early as possible."""
     order = pyblish.api.CollectorOrder - 0.45
-    families = ["ass", "pointcache", "redshiftproxy", "vdbcache", "model"]
+    families = ["ass", "pointcache", "redshiftproxy",
+                "vdbcache", "model", "staticMesh",
+                 "rop.opengl", "usdrop", "camera"]
     targets = ["local", "remote"]
     label = "Collect Data for Cache"
 
     def process(self, instance):
-        creator_attribute = instance.data["creator_attributes"]
-        farm_enabled = creator_attribute.get("farm")
-        if farm_enabled is not None:
-            instance.data["farm"] = farm_enabled
 
         if not instance.data["farm"]:
             self.log.debug("Caching on farm is disabled. "
--- a/client/ayon_houdini/plugins/publish/collect_cache_farm.py
+++ b/client/ayon_houdini/plugins/publish/collect_cache_farm.py
@@ -16,9 +16,11 @@ class CollectFarmCacheFamily(plugin.HoudiniInstancePlugin):
 
     def process(self, instance):
         creator_attribute = instance.data["creator_attributes"]
-        farm_enabled = creator_attribute["farm"]
-        instance.data["farm"] = farm_enabled
-        if not farm_enabled:
+        farm_enabled = creator_attribute.get("farm")
+        if farm_enabled is not None:
+            instance.data["farm"] = farm_enabled
+
+        if not instance.data["farm"]:
             self.log.debug("Caching on farm is disabled. "
                            "Skipping farm collecting.")
             return
--- a/client/ayon_houdini/plugins/publish/extract_rop.py
+++ b/client/ayon_houdini/plugins/publish/extract_rop.py
@@ -19,6 +19,7 @@ class ExtractROP(plugin.HoudiniExtractorPlugin):
         if instance.data.get("farm"):
             self.log.debug("Should be processed on farm, skipping.")
             return
+        creator_attribute = instance.data["creator_attributes"]
 
         files = instance.data["frames"]
         first_file = files[0] if isinstance(files, (list, tuple)) else files
@@ -29,7 +30,8 @@ class ExtractROP(plugin.HoudiniExtractorPlugin):
         )
         ext = ext.lstrip(".")
 
-        self.render_rop(instance)
+        if creator_attribute.get("render_target") in {None, "local"}:
+            self.render_rop(instance)
         self.validate_expected_frames(instance)
 
         # In some cases representation name is not the the extension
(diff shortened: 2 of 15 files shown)


client/ayon_houdini/plugins/publish/extract_rop.py
```diff
@@ -29,7 +30,8 @@ def process(self, instance: pyblish.api.Instance):
         )
         ext = ext.lstrip(".")
 
-        self.render_rop(instance)
+        if creator_attribute.get("render_target") in {None, "local"}:
+            self.render_rop(instance)
```
BigRoy, line 34:
```suggestion
        if creator_attribute.get("render_target", "local") == "local":
            self.render_rop(instance)
```

client/ayon_houdini/plugins/publish/extract_rop.py
```diff
@@ -19,6 +19,7 @@ def process(self, instance: pyblish.api.Instance):
         if instance.data.get("farm"):
             self.log.debug("Should be processed on farm, skipping.")
             return
+        creator_attribute = instance.data["creator_attributes"]
```
BigRoy, line 22: Do be wary that _technically_ an instance may not have `creator_attributes` if it's e.g. a runtime instance and not all data is transferred from a CreateContext instance. It may be worth (for how generic this extractor is) to do:

```suggestion
        creator_attribute = instance.data.get("creator_attributes", {})
```

client/ayon_houdini/plugins/publish/collect_cache_farm.py
```diff
@@ -16,9 +16,11 @@ class CollectFarmCacheFamily(plugin.HoudiniInstancePlugin):
 
     def process(self, instance):
         creator_attribute = instance.data["creator_attributes"]
-        farm_enabled = creator_attribute["farm"]
-        instance.data["farm"] = farm_enabled
-        if not farm_enabled:
+        farm_enabled = creator_attribute.get("farm")
```
BigRoy, line 19: Shouldn't this be `render_target` now?
```suggestion
        render_target = creator_attribute.get("render_target")
        if render_target != "farm":
            self.log.debug("Caching on farm is disabled. "
                   "Skipping farm collecting.")
            return

        instance.data["farm"] = True
```

MustafaJafar: Forgot about adding a comment about backward/other current cache instances compatibility. 
other instances don't have `render_target` they only have `farm` creator attribute.

MustafaJafar: Let me try, I hope it doesn't need further changes.

client/ayon_houdini/plugins/publish/extract_rop.py
```diff
@@ -29,7 +30,8 @@ def process(self, instance: pyblish.api.Instance):
         )
         ext = ext.lstrip(".")
 
-        self.render_rop(instance)
+        if creator_attribute.get("render_target", "local") == "local":
+            self.render_rop(instance)
```
moonyuet, line 34: Maybe using the instance.data["farm"] for this condition?
```suggestion
        if not instance.data.get("farm", False):
            self.render_rop(instance)
```

client/ayon_houdini/plugins/publish/collect_cache_farm.py
moonyuet: I agree with @BigRoy that ditching `farm` BoolDef to avoid the confusion on the codes. It shouldn't take too long to make these changes.

client/ayon_houdini/plugins/publish/extract_rop.py
MustafaJafar: when instance doesn't have the `farm` instance data it'll either be: 
- local publish (do render)
- publish existing frames (skip render)

moonyuet: > when instance doesn't have the `farm` instance data it'll either be:
> 
> * local publish (do render)
> * publish existing frames (skip render)

My bad, Been unaware of the other options of `publish existing frame`.
Take a look one more time, you can just do this.
```suggestion
        if creator_attribute.get("render_target") == "local":
            self.render_rop(instance)
```

MustafaJafar: Some instances won't have `render_target` 
in that case, the fallback should be `local`. 

But, I'll check it again once I make other cache instances types have render targets.

MustafaJafar: I checked the families in `ExtractRop`.
Unfortunately, There'll be some instances without the `render_target` key at the moment. instances with `imagesequence` specifically.

client/ayon_houdini/plugins/publish/collect_cache_farm.py
MustafaJafar: Thank you. 
https://github.com/ynput/ayon-houdini/pull/180/commits/1124f995898efcc10741a44e00ce24f278093556

client/ayon_houdini/plugins/publish/extract_rop.py
MustafaJafar: let me add a comment in the a code about it.

MustafaJafar: https://github.com/ynput/ayon-houdini/pull/180/commits/dbcce0b947f15ec002fc8b23031dd888ebbdb89b

moonyuet (approved):
Looks good to me and test successfully in Houdini.

PR #185, opened 2024-11-26
Containerise and parse container with `project_name` parm

BigRoy:
## Changelog Description

Support loading containers with project name from another project - to allow managing loaded products from e.g. library project

Requires https://github.com/ynput/ayon-core/pull/1005

## Testing notes:
1. Load products from Library Project
2. They should appear fine in scene inventory and should be updatable.
3. Load Asset LOP, Generic Loader, etc. should also work with this (and be updatable via scene inventory).


BigRoy:
> however, actions e.g. `set version` in doesn't work. will report on the core PR.

This is now fixed in the core PR.

Issue #188, opened 2024-12-02
Automatic thumbnail generation on publishing review


Dreamink-Official:
### Is there an existing issue for this?

- [x] I have searched the existing issues and added correct labels.

### Please describe the feature you have in mind and explain what the current shortcomings are?

Currently when publishing a review it does not create a thumbnail automatically. (Because of that it does not display the review as "play button" directly in ftrack).

### Suggested implementation?

I would solve that by adjusting the ExtractActiveViewThumbnail class. 
- Adding the "reviews" to the family
- Setting a representation for the created thumbnail

### Describe alternatives you've considered:

An alternative would be to add another extract step which generates the thumbnail out of the rendered video file.

PR #189, opened 2024-12-02
Automatic thumbnail generation on publishing review 

Dreamink-Official:
Added product family "review" and representation to instance

## Changelog Description
As proposed in #188 I adjusted the ExtractActiveViewThumbnail class to allow to extract a thumbnail for the product family "review" it adds a representation for the extracted thumbnail.

## Testing notes:
1. Open Houdini and publish a review

### DIFF
--- a/client/ayon_houdini/plugins/publish/extract_active_view_thumbnail.py
+++ b/client/ayon_houdini/plugins/publish/extract_active_view_thumbnail.py
@@ -1,4 +1,5 @@
 import tempfile
+import os
 import pyblish.api
 
 from ayon_core.pipeline import OptionalPyblishPluginMixin
@@ -17,7 +18,8 @@ class ExtractActiveViewThumbnail(plugin.HoudiniExtractorPlugin,
     """
     order = pyblish.api.ExtractorOrder + 0.49
     label = "Extract Active View Thumbnail"
-    families = ["workfile"]
+    families = ["workfile",
+                "review"]
 
     def process(self, instance):
         if not self.is_active(instance.data):
@@ -42,6 +44,21 @@ class ExtractActiveViewThumbnail(plugin.HoudiniExtractorPlugin,
                        .format(view_thumbnail)
         )
         instance.data["thumbnailPath"] = view_thumbnail
+      
+        instance.data.setdefault("representations", [])
+        representation = {
+            "name": "thumbnail",
+            "ext": "jpg",
+            "files": os.path.basename(view_thumbnail),
+            "stagingDir": os.path.dirname(view_thumbnail),
+            "thumbnail": True
+        }
+
+        # Adding representation to instance
+        instance.data["representations"].append(representation)
+        self.log.debug(
+            "Adding thumbnail representation: {}".format(representation)
+        )
 
     def get_view_thumbnail(self, instance):
 


client/ayon_houdini/plugins/publish/extract_active_view_thumbnail.py
```diff
@@ -42,6 +44,21 @@ def process(self, instance):
...
+            "files": os.path.basename(view_thumbnail),
+            "stagingDir": os.path.dirname(view_thumbnail),
+            "thumbnail": True
+        }
+
+        # Adding representation to instance
+        instance.data["representations"].append(representation)
+        self.log.debug(
+            "Adding thumbnail representation: {}".format(representation)
+        )
```
MustafaJafar, line 61:
```suggestion
```

you shouldn't add representation for the thumbnail as there are dedicated plugins for creating the representation. 
- [extract_thumbnail_from_source.py](https://github.com/ynput/ayon-core/blob/develop/client/ayon_core/plugins/publish/extract_thumbnail_from_source.py)
- [extract_thumbnail.py](https://github.com/ynput/ayon-core/blob/develop/client/ayon_core/plugins/publish/extract_thumbnail.py)

client/ayon_houdini/plugins/publish/extract_active_view_thumbnail.py
```diff
@@ -17,7 +18,8 @@ class ExtractActiveViewThumbnail(plugin.HoudiniExtractorPlugin,
     """
     order = pyblish.api.ExtractorOrder + 0.49
     label = "Extract Active View Thumbnail"
-    families = ["workfile"]
+    families = ["workfile",
+                "review"]
```
MustafaJafar, line 22: My feeling, this is more of a "workaround" than a "proper solution".

Additionally, a snapshot from the sceneview may not look the same as the first frame of the review.
This is not for review only but also render and image sequence products.

What about using the first frame as the thumbnail? (we'll need to convert it to `png` or `jpg`).

Dreamink-Official: Using the first frame as the thumbnail could also work. The input from our artists was that it would be great to be able to choose a specific frame. But you are right. In the end that will not look like the real render by using the screenshot of the viewport. So perhaps adding a option in the publisher to be able to select which frame should be used as thumbnail (not only the middle frame of the sequence) could be an option?

client/ayon_houdini/plugins/publish/extract_active_view_thumbnail.py
MustafaJafar: yes! 
Houdini host is missing as you have pointed out.

client/ayon_houdini/plugins/publish/extract_active_view_thumbnail.py
MustafaJafar: feel free to create an issue for it on ayon-core.

MustafaJafar:
From previous discussion, the issue was not an implemented feature in Houdini. 
The feature already exists in core addon but it was lacking Houdini host. 

Closing this PR in favor of https://github.com/ynput/ayon-core/pull/1048

PR #190, opened 2024-12-04
Use `PublishError` along publishing 

### LINKED ISSUE #108: Refactor `RuntimeError`s to `PublishError`s

Since https://github.com/ynput/ayon-core/pull/861 was merged, it'd be cool to refactor some of  `RuntimeError`s to `PublishError`.

check code that is used during publishing and refactor errors.

_No response_

### Additional context:

_No response_

MustafaJafar:
## Changelog Description
resolve #108 

This PR implements `PublishError` along the publish plugins. 
- [x] `RuntimeError`
- [x] `assert`
- [x] `KnownPublishError`
- [x] Different type of Errors along publishing

> [!Note]
> This PR shouldn't add `PublishError` to any functions in lib. instead, it should uses `try..except..` in the caller code. More info https://github.com/ynput/ayon-houdini/pull/190#discussion_r1870933370

## Some Screenshots
image
image

## Testing notes:
1. Everything should still work as before
2. Error messages related to the changes should be better.


client/ayon_houdini/plugins/publish/collect_render_products.py
```diff
@@ -107,10 +108,11 @@ def replace(match):
...
                 filename = filename.replace("\\", "/")
 
-            assert "#" in filename, (
-                "Couldn't resolve render product name "
-                "with frame number: %s" % name
-            )
+            if "#" not in filename:
+                raise PublishError(
+                    message="Couldn't resolve render product name"
+                            f" with frame number: {name}"
```
BigRoy, line 114: I feel using the `message=` argument part here is just redundant clutter. For all errors raised usually the first argument is the message - so it's just as clear without naming it here.

```suggestion
                    "Couldn't resolve render product name"
                    f" with frame number: {name}"
```

client/ayon_houdini/plugins/publish/collect_usd_layers.py
```diff
@@ -38,12 +39,17 @@ def copy_instance_data(instance_src, instance_dest, attr):
...
+            )
 
         src_value = src_data[key]
         if i != len(key):
             dest_data = dest_data.setdefault(key, {})
-            assert isinstance(dest_data, dict), "Destination must be a dict"
+            if not isinstance(dest_data, dict):
+                raise PublishError(
+                    message="Destination must be a dict."
+                )
```
BigRoy, line 52: This provides way too little detail to be any meaningful error to any artist or alike. What destination where, for which key - and what did it get instead? We're really (I feel) going the wrong way if this can't just be a `TypeError` but requires a `PublishError` since we suddenly are introducing a super generic error much less clear? (I mean, still better than the assert, don't get me wrong)

@illicit thoughts?

client/ayon_houdini/plugins/publish/save_scene.py
```diff
@@ -16,9 +17,10 @@ def process(self, context):
         # Filename must not have changed since collecting
         host = registered_host()
         current_file = host.get_current_workfile()
-        assert context.data['currentFile'] == current_file, (
-            "Collected filename from current scene name."
-        )
+        if context.data['currentFile'] != current_file:
+            raise PublishError(
+                message="Collected filename from current scene name."
```
BigRoy, line 22: This error would be way clearer if it describes in the message what is the current scene file and what did it compare against.
```suggestion
                "Collected filename differs from current scene name."
```

client/ayon_houdini/plugins/publish/collect_usd_layers.py
MustafaJafar: Should every error (that can be raised along the publishing) be a `PublishError` ?
I've copula of scattered thoughts. 
- Any error that is not `PublishError` or `PublishValidationError`  will cause the publisher UI to show the message (It's not your fault etc).
- Some errors are originated from user interactions e.g. not selecting the correct node or using a wrong colorspace. I believe these definitely should be `PublishError` and `PublishValidationError` .  These errors can be fixed by the user.
- Other errors like the one you highlighted above, honestly, I don't know why/how it can happen and if the user has something to do with it. I assume it requires a developer to deal with them. but using `PublishError` can help developers know about the problem from a screenshot. (Note: I didn't use the title in the `PublishError` so the publish plugin's label will be used.)

client/ayon_houdini/api/lib.py
```diff
@@ -676,7 +676,10 @@ def get_top_referenced_parm(parm):
 def evalParmNoFrame(node, parm, pad_character="#"):
 
     parameter = node.parm(parm)
-    assert parameter, "Parameter does not exist: %s.%s" % (node, parm)
+    if not parameter:
+        raise PublishError(
+            message=f"Parameter does not exist: {node}.{parm}",
+        )
```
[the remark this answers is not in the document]
MustafaJafar: I followed along that line of code, I found out that its function is being called during publishing. 
this one was using `assert` and as far as I know. 
```python
assert condition, message
```
is equivalent to 
```python
if not condition:
    raise AssertionError (message)
```

except that `assert` can be disabled globally using `-O`  flag. (I don't think we use it.)
https://stackoverflow.com/questions/48408692/raise-assertionerror-vs-assert-python

Should I just sandwich the whole code block with try except?

client/ayon_houdini/plugins/publish/collect_render_products.py
MustafaJafar: https://github.com/ynput/ayon-houdini/pull/190/commits/47cd40951abb582d2e2db8bd17c5220653723bae

client/ayon_houdini/plugins/publish/save_scene.py
MustafaJafar: https://github.com/ynput/ayon-houdini/pull/190/commits/4502143ed6891c034bc74eed355f7796d6a0ef19

MustafaJafar:
The function contains this line is used in `imprint` function which is used in publishing and loading.
Should we do something about it?
https://github.com/ynput/ayon-houdini/blob/cacbe68dbaf2297575e5620ff1b1494bfe958763/client/ayon_houdini/api/lib.py#L574

client/ayon_houdini/plugins/publish/validate_frame_token.py
```diff
@@ -31,8 +31,8 @@ def process(self, instance):
 
         invalid = self.get_invalid(instance)
         if invalid:
-            raise RuntimeError(
-                "Output settings do no match for '%s'" % instance
+            raise PublishError(
+                f"Output settings do no match for '{instance}'"
```
MustafaJafar, line 35: Should I used `PublishValidationError` instead here? 
```suggestion
            raise PublishValidationError(
                f"Output settings do no match for '{instance}'",
                title=self.label
```

BigRoy: I think that may be better **if** the user can do something about it. Just seeing "Output setting do not match" I have no idea what to do next. What output settings mismatch? What can I do about it? Why do they mismatch, we should inform the artist what is going on and what to do about it.

client/ayon_houdini/plugins/publish/collect_usd_layers.py
BigRoy: > Any error that is not PublishError or PublishValidationError will cause the publisher UI to show the message (It's not your fault etc).

Yup - this was a _design decision_ at some point. I'm still not 100% confident that it's the right one, but it did force us into writing better `PublishValidationError`. If now we're just going to make EVERYTHING `PublishError` without providing better messages I think we're in the same problematic realm and we could basically revert to allow any error be nicely reported - we'll just still need to make sure that the majority of errors we raise are very clear what do to about them.

> Some errors are originated from user interactions e.g. not selecting the correct node or using a wrong colorspace. I believe these definitely should be PublishError and PublishValidationError . These errors can be fixed by the user.

I can't think of any now - but that does sound right. However, the error must be very clear, describe what is wrong and how to go about fixing it, etc.

> Other errors like the one you highlighted above, honestly, I don't know why/how it can happen and if the user has something to do with it. I assume it requires a developer to deal with them. but using PublishError can help developers know about the problem from a screenshot. (Note: I didn't use the title in the PublishError so the publish plugin's label will be used.)

Some assertions in the codebase were just there - to assert for a _potential_ edge case where it might happen... not necessarily that they would happen. That's why some might not necessarily be bad to remain asserts (or other regular errors). They were just there because we just really wanted to make sure if the code continued that something behaved a certain way.

client/ayon_houdini/plugins/publish/collect_output_node.py
```diff
@@ -67,8 +67,8 @@ def process(self, instance):
             out_node = node.parm("startnode").evalAsNode()
 
         else:
-            raise KnownPublishError(
-                "ROP node type '{}' is not supported.".format(node_type)
+            raise PublishError(
```
moonyuet, line 70: I would still prefer `KnownPublishError` instead of `PublishError` if `KnownPublishError` is not gonna be deprecated.

BigRoy: Why would you prefer it? :)

moonyuet: > Why would you prefer it? :)

Both exceptions are more likely to describe the publish error (to me it sounds like it happens in the extractor), `KnownPublishError` sounds better as the error is `known` before publishing. (It is just me maybe)

BigRoy: To me it is more like: as soon as we raise a dedicated error for something - it is technically known? :D

client/ayon_houdini/api/lib.py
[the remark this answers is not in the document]
MustafaJafar: Great concept! Let me try.

client/ayon_houdini/plugins/publish/collect_output_node.py
BigRoy: [It's also deprecated](https://github.com/ynput/ayon-core/blob/3da898b3440b5fd9ba75887205e033cc834685a2/client/ayon_core/pipeline/publish/publish_plugins.py#L42).

MustafaJafar: Cool let's kill it 🔪 🔫

client/ayon_houdini/api/lib.py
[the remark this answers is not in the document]
MustafaJafar: so, what do you think moving the `PublishError` to publish plugins? 
Personally, I think this is much better as lib functions can be accessed from any where and TDs are free to use them in their automations. 
But, it came at the cost of writing more code :/ 

https://github.com/ynput/ayon-houdini/pull/190/commits/84be3d435dd5ed92842c27e0c5fd42a8f5946ad2

client/ayon_houdini/plugins/publish/collect_arnold_rop.py
```diff
@@ -160,3 +161,12 @@ def replace(match):
                 os.path.join(dir, (file % i)).replace("\\", "/"))
 
         return expected_files
+
+    def evalParmNoFrame(self, rop, parm, **kwargs):
+        try:
+            return evalParmNoFrame(rop, parm, **kwargs)
+        except Exception as exc:
```
BigRoy, line 168: I vaguely recall that `hou.Error` does not inherit from `Exception` - is that still the case? If so, we may need to capture `hou.Error` explicitly if the library function does not?

client/ayon_houdini/api/lib.py
BigRoy, line 1: I assume all `lib` functions, especially generic ones like `get_output_parameter` should not raise `PublishError` and the original errors were better error types.

client/ayon_houdini/plugins/publish/collect_usd_layers.py
```diff
@@ -38,12 +39,15 @@ def copy_instance_data(instance_src, instance_dest, attr):
     keys = attr.split(".")
     for i, key in enumerate(keys):
         if key not in src_data:
-            break
+            raise PublishError(
+                f"key '{key}' does not exist on the source instance."
+            )
```
BigRoy, line 44: Why is this suddenly an error if previously this would `break`?

client/ayon_houdini/plugins/publish/collect_usd_render.py
```diff
@@ -68,3 +69,12 @@ def replace_to_f(match):
...
         instance.data["attachTo"] = []
+
+    def evalParmNoFrame(self, rop, parm, **kwargs):
+        try:
+            return evalParmNoFrame(rop, parm, **kwargs)
+        except Exception as exc:
+            raise PublishError(
+                f"Failed evaluating parameter '{parm}' on Rop node: {rop.path()}",
+                detail=f"{exc}"
+            )
```
BigRoy, line 80: What do you think of implementing this on `HoudiniInstancePlugin` base class?

client/ayon_houdini/plugins/publish/extract_render.py
```diff
@@ -81,7 +89,8 @@ def process(self, instance):
             if not os.path.exists(frame)
         ]
         if missing_frames:
-            # TODO: Use user friendly error reporting.
-            raise RuntimeError("Failed to complete render extraction. "
-                               "Missing output files: {}".format(
-                                   missing_frames))
+            missing_frames = "\n\n ●  ".join(missing_frames)
```
BigRoy, line 92: Please use regular dash?
```suggestion
            missing_frames = "\n\n -  ".join(missing_frames)
```

client/ayon_houdini/plugins/publish/extract_rop.py
```diff
@@ -34,7 +35,17 @@ def process(self, instance: pyblish.api.Instance):
         # This key might be absent because render targets are not yet implemented
         #  for all product types that use this plugin.
         if creator_attribute.get("render_target", "local") == "local":
-            self.render_rop(instance)
+            try:
+                self.render_rop(instance)
+            except Exception as e:
+                import hou
```
BigRoy, line 41: Why import `hou` here?

client/ayon_houdini/api/usd.py
```diff
@@ -205,8 +208,8 @@ def get_configured_save_layers(usd_rop, strip_above_layer_break=True):
     lop_node = get_usd_rop_loppath(usd_rop)
     stage = lop_node.stage(apply_viewport_overrides=False)
     if not stage:
-        raise RuntimeError(
-            "No valid USD stage for ROP node: " "%s" % usd_rop.path()
+        raise PublishError(
+            f"No valid USD stage for ROP node: '{usd_rop.path()}' "
```
BigRoy, line 212: This should not be a publish error - the previous type was better. This may be called outside of publishing if needed.

client/ayon_houdini/api/lib.py
MustafaJafar: on it. I'm going to revert changes in `lib`  one by one.

client/ayon_houdini/plugins/publish/collect_usd_layers.py
MustafaJafar: Tbh, this was a wild guess based on the doc string. 
should I put the `break` back and update the doc string?

client/ayon_houdini/plugins/publish/collect_usd_render.py
MustafaJafar: That's fair

client/ayon_houdini/plugins/publish/extract_render.py
MustafaJafar: You don't like my bullet point 😢. 
The publisher UI doesn't convert `-` to a `●` like we have in markdown.

client/ayon_houdini/api/usd.py
MustafaJafar: I think that is the case already. I've reverted the changes in `usd.py`

client/ayon_houdini/plugins/publish/extract_rop.py
MustafaJafar: To be able to call `hou.node`

client/ayon_houdini/plugins/publish/collect_usd_layers.py
MustafaJafar: I've reverted the changes in this function.

https://github.com/ynput/ayon-houdini/pull/190/commits/65249ccabf63925f7aa369ad001f550914ff2299

client/ayon_houdini/plugins/publish/extract_render.py
MustafaJafar: https://github.com/ynput/ayon-houdini/pull/190/commits/bde4ab355fe5a33651227f5557a66fe81a4b2fef

client/ayon_houdini/plugins/publish/collect_usd_render.py
MustafaJafar: https://github.com/ynput/ayon-houdini/pull/190/commits/0363a57b699b214a901ac79deab352d856f93e7e

client/ayon_houdini/plugins/publish/collect_arnold_rop.py
MustafaJafar: `evalParmNoFrame`  function has `assert` and `RuntimeError`  and they'll be raised before the `PublishError`.
I don't know if that is sufficient.

client/ayon_houdini/plugins/publish/validate_frame_token.py
moonyuet: any update on this comment? I agree with Roy that the message needs more clarification.

moonyuet (approved):
Looks good for me.

client/ayon_houdini/api/plugin.py
```diff
@@ -334,6 +348,34 @@ class HoudiniInstancePlugin(pyblish.api.InstancePlugin):
     hosts = ["houdini"]
     settings_category = SETTINGS_CATEGORY
 
+    def evalParmNoFrame(self, rop, parm, **kwargs):
+        try:
+            return evalParmNoFrame(rop, parm, **kwargs)
+        except Exception as exc:
+            raise PublishError(
+                f"Failed evaluating parameter '{parm}' on Rop node: {rop.path()}",
+                detail=f"{exc}"
```
BigRoy, line 357: The `f{exc}` seems weird.
I assume this doesn't work?
```suggestion
                detail=exc
```

If so, why not do:
```suggestion
                detail=str(exc)
```

client/ayon_houdini/plugins/publish/validate_animation_settings.py
BigRoy, line 1: Why's this deleted?

MustafaJafar: It was an exact duplicate.

client/ayon_houdini/api/plugin.py
```diff
@@ -334,6 +348,34 @@ class HoudiniInstancePlugin(pyblish.api.InstancePlugin):
...
+            )
+    
+    def get_output_parameter(self, node):
+        try:
+            return get_output_parameter(node)
+        except Exception as exc:
```
BigRoy, line 363: Technically we should only be capturing the `TypeError` here. Right?

client/ayon_houdini/api/plugin.py
```diff
@@ -334,6 +348,34 @@ class HoudiniInstancePlugin(pyblish.api.InstancePlugin):
     hosts = ["houdini"]
     settings_category = SETTINGS_CATEGORY
 
+    def evalParmNoFrame(self, rop, parm, **kwargs):
```
BigRoy, line 351: We should do a follow up PR potentially to refactor this to lower case, `eval_parm_no_frame` to adhere to the AYON code style?

client/ayon_houdini/plugins/publish/collect_karma_rop.py
```diff
@@ -28,7 +29,13 @@ def process(self, instance):
 
         rop = hou.node(instance.data.get("instance_node"))
 
-        default_prefix = evalParmNoFrame(rop, "picture")
+        try:
+            default_prefix = evalParmNoFrame(rop, "picture")
```
BigRoy, line 33: Should this also be `self.evalParmNoFrame`?

client/ayon_houdini/plugins/publish/collect_render_colorspace.py
```diff
@@ -32,14 +37,27 @@ def process(self, instance):
...
             return
         aov_name = list(expected_files[0].keys())
-        render_products_data = colorspace.ARenderProduct(aov_name)
+        try:
+            render_products_data = ARenderProduct(aov_name)
+        except Exception as exc:
+            raise PublishError(
+                "Failed to get render products with colorspace.",
+                detail=f"{exc}"
+            )
```
BigRoy, line 46: I really doubt that raising these errors is an improvement over _just letting it error_. The raised errors tells me basically nothing about _what_ caused it exactly and hence provides me as a developer nor as an artist the means to go solve it. Hence, the original error's exception must be (hopefully) clearer as to what is going on or failing.

I feel like our raised error is obfuscating the original error, no?

Also, note that you can do `raise PublishError("message") from exc` or `raise PublishError("message").with_traceback(exc.__traceback__)` although that's maybe a bit verbose. But I guess that's not raising an error in a way that looks nice in the Publisher UI.

What do we think @iLLiCiTiT @dee-ynput - are we obfuscating the errors and just making the code more complex or does this make sense? (Or should we instead still be changing how errors appear in the UI if NOT raised as `PublishError` so that a _regular_ error occurring in code still looks better). I'm a bit worried we'll now be getting tons of `PublishError("error occurred")` like errors added to the code just to provide the stack trace in the end to the UI instead of 'keeping it away from artists'.

Are we battling our own UI decisions here?

Take a look at the changes in this PR - and _what_ does it actually clarify to the raised errors, etc. whilst adding a lot more code for essentially raising the same error, but then as `PublishError` instead of the _actual_ error that occurred.

MustafaJafar: For information, the current status of this PR aims to do the following: 
1. It raises publish errors instead of assert and other errors (from the caller functions in publish plugins).
2. It shows the original error as `detail` in publish error.
3. Enhance some error messages.

Personally I believe that `PublishError` should be used to guide the users about what could be wrong in their scene and how to fix it. e.g. `missing frames` -> `tells the users to ensure that all frames exist` 

Maybe I abused it a little bit to capture the potential errors that could happen from the called functions so that they show up nicely.

client/ayon_houdini/api/plugin.py
MustafaJafar: Should I do it in this PR instead?

client/ayon_houdini/api/plugin.py
MustafaJafar: https://github.com/ynput/ayon-houdini/pull/190/commits/a03e49bc665970ec532140908df82604537e7ae6

client/ayon_houdini/api/plugin.py
MustafaJafar: yes. if the `node.parm(...)` didn't failed. and I think it won't fail.

client/ayon_houdini/plugins/publish/collect_karma_rop.py
MustafaJafar: https://github.com/ynput/ayon-houdini/pull/190/commits/d92689e719a3147c23b59843ace486d627894aeb

client/ayon_houdini/api/plugin.py
MustafaJafar: created an issue for it
https://github.com/ynput/ayon-houdini/issues/191

MustafaJafar:
closing in favor of https://github.com/ynput/ayon-houdini/pull/193

PR #193, opened 2024-12-12
User friendly error messages

### LINKED ISSUE #108: Refactor `RuntimeError`s to `PublishError`s

MustafaJafar:
## Changelog Description
resolve #108 

These changes affect only publish plugins where we:
1. Update the error messages.
2. Errors related to user should raise `PublishError` with a clear message about what a user should do to fix the problem.
3. Errors that are not related user should raise `KnownPublishError` (This change might be reverted).
4. Avoid capturing any errors from lib functions e.g. `get_output_parameter`, `render_rop` and etc.

## Testing notes:
1. Everything should still work as before
2. Error messages related to the changes should be better.

### DIFF
--- a/client/ayon_houdini/api/lib.py
+++ b/client/ayon_houdini/api/lib.py
@@ -5,6 +5,7 @@ import errno
 import re
 import logging
 import json
+import clique
 from contextlib import contextmanager
 
 import six
@@ -1596,3 +1597,10 @@ def connect_file_parm_to_loader(file_parm: hou.Parm):
         f' {file_parm.node().path()} {file_parm.name()} \`{expression}\`'
     )
     show_node_parmeditor(load_node)
+
+
+def format_as_collections(files: list[str], pattern: str = "{head}{padding}{tail} [{ranges}]") -> list[str]:
+    collections, remainder = clique.assemble(files)
+    result = [collection.format(pattern) for collection in collections]
+    result.extend(remainder)
+    return result
--- a/client/ayon_houdini/api/lib.py
+++ b/client/ayon_houdini/api/lib.py
@@ -1600,6 +1600,8 @@ def connect_file_parm_to_loader(file_parm: hou.Parm):
 
 
 def format_as_collections(files: list[str], pattern: str = "{head}{padding}{tail} [{ranges}]") -> list[str]:
+    """Return list of files as formatted sequence collections."""
+    
     collections, remainder = clique.assemble(files)
     result = [collection.format(pattern) for collection in collections]
     result.extend(remainder)
--- a/client/ayon_houdini/plugins/publish/collect_componentbuilder_lop.py
+++ b/client/ayon_houdini/plugins/publish/collect_componentbuilder_lop.py
@@ -30,22 +30,23 @@ class CollectComponentBuilderLOPs(plugin.HoudiniInstancePlugin,
 
         node = hou.node(instance.data["instance_node"])
 
-        try:
-            node.cook(force=True)  # required to clear existing errors
-        except hou.Error as exc:
-            errors = node.errors()
-            if errors:
-                errors = "\n\n - ".join(errors)
-                raise PublishError(
-                    f"Failed to cook node: '{node.path()}'. Please fix it to proceed.",
-                    detail=f"{exc} \n\n - {errors}"
-                )
-
         # Render the component builder LOPs
         # TODO: Do we want this? or use existing frames? Usually a Collector
         #  should not 'extract' but in this case we need the resulting USD
         #  file.
+        node.cook(force=True)  # required to clear existing errors
         node.parm("execute").pressButton()
+
+        errors = node.errors()
+        if errors:
+            for error in errors:
+                self.log.error(error)
+            raise PublishError(
+                f"Failed to save to disk '{node.path()}'. "
+                "Please fix your scene to ensure it renders correctly "
+                "and re-publish. Check the log for more information."
+            )
+
         # Define the main asset usd file
         filepath = node.evalParm("lopoutput")
         representations = instance.data.setdefault("representations", [])
--- a/client/ayon_houdini/plugins/publish/collect_componentbuilder_lop.py
+++ b/client/ayon_houdini/plugins/publish/collect_componentbuilder_lop.py
@@ -30,23 +30,22 @@ class CollectComponentBuilderLOPs(plugin.HoudiniInstancePlugin,
 
         node = hou.node(instance.data["instance_node"])
 
-        # Render the component builder LOPs
-        # TODO: Do we want this? or use existing frames? Usually a Collector
-        #  should not 'extract' but in this case we need the resulting USD
-        #  file.
         try:
             node.cook(force=True)  # required to clear existing errors
-            node.parm("execute").pressButton()
-        except Exception as exc:
+        except hou.Error as exc:
             errors = node.errors()
             if errors:
                 errors = "\n\n - ".join(errors)
                 raise PublishError(
-                    "Failed to save to disk."
-                    f"Please fix node: '{node.path()}'",
+                    f"Failed to cook node: '{node.path()}'. Please fix it to proceed.",
                     detail=f"{exc} \n\n - {errors}"
                 )
 
+        # Render the component builder LOPs
+        # TODO: Do we want this? or use existing frames? Usually a Collector
+        #  should not 'extract' but in this case we need the resulting USD
+        #  file.
+        node.parm("execute").pressButton()
         # Define the main asset usd file
         filepath = node.evalParm("lopoutput")
         representations = instance.data.setdefault("representations", [])
--- a/client/ayon_houdini/plugins/publish/collect_componentbuilder_lop.py
+++ b/client/ayon_houdini/plugins/publish/collect_componentbuilder_lop.py
@@ -34,14 +34,18 @@ class CollectComponentBuilderLOPs(plugin.HoudiniInstancePlugin,
         # TODO: Do we want this? or use existing frames? Usually a Collector
         #  should not 'extract' but in this case we need the resulting USD
         #  file.
-        node.cook(force=True)  # required to clear existing errors
-        node.parm("execute").pressButton()
-
-        errors = node.errors()
-        if errors:
-            for error in errors:
-                self.log.error(error)
-            raise PublishError(f"Failed to save to disk '{node.path()}'")
+        try:
+            node.cook(force=True)  # required to clear existing errors
+            node.parm("execute").pressButton()
+        except Exception as exc:
+            errors = node.errors()
+            if errors:
+                errors = "\n\n - ".join(errors)
+                raise PublishError(
+                    "Failed to save to disk."
+                    f"Please fix node: '{node.path()}'",
+                    detail=f"{exc} \n\n - {errors}"
+                )
 
         # Define the main asset usd file
         filepath = node.evalParm("lopoutput")
--- a/client/ayon_houdini/plugins/publish/collect_usd_layers.py
+++ b/client/ayon_houdini/plugins/publish/collect_usd_layers.py
@@ -4,6 +4,7 @@ import re
 
 import pyblish.api
 
+from ayon_core.pipeline import KnownPublishError
 from ayon_core.pipeline.create import get_product_name
 from ayon_houdini.api import plugin
 import ayon_houdini.api.usd as usdlib
@@ -27,8 +28,7 @@ def copy_instance_data(instance_src, instance_dest, attr):
             in the source instance's data.
 
     Raises:
-        KeyError: If the key does not exist on the source instance.
-        AssertionError: If a parent key already exists on the destination
+        KnownPublishError: If a parent key already exists on the destination
             instance but is not of the correct type (= is not a dict)
 
     """
@@ -43,7 +43,8 @@ def copy_instance_data(instance_src, instance_dest, attr):
         src_value = src_data[key]
         if i != len(key):
             dest_data = dest_data.setdefault(key, {})
-            assert isinstance(dest_data, dict), "Destination must be a dict"
+            if not isinstance(dest_data, dict):
+                raise KnownPublishError("Destination must be a dict")
             src_data = src_value
         else:
             # Last iteration - assign the value
--- a/client/ayon_houdini/plugins/publish/extract_render.py
+++ b/client/ayon_houdini/plugins/publish/extract_render.py
@@ -6,6 +6,7 @@ import pyblish.api
 
 from ayon_core.pipeline import PublishError
 from ayon_houdini.api import plugin
+from ayon_houdini.api.lib import format_as_collections
 
 
 class ExtractRender(plugin.HoudiniExtractorPlugin):
@@ -82,15 +83,13 @@ class ExtractRender(plugin.HoudiniExtractorPlugin):
             for frame in all_frames
             if not os.path.exists(frame)
         ]
-        missing_frames , _ = clique.assemble(
-            missing_frames,
-            patterns=[clique.PATTERNS["frames"]],
-            minimum_items=1
-        )
-        missing_frames = "\n - ".join(f"{sequence}" for sequence in missing_frames)
+
         if missing_frames:
+            # Combine collections for simpler logs of missing files
+            missing_frames  = format_as_collections(missing_frames)
+            missing_frames = "\n ".join(f"- {sequence}" for sequence in missing_frames)
             raise PublishError(
                 "Failed to complete render extraction.\n"
                 "Please render any missing output files.",
-                detail=f"Missing output files: \n - {missing_frames}"
+                detail=f"Missing output files: \n {missing_frames}"
             )
--- a/client/ayon_houdini/plugins/publish/extract_render.py
+++ b/client/ayon_houdini/plugins/publish/extract_render.py
@@ -87,10 +87,10 @@ class ExtractRender(plugin.HoudiniExtractorPlugin):
             patterns=[clique.PATTERNS["frames"]],
             minimum_items=1
         )
-        missing_frames = "\n\n - ".join(f"{sequence}" for sequence in missing_frames)
+        missing_frames = "\n - ".join(f"{sequence}" for sequence in missing_frames)
         if missing_frames:
             raise PublishError(
                 "Failed to complete render extraction.\n"
                 "Please render any missing output files.",
-                detail=f"Missing output files: \n\n - {missing_frames}"
+                detail=f"Missing output files: \n - {missing_frames}"
             )
--- a/client/ayon_houdini/plugins/publish/extract_render.py
+++ b/client/ayon_houdini/plugins/publish/extract_render.py
@@ -1,5 +1,4 @@
 import os
-import clique
 import hou
 
 import pyblish.api
--- a/client/ayon_houdini/plugins/publish/extract_render.py
+++ b/client/ayon_houdini/plugins/publish/extract_render.py
@@ -3,6 +3,7 @@ import hou
 
 import pyblish.api
 
+from ayon_core.pipeline import PublishError
 from ayon_houdini.api import plugin
 
 
@@ -75,13 +76,14 @@ class ExtractRender(plugin.HoudiniExtractorPlugin):
                 all_frames.extend(value)
         # Check missing frames.
         # Frames won't exist if user cancels the render.
-        missing_frames = [
+        missing_frames = "\n\n - ".join(
             frame
             for frame in all_frames
             if not os.path.exists(frame)
-        ]
+        )
         if missing_frames:
-            # TODO: Use user friendly error reporting.
-            raise RuntimeError("Failed to complete render extraction. "
-                               "Missing output files: {}".format(
-                                   missing_frames))
+            raise PublishError(
+                "Failed to complete render extraction.\n"
+                "Please render any missing output files.",
+                detail=f"Missing output files: \n\n - {missing_frames}"
+            )
--- a/client/ayon_houdini/plugins/publish/extract_render.py
+++ b/client/ayon_houdini/plugins/publish/extract_render.py
@@ -1,4 +1,5 @@
 import os
+import clique
 import hou
 
 import pyblish.api
@@ -76,11 +77,17 @@ class ExtractRender(plugin.HoudiniExtractorPlugin):
                 all_frames.extend(value)
         # Check missing frames.
         # Frames won't exist if user cancels the render.
-        missing_frames = "\n\n - ".join(
+        missing_frames = [
             frame
             for frame in all_frames
             if not os.path.exists(frame)
+        ]
+        missing_frames , _ = clique.assemble(
+            missing_frames,
+            patterns=[clique.PATTERNS["frames"]],
+            minimum_items=1
         )
+        missing_frames = "\n\n - ".join(f"{sequence}" for sequence in missing_frames)
         if missing_frames:
             raise PublishError(
                 "Failed to complete render extraction.\n"
--- a/client/ayon_houdini/plugins/publish/extract_usd.py
+++ b/client/ayon_houdini/plugins/publish/extract_usd.py
@@ -3,7 +3,7 @@ from typing import List, AnyStr
 
 import pyblish.api
 
-from ayon_core.pipeline import KnownPublisherError, PublishError
+from ayon_core.pipeline import KnownPublishError, PublishError
 from ayon_core.pipeline.entity_uri import construct_ayon_entity_uri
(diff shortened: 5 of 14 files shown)


client/ayon_houdini/plugins/publish/collect_componentbuilder_lop.py
```diff
@@ -34,14 +34,18 @@ def process(self, instance):
...
-
-        errors = node.errors()
-        if errors:
-            for error in errors:
-                self.log.error(error)
-            raise PublishError(f"Failed to save to disk '{node.path()}'")
+        try:
+            node.cook(force=True)  # required to clear existing errors
+            node.parm("execute").pressButton()
+        except Exception as exc:
```
BigRoy, line 40: Why is this needed?

client/ayon_houdini/plugins/publish/extract_usd.py
```diff
@@ -3,6 +3,7 @@
 
 import pyblish.api
 
+from ayon_core.pipeline import KnownPublisherError, PublishError
```
moonyuet, line 6:
```suggestion
from ayon_core.pipeline import KnownPublishError, PublishError
```

client/ayon_houdini/plugins/publish/extract_usd.py
```diff
@@ -135,5 +137,6 @@ def get_source_paths(
         # Single file
         return [os.path.join(staging, files)]
 
-    raise TypeError(f"Unsupported type for representation files: {files} "
-                    "(supports list or str)")
+    raise KnownPublisherError(
```
moonyuet, line 140:
```suggestion
    raise KnownPublishError(
```

client/ayon_houdini/plugins/publish/collect_usd_layers.py
```diff
@@ -43,7 +43,8 @@ def copy_instance_data(instance_src, instance_dest, attr):
         src_value = src_data[key]
         if i != len(key):
             dest_data = dest_data.setdefault(key, {})
-            assert isinstance(dest_data, dict), "Destination must be a dict"
+            if not isinstance(dest_data, dict):
+                raise KnownPublishError("Destination must be a dict")
```
moonyuet, line 47:
```suggestion
                raise KnownPublishError("Destination must be a dict.")
```

client/ayon_houdini/plugins/publish/increment_current_file.py
```diff
@@ -46,9 +48,19 @@ def process(self, context):
         host = registered_host()
         current_file = host.current_file()
         if context.data["currentFile"] != current_file:
-            raise KnownPublishError(
-                "Collected filename mismatches from current scene name."
+            raise PublishError(
```
moonyuet, line 51: Just out of curiosity why there is a specific check on the currentFile in both increment and save file? Also why do we need to save the workfile twice for render and workfile families?

client/ayon_houdini/plugins/publish/save_scene.py
```diff
@@ -16,12 +17,23 @@ def process(self, context):
         # Filename must not have changed since collecting
         host = registered_host()
         current_file = host.get_current_workfile()
-        assert context.data['currentFile'] == current_file, (
-            "Collected filename from current scene name."
-        )
-
+        if context.data['currentFile'] != current_file:
```
moonyuet, line 20: Same question as mentioned in https://github.com/ynput/ayon-houdini/pull/193/files/a0bb6163672fbd6a6078e8f48e8dc61615150709#r1883402173

client/ayon_houdini/plugins/publish/collect_componentbuilder_lop.py
MustafaJafar: When it fails to cook, it throws this error: 
(it happens a little before `raise PublishError`)
```python
  File "C:\Users\Mustafa Taher\AppData\Local\Ynput\AYON\dependency_packages\ayon_2411271645_windows.zip\dependencies\pyblish\plugin.py", line 528, in __explicit_process
    runner(*args)
  File "E:\Ynput\ayon-houdini\client\ayon_houdini\plugins\publish\collect_componentbuilder_lop.py", line 38, in process
    node.cook(force=True)  # required to clear existing errors
    ^^^^^^^^^^^^^^^^^^^^^
  File "C:\PROGRA~1/SIDEEF~1/HOUDIN~1.370/houdini/python3.11libs\hou.py", line 17842, in cook
    return _hou.OpNode_cook(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error while cooking.
```

client/ayon_houdini/plugins/publish/extract_usd.py
MustafaJafar: https://github.com/ynput/ayon-houdini/pull/193/commits/04bbcc07fbaf530798d0b2d821c44d9e3ff9f4f8

client/ayon_houdini/plugins/publish/extract_usd.py
MustafaJafar: https://github.com/ynput/ayon-houdini/pull/193/commits/04bbcc07fbaf530798d0b2d821c44d9e3ff9f4f8

client/ayon_houdini/plugins/publish/increment_current_file.py
MustafaJafar: Honestly, I don't know. 
And, not only that but a lot of  plugins have the same checks e.g. alembic validators. 
I've created this issue yesterday 
also, I've added your note. thank you.
https://github.com/ynput/ayon-houdini/issues/194

BigRoy: > Just out of curiosity why there is a specific check on the currentFile in both increment and save file? Also why do we need to save the workfile twice for render and workfile families?

Where are we saving twice?

We're saving (before incrementing)... and then incrementing the workfile so the user doesn't continue working in the original version.

---

Also, they both check the same thing to ensure that if the plug-ins would be operating in isolation of each other this thing is checked, because they don't NEED to run together.

client/ayon_houdini/plugins/publish/collect_componentbuilder_lop.py
BigRoy: Does this work? I vaguely recall `hou.Error` not inheriting from Python exceptions, hence not getting caught.

Also, I'm not fan of capturing this as globally as this.

Should we be try excepting the `node.cook()` itself, and not the full executing of the parm too - just so we're more explicitly capturing that it's _that_ cook that's failing - and not the render cook.

Technically the user may e.g. currently be on frame 1200 but the ROP set to 1001-1100. A cook on 1200 may hence fail, even though the render could technically exists.

It may be even better if we can improve the code and find a way to 'clear the errors' without forcing a cook of course.

Let's be very explicit about what error we're capturing here and only try..except the `node.cook()` itself.

client/ayon_houdini/plugins/publish/extract_render.py
```diff
@@ -75,13 +76,14 @@ def process(self, instance):
                 all_frames.extend(value)
         # Check missing frames.
         # Frames won't exist if user cancels the render.
-        missing_frames = [
+        missing_frames = "\n\n - ".join(
```
BigRoy, line 79: Why do we need two new lines for each individual line, that seems quite overkill?

client/ayon_houdini/plugins/publish/validate_frame_token.py
```diff
@@ -42,11 +43,11 @@ def get_invalid(cls, instance):
...
-            return []
+            return
 
         output_parm = lib.get_output_parameter(node)
         unexpanded_str = output_parm.unexpandedString()
 
         if "$F" not in unexpanded_str:
             cls.log.error("No frame token found in '%s'" % node.path())
-            return [instance]
+            return [node]
```
BigRoy, line 53: This seems like a code change - it may fix a bug of course, but it doesn't sound very related to the PR title?

client/ayon_houdini/plugins/publish/extract_render.py
MustafaJafar: Yes, unfortunately, we need two new lines in order to get one new line in the publish report.
image

client/ayon_houdini/plugins/publish/increment_current_file.py
MustafaJafar: Should we refactor the error in `IncrementCurrentFile` to `KnownPublishError`. 
as a user I can change the file name between collection (validation to be percise) and before extraction by clicking validate button. and change the file then continue publishing. 

But, during integration (`IncrementCurrentFile`), I'm not sure how can I break it.

MustafaJafar: Also, I'm thinking about changing the comment in the code to 
```
# Filename must not have changed since extraction.
```

client/ayon_houdini/plugins/publish/validate_frame_token.py
MustafaJafar: Thank you. it should be done in another PR.

MustafaJafar: https://github.com/ynput/ayon-houdini/pull/193/commits/ed447df6f63e8a3912b53487a679fc0205e0117d

client/ayon_houdini/plugins/publish/collect_componentbuilder_lop.py
```diff
@@ -30,19 +30,22 @@ def process(self, instance):
 
         node = hou.node(instance.data["instance_node"])
 
+        try:
+            node.cook(force=True)  # required to clear existing errors
+        except hou.Error as exc:
+            errors = node.errors()
```
BigRoy, line 36: This technically is incorrect as far as I know. As far as I'm aware a node _may_ have errors even though the cook itself succeeded.

As such, the `node.errors()` check should actually (or STILL) be done after the `node.parm("execute").pressButton()` ;)

client/ayon_houdini/plugins/publish/collect_componentbuilder_lop.py
```diff
@@ -30,19 +30,22 @@ def process(self, instance):
...
+                    f"Failed to cook node: '{node.path()}'. Please fix it to proceed.",
+                    detail=f"{exc} \n\n - {errors}"
+                )
+
         # Render the component builder LOPs
         # TODO: Do we want this? or use existing frames? Usually a Collector
         #  should not 'extract' but in this case we need the resulting USD
         #  file.
-        node.cook(force=True)  # required to clear existing errors
         node.parm("execute").pressButton()
```
BigRoy, line 48:
```suggestion
        node.parm("execute").pressButton()
        errors = node.errors()
        if errors:
            for error in errors:
                self.log.error(error)
            raise PublishError(f"Failed to save to disk '{node.path()}'")
```

client/ayon_houdini/plugins/publish/collect_componentbuilder_lop.py
```diff
@@ -30,19 +30,22 @@ def process(self, instance):
...
+        try:
+            node.cook(force=True)  # required to clear existing errors
+        except hou.Error as exc:
+            errors = node.errors()
+            if errors:
+                errors = "\n\n - ".join(errors)
+                raise PublishError(
+                    f"Failed to cook node: '{node.path()}'. Please fix it to proceed.",
+                    detail=f"{exc} \n\n - {errors}"
+                )
```
BigRoy, line 42:
```suggestion
                ) from exc
```
How does this look in UI if we raise the error from the original exception?

client/ayon_houdini/plugins/publish/increment_current_file.py
BigRoy: Both sound good.

client/ayon_houdini/plugins/publish/extract_render.py
BigRoy: @iLLiCiTiT this feels fishy, right? It should just be fine with a single new line?

Also, I really think we should improve how these files are listed here - maybe we're better off actually printing them as collections? (if they are detected as collections).

client/ayon_houdini/plugins/publish/collect_componentbuilder_lop.py
moonyuet: This seems to be better as sometimes the cooking time can be taken very long (Pretty sure it happened in bgeo/vdb not sure about LOP though)

MustafaJafar: The main reason for  `try..except..`  clause was to capture the error which make it show up nicely in the publisher. 
So, will it be the same case for`node.errors()` ?

client/ayon_houdini/plugins/publish/increment_current_file.py
MustafaJafar: https://github.com/ynput/ayon-houdini/pull/193/commits/25850ea18a7a3c3326edcadfed693b5d94a3f391

client/ayon_houdini/plugins/publish/extract_render.py
MustafaJafar: > Also, I really think we should improve how these files are listed here - maybe we're better off actually printing them as collections? (if they are detected as collections).

`all_frames` object at this point of execution is a flat list of files.

client/ayon_houdini/plugins/publish/collect_componentbuilder_lop.py
MustafaJafar: I can't see any difference. 
What should I expect? 
image

client/ayon_houdini/plugins/publish/collect_componentbuilder_lop.py
MustafaJafar: This is the original file from develop without the `node.cook(force=True)` 
image

BigRoy: The issue was originally with the `node.cook()` removed is that if there were errors (e.g. make it break now in your local scene)... then fix your scene so it should render correctly, and then just publish but don't recook the node manually that the original errors would still be listed from `node.errors()`

The comment in code mentions that to "force a cook" purely to clear out EXISTING errors prior to the render.

client/ayon_houdini/plugins/publish/collect_componentbuilder_lop.py
BigRoy: See the difference of the stack trace in the Houdini Console - it includes information of the original raised error instead of only the one we raised. If it doesn't influence the UI let's add it so that at least the stack trace provides more info on what error actually caused it.

client/ayon_houdini/plugins/publish/collect_componentbuilder_lop.py
MustafaJafar: ehm.. I fixed my scene and restarted the publish and it didn't complain. 
image

client/ayon_houdini/plugins/publish/extract_render.py
BigRoy: I meant something like this:
```python
collections, remainder = clique.assemble(files)
collections.extend(remainder)
collections_str = "\n".join(f"- {sequence}" for sequence in collections)
```
So that anything that is considered a sequence would just be reported as a single sequence.

client/ayon_houdini/plugins/publish/collect_componentbuilder_lop.py
BigRoy: If that's the case - let's make a separate PR (sorry, I just want to avoid this PR making more than cosmetic changes) and remove the cooking.

The issue we face initially was that if the ROP or graph had previously errored, that just doing the execute button would fail because it wouldn't resolve that the problem has been 'fixed' since. (It may have been that the Houdini session was set to "Update on mouse up" or whatever - but if we can't reproduce it. Then let's clean it up and remove the cook.

client/ayon_houdini/plugins/publish/collect_componentbuilder_lop.py
MustafaJafar: printing the error in log is better than using the left side of the publisher window.

MustafaJafar: https://github.com/ynput/ayon-houdini/pull/193/commits/6513a86567e88c9cd983f9c921c6803b960c34ad

client/ayon_houdini/plugins/publish/collect_componentbuilder_lop.py
MustafaJafar: https://github.com/ynput/ayon-houdini/pull/193/commits/6513a86567e88c9cd983f9c921c6803b960c34ad

client/ayon_houdini/plugins/publish/collect_componentbuilder_lop.py
MustafaJafar: https://github.com/ynput/ayon-houdini/pull/193/commits/6513a86567e88c9cd983f9c921c6803b960c34ad

client/ayon_houdini/plugins/publish/extract_render.py
MustafaJafar: https://github.com/ynput/ayon-houdini/pull/193/commits/33568f8bfcd862f816b938e5d38f3b93fe80b11c
image

BigRoy: The way you implemented it - it would never report the `remainder` which is technically incorrect because those would still be frames it was expecting and had missing. It may happen if for whatever reason `clique` is unable to detect the sequence, e.g. with the current call it won't detect a sequence if it's a single frame (which is fine because we don't need to list those as a sequence). However, we do still want to list is.

That's why in my psuedocode I added them to the list of collections (which is weird due to the variable name of course) but this might make more sense:
```
missing_file_labels = [str(collection) for collection in collections]
missing_file_labels.extend(remainder)
```
And then reporting these file labels.

client/ayon_houdini/plugins/publish/extract_render.py
```diff
@@ -80,8 +82,15 @@ def process(self, instance):
...
         if missing_frames:
-            # TODO: Use user friendly error reporting.
-            raise RuntimeError("Failed to complete render extraction. "
-                               "Missing output files: {}".format(
-                                   missing_frames))
+            raise PublishError(
+                "Failed to complete render extraction.\n"
+                "Please render any missing output files.",
+                detail=f"Missing output files: \n\n - {missing_frames}"
+            )
```
BigRoy, line 96:
```suggestion
        # Combine collections for simpler logs of missing files
        collections, remainder = clique.assemble(
            missing_frames,
            patterns=[clique.PATTERNS["frames"]]
        )
        missing = [str(collection) for collection in collections]
        missing.extend(remainder)
        missing = "\n\n".join(f"- {sequence}" for sequence in missing)
            raise PublishError(
                "Failed to complete render extraction.\n"
                "Please render any missing output files.",
                detail=f"Missing output files: \n\n{missing}"
            )
```

client/ayon_houdini/plugins/publish/extract_render.py
MustafaJafar: > It may happen if for whatever reason `clique` is unable to detect the sequence, e.g. with the current call it won't detect a sequence if it's a single frame

For this particular case, I've used `minimum_items=1`
```python
missing_frames , _ = clique.assemble(
            missing_frames,
            patterns=[clique.PATTERNS["frames"]],
            minimum_items=1
        )
```

iLLiCiTiT: > @iLLiCiTiT this feels fishy, right? It should just be fine with a single new line?

It should. How it does look like with single new line then?

MustafaJafar: @iLLiCiTiT 
Here's the output of using only one `\n` and no `-` in `detail`
image
The output of combining `\n` and `-` in `detail`
image
The output of using two `\n`
image

iLLiCiTiT: Can you please create issue in ayon-core then?

MustafaJafar: https://github.com/ynput/ayon-core/issues/1062

client/ayon_houdini/plugins/publish/extract_render.py
MustafaJafar: https://github.com/ynput/ayon-houdini/pull/193/commits/de655f422b6ed845fbcf25cb8a523add5862abde

client/ayon_houdini/plugins/publish/extract_render.py
```diff
@@ -80,8 +82,15 @@ def process(self, instance):
             for frame in all_frames
             if not os.path.exists(frame)
         ]
+        missing_frames , _ = clique.assemble(
+            missing_frames,
+            patterns=[clique.PATTERNS["frames"]],
+            minimum_items=1
+        )
+        missing_frames = "\n\n - ".join(f"{sequence}" for sequence in missing_frames)
```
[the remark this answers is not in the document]
MustafaJafar: https://github.com/ynput/ayon-houdini/pull/193/commits/de655f422b6ed845fbcf25cb8a523add5862abde

MustafaJafar:
For reference, here are some screenshots of the result of my last commits.
image
image

client/ayon_houdini/plugins/publish/increment_current_file.py
```diff
@@ -42,12 +42,13 @@ def process(self, context):
                 "submission to deadline failed."
             )
 
-        # Filename must not have changed since collecting
+        # Filename must not have changed since extraction.
```
BigRoy, line 45: I understand why you changed the comment - it's because there's a Save current scene extractor that ALSO happens to check something similar. Nonetheless, it checks here against the "collected" value. It's technically irrelevant whether the other extractor ran or not. So to me, the original comment was actually more correct.

```suggestion
        # Filename must not have changed since collecting.
```

client/ayon_houdini/plugins/publish/save_scene.py
```diff
@@ -16,12 +17,23 @@ def process(self, context):
...
         if host.workfile_has_unsaved_changes():
             self.log.info("Saving current file: {}".format(current_file))
             host.save_workfile(current_file)
         else:
             self.log.debug("No unsaved changes, skipping file save..")
+
+
+    def get_error_description(self):
+        return inspect.cleandoc(
+            """### Scene File Name Change During Publishing
```
BigRoy, line 35:
```suggestion
            """### Scene File Name Changed During Publishing
```

client/ayon_houdini/api/lib.py
```diff
@@ -1596,3 +1597,10 @@ def connect_file_parm_to_loader(file_parm: hou.Parm):
         f' {file_parm.node().path()} {file_parm.name()} \`{expression}\`'
     )
     show_node_parmeditor(load_node)
+
+
+def format_as_collections(files: list[str], pattern: str = "{head}{padding}{tail} [{ranges}]") -> list[str]:
```
BigRoy, line 1602: Let's add a small docstring
```suggestion
def format_as_collections(files: list[str], pattern: str = "{head}{padding}{tail} [{ranges}]") -> list[str]:
    """Return list of files as formatted sequence collections."""
```

BigRoy (approved):
Looks good to me - just some minor cosmetic notes.

client/ayon_houdini/api/lib.py
MustafaJafar: https://github.com/ynput/ayon-houdini/pull/193/commits/78e15a5ed327a27eabacc3dfa688d3d39c6e1557

client/ayon_houdini/plugins/publish/save_scene.py
MustafaJafar: https://github.com/ynput/ayon-houdini/pull/193/commits/78e15a5ed327a27eabacc3dfa688d3d39c6e1557

client/ayon_houdini/plugins/publish/increment_current_file.py
MustafaJafar: https://github.com/ynput/ayon-houdini/pull/193/commits/78e15a5ed327a27eabacc3dfa688d3d39c6e1557

moonyuet (approved):
Looks good to me after testing in Houdini

PR #196, opened 2024-12-15
Add version up workfile menu entry when enabled in core settings

BigRoy:
## Changelog Description

Add version up workfile menu entry when enabled in core settings

Similar to Blender PR https://github.com/ynput/ayon-blender/pull/84 but for Houdini
As requested on discord [here](https://discord.com/channels/517362899170230292/563751989075378201/1317086847238930434).

Note that it currently only saves if there are unsaved changes - this core PR https://github.com/ynput/ayon-core/pull/1054 would change it to always save.

Setting disabled:
image

Setting enabled:
image

## Testing notes:

1. Enable core setting: `ayon+settings://core/tools/ayon_menu/version_up_current_workfile`
2. When enabled the menu entry should show in Houdini after houdini startup.
3. Version up should work

### DIFF
--- a/client/ayon_houdini/api/lib.py
+++ b/client/ayon_houdini/api/lib.py
@@ -5,6 +5,7 @@ import errno
 import re
 import logging
 import json
+from functools import lru_cache
 from contextlib import contextmanager
 
 import six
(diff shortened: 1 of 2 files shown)


moonyuet (approved):
Works. Looks good for me. 
image

MustafaJafar (approved):
It works. 
Unfortunately, it's not dynamic. 
The expression is computed once on starting Houdini.
i.e. when changing the setting in core. we'd need to restart Houdini to refresh main menu.

`context` is dynamic but it's used to [Houdini contexts](https://www.sidefx.com/docs/houdini/basics/config_menus.html#context-menu-files) only, 

also [menurefresh](https://www.sidefx.com/docs/houdini/commands/menurefresh.html) won't help :(

BigRoy:
> It works.
> Unfortunately, it's not dynamic.
> The expression is computed once on starting Houdini.
> Therefore, when changing the setting in core. we'd need to restart Houdini to refresh main menu.

That's actually intentional - I cached it to avoid needing to re-query project settings on each menu open. This really doesn't feel like a setting that should be live. It's also not live in other DCCs so it matches that correctly.

MustafaJafar:
> I cached it to avoid needing to re-query project settings on each menu open.

I made few test runs and Houdini only query it once. and it seems there is no way to refresh the main menu.

PR #199, opened 2024-12-17
Use Generic Loader as internal building block for Load Asset/Shot LOP

### LINKED ISSUE #165: Load Asset/Shot LOP use Generic Loader as internal building block

With the Generic Loader implementation now merged it's time to refactor the Load Asset LOP and Load Shot LOP to also use it.

### Suggested implementation?

- [ ] Use the Generic Loader internally to Load Asset LOP and Load Shot LOP
- [ ] Add support to still manage the Thumbnail on the "outer node", e.g. using a `thumbnail_node` parm that thumbnails are applied to that node instead of the generic loader internally. The default value for that parm would then be the node itself. (We could technically spawn of thumbnail also into a separate node, but I wouldn't bother with that for now)
- [ ] Make sure LOPs context options still correctly work.
- [ ] Make sure the changes are backwards compatible.
- [ ] The Loader and Scene Inventory (AYON > Load... and AYON > Manage...) should still work and nicely detect them
    - [ ] This of course also needs to be backwards compatible with existing LOPs, etc.

_No response_

MustafaJafar:
## Changelog Description
resolve #165 

TODO: 
- [x] Use `Generic Loader` as internal building block for `lop_import`
- [ ] Use `Generic Loader` as internal building block for `Load Shot`
- [ ] Fix double listing in Scene Inventory (AYON > Load... and AYON > Manage...). it should only list the Load Asset/Shot.
- [ ] ...

## Testing notes:
1. `lop_import` should work as before without noticing any difference.
2. `lop_import` should update seamlessly when opening older scenes.


MustafaJafar:
I wonder how this should be fixed. 
image

moonyuet:
Tested LOP import with alembic import(very simple one). Works smoothly
image
Scene inventory also working too.

MustafaJafar:
Here's a demo for testing context options and thumbnails.

https://github.com/user-attachments/assets/4fca9205-d489-47b7-bfb7-2cf3a5ebfeaf

client/ayon_houdini/api/hda_utils.py
```diff
@@ -808,7 +813,9 @@ def get_available_representations(node):
     representations = get_representations(
             project_name,
             version_ids={version_entity["id"]},
-            fields={"name"}
+            fields={"name"},
+            representation_names=ALLOWED_USD_REPRESENTATIONS \
+                if node.parm("usd_compatible_representations") and node.parm("usd_compatible_representations").eval() else None
```
BigRoy, line 818: Why is this `usd` specific?

Also, by adding a default filter here it basically avoids the generic loader being useful **everywhere**. It should not filter by default, unless e.g. something on the node specifies a filter?

MustafaJafar: This toggle only found on `Generic Loader` lop node. 
and, I had to provide such a filter because the Asset/Shot load node will then have representations it can't load like `exr` or `vdb`.  
image

BigRoy: Instead of doing that - add a built-in feature for any generic loader to specify a (unique) filter  by having a hidden parm on the generic loader that takes a space separated list of representation names to filter to- when empty no filtering is applied. By default, it's empty.

Then your node can just set that parm to e.g. `abc usd` on the node itself

moonyuet:
I have tested with the original loaded asset and tries to switch across the library projects. it runs smoothly
image
But I do wonder how the generic loader works for this case(or is it still work in progress?) as no related geometries/caches show up in the scene.
image

MustafaJafar:
> But I do wonder how the generic loader works for this case(or is it still work in progress?) as no related geometries/caches show up in the scene.

The generic loader is used to obtain the file path of the published product, allowing management through the AYON scene inventory tool.
Also, in this PR https://github.com/ynput/ayon-houdini/pull/121, Generic loader is used with file path loader 
> The filepath loader is now using the generic loader instead of a null node.
> image

MustafaJafar:
I've added a representation filter parm to generic loaders. 
it can be toggled on/off by users.
Animation_51

moonyuet:
I have tested as stated in your documentation
Looks good now. 
image

moonyuet (approved):

client/ayon_houdini/api/hda_utils.py
MustafaJafar: https://github.com/ynput/ayon-houdini/pull/199/commits/714fe325f5aad3189cf82a42c1f50e4084dc8e90

MustafaJafar:
@BigRoy

Regarding the last piece of the puzzle, the `thumbnail`, I'm not entirely sure how the system works. However, it seems that every node using the thumbnail logic—whether it's a generic loader or LOP loaders—utilizes the same thumbnail file as long as they load the same asset. It appears to be cached, so I don't think we need to optimize LOP loaders to point the thumbnail to the inner thumbnail of the inner generic loader node.

image

philippe-ynput:
Hi @MustafaJafar :-)
What's holding up this PR ? It looks ready, right ?

MustafaJafar:
> Hi @MustafaJafar :-) What's holding up this PR ? It looks ready, right ?

Hello, I just wanted to get @BigRoy's opinion.

philippe-ynput:
Thanks @MustafaJafar ! I'll let @BigRoy chime in...

BigRoy:
Correct, on my plate! Wanted to do today, but another thing came in between. So, tomorrow!

philippe-ynput:
Once you're done @BigRoy, let's not forget to merge the [docs part](https://github.com/ynput/ayon-documentation/pull/321).

PR #200, opened 2024-12-18
130 ay 6965 default output paths and support custom staging directories.

### LINKED ISSUE #21: Use custom staging dir function for Houdini

Use custom staging dir function for Houdini.

Not sure, will update it later.
I think it's related to ayon core settings:
```
ayon+settings://core/tools/publish/custom_staging_dir_profiles
```

_No response_

### Additional context:

I found this on the internal request description:
```
Based on  transient dir function pipeline.publish.lib.get_transient_dir_info should be
 used everywhere local render folders are used.
```

### LINKED ISSUE #130: AY-6965_ROP default output path, update '/pyblish/' to something less legacy

MustafaJafar:
## Changelog Description
This PR adds a main setting for configuring output directory for created ROP nodes via AYON creator plugins.
- [x] Add settings toggle to enable/disable path control completely (to allow a studio to completely manage it themselves) 
- [x] If enabled, 
    - we use the default path as follows ```{staging_dir}/hardcoded_file_name.{ext}```
    - intermediate render directories get a hardcoded folder name as follows```{staging_dir}/intermediate_file_type/hardcoded_file_name```
- [x] However, **if** matching custom staging dir profile is specified in `ayon-core` settings then we use that custom staging dir directory instead. (This also _only_ applies if "directory management" is enabled" as per the first point).
- [x] Expand Houdini vars when enabled, vars like `$HIP` will be expanded while not expanding any expressions.

image

resolve #130 and resolve #21

## Demo
- staging templates
image
-  custom staging directories
image

image
1. follows `houdini_cache` template
2. follows `houdini_Render` template
3. follows the default value in Houdini settings as it doesn't match any profile.

## Testing notes:
1. Create new instances in Houdini, you should find the new output paths.
4. Publishing new and old instances should work without issues.


moonyuet:
I tested with publishing the model product. the staging directory set correctly
image
But later I find out that two of the validators are broken, guess we need to implement the fixes on these two.
They are `Validate Primitive to Detail(abc)` and `Validate Single Frame`
The attachments are the json report to show the error of the validator(s).
[publish-report-241223-16-42.json](https://github.com/user-attachments/files/18227424/publish-report-241223-16-42.json)
This error of `Validate Single Frame` showed after the fix in `Validate Primitive to Detail(abc)`
[publish-report-241223-16-44.json](https://github.com/user-attachments/files/18227426/publish-report-241223-16-44.json)

moonyuet (approved):
See the comment above

BigRoy:
Nice work!

1. I'm not entirely convinced that ONLY allowing to customize the output directory is sufficient - but maybe it is! ❤️ Would need input from the community to fully grasp whether this is the best approach. Like it may make sense to be more concrete about the filename as well?
2. Whenever we discuss this 'feature' I always keep thinking whether we are doing it the right Houdini-way. I'm wondering if we should have a checkbox in settings that instead disables setting the path completely so that it's possible for a studio to _actually_ define the output path completely by setting their own node attribute defaults directly in Houdini. Again, possibly something we should check with the Houdini-community using AYON. Does it actually make sense that we set it through code instead of relying on the node's defaults (which can be set by a studio, or even an artist specifically) to be adopted instead. 
3. Having a single configurable entry does keep it VERY simple to manage - I like that. However, it does disallow completely to go more granular (which wasn't possible anyway, so could be something for later instead).

Also wanted to mention that there _is_ quite some overlap between this feature and the **Custom Staging Dir** feature which would define where published work renders should go to. As such, even though I like this implementation it may very well happen down the line that this should get superseded by using the custom staging dir setings (which I think are in `core` as a feature?)

BigRoy (commented):
My main concerns are mentioned [here](https://github.com/ynput/ayon-houdini/pull/200#issuecomment-2574381567) - other than that I think the implementation is very clean and nice.

However, I do think the:
```
`chs('AYON_productName')`
```
part should be part of the setting, not hardcoded in the setting of the path.

As such, the default would be this instead:

```
staging_dir = "$HIP/ayon/`chs('AYON_productName')`"
```

I also wonder whether making the new default `render/` instead of `ayon/` would make way more sense. It'd at least align more with other host integrations like Maya.

@antirotor thoughts?

MustafaJafar:
> However, I do think the:
> 
> ```
> `chs('AYON_productName')`
> ```
> 
> part should be part of the setting, not hardcoded in the setting of the path.

This won't work with HDAs, as its file name is not saved in a string parameter. 
Also for reference, path in settings is not actually a staging directory it's more of a root output directory and every product can have multiple files like render has render and intermediate render files.
Here's a list of the current paths in this PR (the root is the path in settings: `$HIP/ayon`)
```
- Camera, Model, abc pointcache: "{root}/`chs('AYON_productName')`/$OS.abc"
- Arnold 🍑: "{root}/`chs('AYON_productName')`/$OS.$F4{ext}"
- Arnold: 
  - render: "{root}/`chs('AYON_productName')`/$OS.$F4.{ext}"
  - 🍑: "{root}/`chs('AYON_productName')`/ass/$OS.$F4.ass"
- Bgeo pointcache: "{root}/`chs('AYON_productName')`/$OS.$F4.{ext}"
- Cop: "{root}/`chs('AYON_productName')`/$OS.$F4{ext}"
- HDAs: "{root}/HDAs/{node_name}.hda"
- Karma: 
  - render: "{root}/`chs('AYON_productName')`/$OS.$F4.{ext}"
  - checkpoint: "{root}/`chs('AYON_productName')`/checkpoint/$OS.$F4.checkpoint"
  - USD output: "{root}/`chs('AYON_productName')`/usd/$OS_$RENDERID"
- Mantra: 
  - render: "{root}/`chs('AYON_productName')`/$OS.$F4.{ext}"
  - ifd: "{root}/`chs('AYON_productName')`/ifd/$OS.$F4.ifd"
- RS proxy: "{root}/`chs('AYON_productName')`/$OS.$F4.rs"
- Redshift:
  - render: "{root}/`chs('AYON_productName')`/$OS.$AOV.$F4.{ext}"
  - proxy: "{root}/`chs('AYON_productName')`/rs/$OS.$F4.rs
- review: "{root}/`chs('AYON_productName')`/$OS.$F4.{ext}"
- staticmesh: "{root}/`chs('AYON_productName')`/$OS.fbx"
- USD: "{root}/`chs('AYON_productName')`/$OS.usd"
- USD render: "{root}/`chs('AYON_productName')`/usd/$HIPNAME/$OS
- VDB: "{root}/`chs('AYON_productName')`/$OS.$F4.vdb"
- VRay:
  - with AOV: "{root}/`chs('AYON_productName')`/$OS.$AOV.$F4.{ext}"
  - without AOV: "{root}/`chs('AYON_productName')`/$OS.$F4.{ext}"
```

MustafaJafar:
Regarding point 1 and 3 about being more granular about file names and not using hardcoded subdirectories.
This was implemented in the far first few commits in the PR before reverting these changes.
IMO, the settings were too many making them hard to use.
https://github.com/ynput/ayon-core/pull/9

BigRoy:
> This won't work with HDAs, as its file name is not saved in a string parameter. Also for reference, path in settings is not actually a staging directory it's more of a root output directory and every product can have multiple files like render has render and intermediate render files. Here's a list of the current paths in this PR (the root is the path in settings: `$HIP/ayon`)

I see, thanks for clarifying that. Still - it feels odd to still have that part of the path be hardcoded for all the others. Not sure what the best middle ground is - but in a way, it's still not really customizable.

Also, similar to "custom staging dir" - a studio may want their renders in a different folder than their geo caches. Which also hints at needing settings PER creator instead of a single global one. Yet I do agree - it'd be a lot of settings to maintain. It could be profile based - where the HDA creator just happens to have a different default profile.

Tagging @dee-ynput @antirotor for their thoughts.

---

> > I also wonder whether making the new default `render/` instead of `ayon/` would make way more sense. It'd at least align more with other host integrations like Maya.

> I wonder if Houdini artists might find it confusing because it doesn't align with Houdini defaults where geo files will be saved in `render/` instead of `/geo`.

Valid point - potentially even more reason to 'remain at the defaults' as per my point 2) above.

BigRoy (changes requested):
After the meeting we had with @antirotor and @MustafaJafar this became the changes:
- [x] Add settings toggle to enable/disable path control completely (to allow a studio to completely manage it themselves) 
- [x] If enabled, we use the default path as you have it set up ✅ 
- [x] However, **if** matching custom staging dir profile is specified in `ayon-core` settings then we use that custom staging dir directory instead. (This also _only_ applies if "directory management" is enabled" as per my first point).

client/ayon_houdini/plugins/create/create_mantra_rop.py
```diff
@@ -32,29 +34,24 @@ def create(self, product_name, instance_data, pre_create_data):
...
-
         parms = {
             # Render Frame Range
             "trange": 1,
-            # Mantra ROP Setting
-            "vm_picture": filepath,
         }
+        if self.enable_staging_dir:
+            # keep dynamic link to product name in file path.
+            self.staging_dir = get_custom_staging_dir("render", product_name) or self.staging_dir
```
MustafaJafar, line 43: I'm here using `"render"` as the product name because it's the real product name of the published product (exr).

server/settings/general.py
```diff
@@ -36,6 +61,10 @@ class GeneralSettingsModel(BaseSettingsModel):
 
 DEFAULT_GENERAL_SETTINGS = {
     "add_self_publish_button": False,
+    "rop_output": {
+        "enabled": True,
+        "default_output_dir": "$HIP/ayon/"
```
MustafaJafar, line 66: (cosmetics) 
we don't need the trailing `/` although keeping it give the feeling it's a directory. 
so, I think we should remove it when reading/getting the staging directory.
```suggestion
        "default_output_dir": "$HIP/ayon"
```

MustafaJafar (commented):
Here are some note.

client/ayon_houdini/plugins/create/create_arnold_ass.py
```diff
@@ -41,17 +43,20 @@ def create(self, product_name, instance_data, pre_create_data):
...
             # Render frame range
             "trange": 1,
             # Arnold ROP settings
-            "ar_ass_file": filepath,
             "ar_ass_export_enable": 1
         }
+        
+        if self.enable_staging_dir:            
+            # keep dynamic link to product name in file path.
+            self.staging_dir = get_custom_staging_dir(self.product_type, product_name) or self.staging_dir
```
BigRoy, line 55: When calling this - it'd be good to pass the project settings from the create context to the `get_custom_staging_dir` which can then be passed to `get_staging_info`

I'm also not sure about using the `get_current_context` in that function. We should be passing the context of the `create_context` to be explicit, and preferably even that of the instance (if targeting another folder path and task than current context, which of the ones should it be using? I assume the target one?)

I'd move that to the `HoudiniCreator` base class and just create a `get_staging_dir(self, instance: CreatedInstance)` from which I assume we should be able to get all data to compute the resulting staging dir - moving all the heavy lifting of that method to the base class instead of having to duplicate it to each inherited class.

Also, I'd avoid overwriting `self.staging_dir` with this result - AYON plugin systems tend to re-use base classes and instances in different areas (not sure if creators do that too) but it's prone to somehow 'spilling' into the next call of a creator if it does. In this case there's also no need to store it on `self` since you're just using the result anyway. ;) So just make it the result of that `self.get_staging_dir()` call.

client/ayon_houdini/api/plugin.py
```diff
@@ -295,6 +297,9 @@ def apply_settings(self, project_settings):
         houdini_general_settings = project_settings["houdini"]["general"]
         self.add_publish_button = houdini_general_settings.get(
             "add_self_publish_button", False)
+        
+        self.enable_staging_dir = houdini_general_settings["rop_output"]["enabled"]
```
BigRoy, line 301: I'm not entirely convinced by the `enable_staging_dir` attribute name. I wonder if it should be more like `enable_staging_path_management` or whatever because it involves more than just a directory, it completely disables the setting of a path on the created nodes. Or `self.set_render_path_on_create` (just throwing this in to give another idea).

client/ayon_houdini/plugins/create/create_vray_rop.py
```diff
@@ -104,20 +98,14 @@ def create(self, product_name, instance_data, pre_create_data):
...
             })
-
-        else:
-            filepath = "{renders_dir}{product_name}/{product_name}.{fmt}".format(
-                renders_dir=hou.text.expandString("$HIP/pyblish/renders/"),
-                product_name=product_name,
-                fmt="$F4.{ext}".format(ext=ext)
-            )
-            parms.update({
-                "use_render_channels": 0,
```
BigRoy, line 118: Did we completely lose setting this parm?

MustafaJafar: Prior to my commit https://github.com/ynput/ayon-houdini/pull/200/commits/d93fe8b89e68a260b635c4abbb5bf7d8ea7ad146, we had an if-else that sets that parm.
and in my commit, I moved the part of the else clause earlier in the code.

client/ayon_houdini/plugins/create/create_arnold_ass.py
MustafaJafar: Thank you for pointing that out, I didn't know that [Creator.get_staging_dir](https://github.com/ynput/ayon-core/blob/cfa5815fdab70ec77b8d2906ea17bd42d8506b6d/client/ayon_core/pipeline/create/creator_plugins.py#L836-L894) exists. 

I only knew about [get_staging_dir_info](https://github.com/ynput/ayon-core/blob/cfa5815fdab70ec77b8d2906ea17bd42d8506b6d/client/ayon_core/pipeline/staging_dir.py#L118-L238), [Extractor.staging_dir](https://github.com/ynput/ayon-core/blob/cfa5815fdab70ec77b8d2906ea17bd42d8506b6d/client/ayon_core/pipeline/publish/publish_plugins.py#L390-L397) and [get_instance_staging_dir](https://github.com/ynput/ayon-core/blob/cfa5815fdab70ec77b8d2906ea17bd42d8506b6d/client/ayon_core/pipeline/publish/lib.py#L673-L723)


> (if targeting another folder path and task than current context, which of the ones should it be using?

The creator tab (at least what I'm seeing in the UI) doesn't allow creating an instance in a different context.
and users can change that in the publish tab later.

Maybe for that case, we may just create a validator that checks the staging dir path🤔?

client/ayon_houdini/api/plugin.py
MustafaJafar: https://github.com/ynput/ayon-houdini/pull/200/commits/35b1c04ad9d3e2304f2947c3261a5a09c1488db4

client/ayon_houdini/plugins/create/create_arnold_ass.py
MustafaJafar: Any ideas about how to use the  [get_staging_dir_info](https://github.com/ynput/ayon-core/blob/cfa5815fdab70ec77b8d2906ea17bd42d8506b6d/client/ayon_core/pipeline/staging_dir.py#L118-L238) with HDA creator which uses `create_instance_node` method that doesn't have access to the `instance` object.

MustafaJafar: I tried using  [get_staging_dir_info](https://github.com/ynput/ayon-core/blob/cfa5815fdab70ec77b8d2906ea17bd42d8506b6d/client/ayon_core/pipeline/staging_dir.py#L118-L238) 
By overriding it in `HoudiniCreator`, I wanted to allow overriding the instance type so that I can use the real product type for render product creators e.g. `"mantra_rop"` creator should actually get the staging directory of `"render"` product type. 
The following snippet doesn't work because of `TypeError: cannot pickle 'WeakMethod' object` error. 
```python
    def get_staging_dir(self, instance, product_type_override=None):
        if product_type_override:
            instance = copy.deepcopy(instance)
            instance["productType"] = product_type_override
        return super().get_staging_dir(instance) or self.staging_dir
```

MustafaJafar: I hope this to be a good middle ground, please let me know what do you think.
https://github.com/ynput/ayon-houdini/pull/200/commits/6e156043aaabdb754ba3bb06e8218039a7007d80

BigRoy: Yes  nice. Now should we move the `expandString` call into the method too? So that it's just one place where we're expanding it?

MustafaJafar: The current status of this PR doesn't respect the changes in the context (`folder_path` and `task`) which leads to exporting/saving the files into the same context.. which is mostly the same directory of my workfile.

Very important, the creator tab in the publisher UI doesn't allow changing the project, so, we are always using the current project which means we won't need to get and pass any project settings.

---

we should also take into considerations that when respecting the changes in the context, our files will be exported in another location other that the current working directory.  

---

anyways, I tried finding ways to consider the changes in the context. 

I thought about using `self.create_context`. but I found out it is just a `ayon_core.pipeline.create.CreateContext` object which doesn't include info about the selected folder path and task in the publisher UI. 

Also, found out that the selection is saved in the `instance` object, so we would need to pass the `instance` like the `get_staging_dir` in super class.

The `get_staging_dir` in super class covers more use cases but sometimes, we don't want to use it in Houdini as
- it doesn't work with HDA creator  as `create_instance_node` is called before the `instance` object creation.
- it doesn't allow overriding the product type (which is needed with render rops). so admins will need to use the product type for render rops like `mantra_rop`, `karma_rop` and etc.

I'm not sure what would be the ideal solution.. I can think of these ideas: 
1. use `get_staging_dir` from super class and I've no idea what to do with HDAs and render creators.
2. copy the implementation of `get_staging_dir` from super class and tweak it to be Houdini friendly where I can implement some overrides. I can pass `product_name`, `product_type` and `folder_path` but no idea about getting the version as `get_staging_dir` includes this call `self.get_next_versions_for_instances([instance])`.. the version is used as template data.
3. continue with the current implementation and add a note/todo that mentions the limitations about respecting changes in the context selection in the publisher UI leading to always following the current context. 


Regarding the second idea, Maybe something like this will work: 
```python
    def get_staging_dir(self, product_type, product_name, folder_path, version):

        fake_instance = {
            "productName": product_name,
            "productType": product_type,
            "folderPath": folder_path,
            "version": version
        }
        return super().get_staging_dir(fake_instance)
```

moonyuet: > The current status of this PR doesn't respect the changes in the context (`folder_path` and `task`) which leads to exporting/saving the files into the same context.. which is mostly the same directory of my workfile.
> 
> Very important, the creator tab in the publisher UI doesn't allow changing the project, so, we are always using the current project which means we won't need to get and pass any project settings.
> 
> we should also take into considerations that when respecting the changes in the context, our files will be exported in another location other that the current working directory.
> 
> anyways, I tried finding ways to consider the changes in the context.
> 
> I thought about using `self.create_context`. but I found out it is just a `ayon_core.pipeline.create.CreateContext` object which doesn't include info about the selected folder path and task in the publisher UI.
> 
> Also, found out that the selection is saved in the `instance` object, so we would need to pass the `instance` like the `get_staging_dir` in super class.
> 
> The `get_staging_dir` in super class covers more use cases but sometimes, we don't want to use it in Houdini as
> 
> * it doesn't work with HDA creator  as `create_instance_node` is called before the `instance` object creation.
> * it doesn't allow overriding the product type (which is needed with render rops). so admins will need to use the product type for render rops like `mantra_rop`, `karma_rop` and etc.
> 
> I'm not sure what would be the ideal solution.. I can think of these ideas:
> 
> 1. use `get_staging_dir` from super class and I've no idea what to do with HDAs and render creators.
> 2. copy the implementation of `get_staging_dir` from super class and tweak it to be Houdini friendly where I can implement some overrides. I can pass `product_name`, `product_type` and `folder_path` but no idea about getting the version as `get_staging_dir` includes this call `self.get_next_versions_for_instances([instance])`.. the version is used as template data.
> 3. continue with the current implementation and add a note/todo that mentions the limitations about respecting changes in the context selection in the publisher UI leading to always following the current context.
> 
> Regarding the second idea, Maybe something like this will work:
> 
> ```python
>     def get_staging_dir(self, product_type, product_name, folder_path, version):
> 
>         fake_instance = {
>             "productName": product_name,
>             "productType": product_type,
>             "folderPath": folder_path,
>             "version": version
>         }
>         return super().get_staging_dir(fake_instance)
> ```
Have you tried to get staging directory without super class?(I know it's dumb to ask but it should be possible to do so?)
Speaking of this, you may consider taking reference from what Robin and Jeza did for the custom staging directory in Nuke. https://github.com/ynput/ayon-nuke/pull/19.

MustafaJafar: I've traced setting stage directory in nuke, so basically: 
1. It doesn't include any toggles to enable/disable setting staging directories. The way to enable it in Nuke by adding a profile to `custom_staging_dir_profiles` setting.
2. It creates an instance, applies/gets staging director and create instance node. as shown [here](https://github.com/ynput/ayon-nuke/blob/aee0fbeaf78f738b961bcb80bae1193a936888bb/client/ayon_nuke/api/plugin.py#L471-L484)

So, creating the instance precedes creating the node. which allows using methods like `apply_staging_dir` from the super `creator` class.

In Houdini we have a different situation where we: 
1. want to define default values in settings. 
2. add a toggle to disable setting staging directory either default value or the value in `custom_staging_dir_profiles` setting.
3. create instance node precedes creating AYON instance as shown [here](https://github.com/ynput/ayon-houdini/blob/26f8cfb9d7f1b0a3a651d89161a825e8617a7d91/client/ayon_houdini/api/plugin.py#L127-L144). I'm not sure if we can swap them because we are using the node path as the `instance_id`. 
4. want (in some cases) to override the product type when getting the staging directory. 

> Have you tried to get staging directory without super class?

This is the current implementation in this PR but it lacks respecting the AYON context (folder path and task) because these info are saved on the instance itself and at the point of time when the instance node is created, the `instance` object doesn't exist. **but they do exist on the `instance_data`.. let me try passing that to `create_instance_node` that!**

server/settings/general.py
MustafaJafar: https://github.com/ynput/ayon-houdini/pull/200/commits/1543271a0d3d6962977945d51df711160338e5d2

MustafaJafar:
This should be working now. 
Also, I've explained in code why some functions and methods weren't used to obtain the staging dir. 
Anyways, I gave it a test run and everything works as before.

client/ayon_houdini/api/lib.py
```diff
@@ -23,6 +23,7 @@
     get_current_context,
     get_current_host_name,
     get_staging_dir_info,
+    get_current_project_name,
```
moonyuet, line 26: Is this used anywhere in this script?

client/ayon_houdini/api/lib.py
```diff
@@ -21,6 +22,8 @@
     registered_host,
     get_current_context,
     get_current_host_name,
+    get_staging_dir_info,
+    get_current_project_name,
```
moonyuet, line 26: Is this used anywhere in this script?

MustafaJafar: good catch! 
https://github.com/ynput/ayon-houdini/pull/200/commits/08b7958e725572db2f53e7abf911941e55cb39a3

client/ayon_houdini/api/plugin.py
```diff
@@ -312,6 +320,44 @@ def apply_settings(self, project_settings):
         for key, value in settings.items():
             setattr(self, key, value)
 
+    def get_staging_dir(self, product_type, product_name, instance_data=None):
```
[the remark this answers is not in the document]
MustafaJafar: Currently, all creator plugins inherits `HoudiniCreator` which defines an override for `get_staging_dir`.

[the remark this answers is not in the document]
BigRoy: If we are to override a method that exists on the `Creator` plug-in, then we should match the function signature and have the same arguments. Otherwise, any code in core calling `get_staging_dir` on the creator will fail because it doesn't know about the houdini-specific arguments.

We should not change the signature of methods in inherited children classes.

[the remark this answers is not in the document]
MustafaJafar: does renaming the method suffice? e,g, `get_custom_staging_dir`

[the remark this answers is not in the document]
BigRoy: Using an alternate method then replacing the original one does work - yes. However, be aware that there _may_ still be calls being done by other addons calling the original implementation, not knowing about the specialized one. (I'm not saying they do now, but they might some day if they don't yet now)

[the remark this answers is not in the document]
MustafaJafar: Then, I think it's fair enough to change the method name and leave the original thing as it's.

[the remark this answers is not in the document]
MustafaJafar: Thank you https://github.com/ynput/ayon-houdini/pull/200/commits/a4ffbaf91ceef605992380f5db78d514d198f9e7

BigRoy (changes requested):
Overall this works great - however, I still think we should make the expanding of `$HIP` optional (even if another validator is currently in the way of making that super useful.)

What I'd like to do is:
1. Move the `hou.text.expandString(staging_dir)` call into `HoudiniCreator.get_custom_staging_dir` instead of per creator
2. Add an additional setting in settings that enables/disables that `if expand_vars: ...` if managing the paths is enabled.

I also think we should update `ROP Output Directory` label to `Set ROP Output Directory on Create` or something so it's explicit when it triggers from the label, instead of needing to look at the tooltips. What do you think?

Then, after this PR is merged - we should maybe create an issue to look into ways how we should potentially _update_ the render paths _after_ e.g. context switching if needed so that we avoid re-using a scene using the old render paths accidentally. Whether that's a validator, or the creator handling that out of the box is something we can discuss and decide then.

Other than that, great work - very clean!

MustafaJafar:
@BigRoy 
I did couple of things:
> * Move the `hou.text.expandString(staging_dir)` call into `HoudiniCreator.get_custom_staging_dir` instead of per creator
> * Add an additional setting in settings that enables/disables that `if expand_vars: ...` if managing the paths is enabled.
> * the default would be this instead:
> 
> ```
> staging_dir = "$HIP/ayon/`chs('AYON_productName')`"
> ```

moonyuet:
Did some test with farm rendering, looks good.
image

moonyuet (approved):

client/ayon_houdini/api/plugin.py
```diff
@@ -312,6 +326,48 @@ def apply_settings(self, project_settings):
         for key, value in settings.items():
             setattr(self, key, value)
 
+    def get_custom_staging_dir(self, product_type, product_name, instance_data=None):
```
iLLiCiTiT, line 329: You know there is `get_staging_dir` on create plugin, which replaces almost everything you've implemented here?

client/ayon_houdini/api/lib.py
```diff
@@ -1623,4 +1625,78 @@ def format_as_collections(files: list[str], pattern: str = "{head}{padding}{tail
...
+        )
+
+    staging_dir_path = None
+    if staging_dir_info:
+        staging_dir_path = staging_dir_info.directory
+
+    return staging_dir_path
+
+
+def expand_houdini_string(text: str, pattern=r"\$[a-zA-Z0-9_]+") -> str:
```
BigRoy, line 1686: Does the `hou.text.expandString` `expand_tilde` argument suffice for our needs? or do we still need to wrap this?

client/ayon_houdini/api/plugin.py
BigRoy: Does the note below help  answer that? https://github.com/ynput/ayon-houdini/pull/200/files#diff-02368e53ec1a73faaed14dfeee7bba367ff684790b1dd08cfe57a781422eeca5R338-R34

server/settings/general.py
```diff
@@ -36,6 +70,11 @@ class GeneralSettingsModel(BaseSettingsModel):
 
 DEFAULT_GENERAL_SETTINGS = {
     "add_self_publish_button": False,
+    "set_rop_output": {
+        "enabled": True,
+        "expand_vars": False,
+        "default_output_dir": "$HIP/ayon/`chs('AYON_productName')`"
```
BigRoy, line 76: Let's use `"` in the houdini expression by default. I rarely see anyone use `'` there.
```suggestion
        "default_output_dir": "$HIP/ayon/`chs(\"AYON_productName\")`"
```

client/ayon_houdini/api/plugin.py
iLLiCiTiT: NOTE: `instance._create_context` and `instance._creator.create_context` can be missing, but it is method on create plugin, so you have access to create plugin -> you have access to create context.

client/ayon_houdini/api/lib.py
MustafaJafar: The reason for wrapping this is to expand vars selectively.  
Unfortunately, `hou.text.expandString` doesn't provide an argument for that purpose. 
e.g. I can use `$HIP` and `chs("AYON_productName")` in my staging path. but, I want keep the dynamic reference `chs("AYON_productName")`.
After all, the toggle is called `Expand Houdini Vars` not `Expand String`.

Does it sound like magic? 
For reference, I've added this info in the tooltip on the server. 
image

MustafaJafar: To make it less magical, we can have something like this which exposes the selection to admins? 

image

BigRoy (dismissed):
As discussed this morning @MustafaJafar we'll be doing some remaining tweaks to this:

1. Do not hardcode the render rops (e.g. `karma_rop`) to `render` product type for custom staging dir. (This will likely get superseded by https://github.com/ynput/ayon-houdini/pull/214)
2. Reduce the amount of customization on the staging dir logic. Preferably the `HoudiniCreator.get_staging_dir` suffices and any customizations needed are solely based on calling `super().get_staging_dir()` with only deviating on that result where we need.
3. There are issues with the usage of `hou.text.expandString()` because you're not calling them in the context of the created node and its parameters, as such any `$OS` or relative parameter references like the 'chs("productName")'-type of usage will be evaluated correctly, when calling we should make sure to call it from the `pwd` of the instance node.
4. The only reason we're deviating this much from `get_staging_dir` was because at the time we want to call `get_staging_dir` the `CreatedInstance` does not exist yet. As such, I think we should refactor the code so that it does, what I'd propose is separating the logic of the "creation of the node" and the "applying of the parameters" of the node, so that the applying of the parameters happens after the generation of the `CreatedInstance`, like:
```python
node = self.create_instance_node(...)
instance_data["instance_node"] = node.path()
instance = CreatedInstance(...)
staging_dir = self.get_staging_dir(instance)
self.set_node_staging_dir(node: hou.Node, staging_dir: str, instance: CreatedInstance)
```
Or something along those lines.
This would also allow to turn that directly into this on the base class.
```python
node = self.create_instance_node(...)
instance_data["instance_node"] = node.path()
instance = CreatedInstance(...)
if if self.enable_staging_path_management:
    staging_dir = self.get_staging_dir(instance)
    self.set_node_staging_dir(node: hou.Node, staging_dir: str, instance: CreatedInstance)
```
So that the if statement does not need to be copied to each inherited class but only `set_node_staging_dir` would need to be implemented.

MustafaJafar:
I've applied most of the mentioned things in https://github.com/ynput/ayon-houdini/pull/200#pullrequestreview-2582479596 but I've no ideas how to make it work with HDA product without having duplicate files because `create_instance_node` will create one and `set_node_staging_dir` (that apply the staging path) will create another file.

client/ayon_houdini/api/lib.py
```diff
@@ -1623,4 +1623,37 @@ def format_as_collections(files: list[str], pattern: str = "{head}{padding}{tail
...
+
+    def expand_match(match):
+        matched_string  = match.group(0)
+        return hou.text.expandString(matched_string)
+
+    return re.sub(pattern, expand_match, text)
+
+
+@contextmanager
+def set_pwd(node):
```
BigRoy, line 1649: I wonder if `hou.ScriptEvalContext` behaves the same for our use case: https://www.sidefx.com/docs/houdini/hom/hou/ScriptEvalContext.html

client/ayon_houdini/api/plugin.py
```diff
@@ -142,6 +152,17 @@ def create(self, product_name, instance_data, pre_create_data):
...
                 instance_data,
                 self)
+            
+            if self.enable_staging_path_management:
+                staging_dir = self.get_staging_dir(instance)
+
+                with set_pwd(instance_node):
+                    if self.expand_staging_dir:
+                        # Expand vars only without expanding expressions to keep dynamic link to ROP parameters.
+                        staging_dir = expand_houdini_string(staging_dir)
```
BigRoy, line 162: We could swap this, right?
```suggestion
                if self.expand_staging_dir:
                    with set_pwd(instance_node):
                        # Expand vars only without expanding expressions to keep dynamic link to ROP parameters.
                        staging_dir = expand_houdini_string(staging_dir)
```

client/ayon_houdini/api/lib.py
MustafaJafar: wow, huge thanks.
https://github.com/ynput/ayon-houdini/pull/200/commits/708268018f42b8e0fe1dec98a201e6608dfd806e

client/ayon_houdini/api/plugin.py
MustafaJafar: good catch. I tweaked it a bit.
https://github.com/ynput/ayon-houdini/pull/200/commits/a3123a75a1d22872642d5f28a2cafc6f230e1afa

server/settings/general.py
MustafaJafar: https://github.com/ynput/ayon-houdini/pull/200/commits/47f09fdd093c25372baf39ec6855244060a566d1

MustafaJafar:
> 4\. The only reason we're deviating this much from `get_staging_dir` was because at the time we want to call `get_staging_dir` the `CreatedInstance` does not exist yet. As such, I think we should refactor the code so that it does, what I'd propose is separating the logic of the "creation of the node" and the "applying of the parameters" of the node, so that the applying of the parameters happens after the generation of the `CreatedInstance`, like:

This logic works with all product types we have in Houdini except `HDA` as `set_node_staging_dir` in your example will actually create a duplicate for the HDA to save it in the correct staging path. (I'm not sure how much frustration it will lead to)

The problem with HDA product is a circular dependency 
1. Create Houdini node (HDA need the staging directory on creation)
2. Store its path in the scene as `ID`
3. Create the publish instance with the stored `ID`.
4. Use the instance to get the staging directory.

I couldn't create the instance and get the staging directory first because Key "instance_id" is immutable and does not allow changes.

MustafaJafar:
@BigRoy 

I've found out a native Houdini way to fix it by creating the HDA _embedded_ in the scene and then save it later inside  `set_node_staging_dir` function.

client/ayon_houdini/api/plugin.py
```diff
@@ -312,6 +336,42 @@ def apply_settings(self, project_settings):
...
+
+        Args:
+            instance (CreatedInstance): Instance for which should be staging
+                dir gathered.
+
+        Returns:
+            str: Staging dir path
+        """
+
+        staging_dir_info = super(HoudiniCreator, self).get_staging_dir(instance)
```
iLLiCiTiT, line 355:
```suggestion
        staging_dir_info = super().get_staging_dir(instance)
```

client/ayon_houdini/api/plugin.py
```diff
@@ -312,6 +336,42 @@ def apply_settings(self, project_settings):
...
+
+        staging_dir_path =  self.default_staging_dir
```
iLLiCiTiT, line 357:
```suggestion
        staging_dir_path = self.default_staging_dir        
```

client/ayon_houdini/api/plugin.py
```diff
@@ -312,6 +336,42 @@ def apply_settings(self, project_settings):
...
+    def set_node_staging_dir(self, node: hou.Node, staging_dir: str, instance: CreatedInstance, pre_create_data: dict):
+        """Set Node Staging Dir
+
+        Args:
+            node (hou.Node): Houdini node to set its output directory.
+            staging_dir (str): Staging output directory.
+            instance (CreatedInstance): Instance object associated with the given node.
+            pre_create_data(dict): Data based on pre creation attributes.
+        """
+
```
iLLiCiTiT, line 372:
```suggestion

        """
```

iLLiCiTiT:
I can review code and some cores of logic, but can't help with houdini logic.

My point of view:
1. Shouldn't we always use staging dir logic? I see it is defaulting to `$HIP/ayon` so probably not, but that's why there is question mark.
2. What `enable_staging_path_management` acually do? We need staging all the time, or?
3. I think the settings to expand variables makes it uber confusing. I guess you can't use `$HIP/ayon` anymore? Does it make sense to have the option to change it?

If 1. makes sense, then `expand_staging_dir` is kinda requirement for `enable_staging_path_management` automatically, or?

iLLiCiTiT:
BTW it looks like Houdini addon does not have strict rules for ruff.

client/ayon_houdini/plugins/create/create_vray_rop.py
```diff
@@ -129,6 +97,15 @@ def create(self, product_name, instance_data, pre_create_data):
         to_lock = ["productType", "id"]
         self.lock_parameters(instance_node, to_lock)
 
+    def set_node_staging_dir(self, node, staging_dir, instance, pre_create_data):
+        node.parm("render_export_filepath").set(f"{staging_dir}/vrscene/$OS.$F4.vrscene")
+        
+        if pre_create_data.get("render_element_enabled", True):
+            node.parm("SettingsOutput_img_file_path").set(f"{staging_dir}/$OS.$AOV.$F4.{pre_create_data['image_format']}") 
+        else:
+             node.parm("SettingsOutput_img_file_path").set(f"{staging_dir}/$OS.$F4.{pre_create_data['image_format']}")
```
iLLiCiTiT, line 106:
```suggestion
        image_format = pre_create_data["image_format"]
        aov_value = ""
        if pre_create_data.get("render_element_enabled", True):
            aov_value = ".$AOV"

        node.parm("SettingsOutput_img_file_path").set(
            f"{staging_dir}/$OS{aov_value}.$F4.{image_format}"
        )
```

client/ayon_houdini/plugins/create/create_vray_rop.py
```diff
@@ -53,19 +53,14 @@ def create(self, product_name, instance_data, pre_create_data):
 
         parms = {
             "trange": 1,
-            "SettingsEXR_bits_per_channel": "16"   # half precision
+            "SettingsEXR_bits_per_channel": "16",   # half precision
+            "use_render_channels": 0,
         }
-
+
```
iLLiCiTiT, line 59:
```suggestion

```

MustafaJafar: https://github.com/ynput/ayon-houdini/pull/200/commits/385a239730cc2926528d813d77f13875baac954d

client/ayon_houdini/plugins/create/create_arnold_rop.py
```diff
@@ -35,37 +35,28 @@ def create(self, product_name, instance_data, pre_create_data):
...
         parms = {
             # Render frame range
             "trange": 1,
-
-            # Arnold ROP settings
-            "ar_picture": filepath,
+            # Arnold ROP settings            
             "ar_exr_half_precision": 1           # half precision
-        }
+        }
```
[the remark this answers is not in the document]
MustafaJafar: https://github.com/ynput/ayon-houdini/pull/200/commits/385a239730cc2926528d813d77f13875baac954d

client/ayon_houdini/api/plugin.py
MustafaJafar: https://github.com/ynput/ayon-houdini/pull/200/commits/385a239730cc2926528d813d77f13875baac954d

client/ayon_houdini/api/plugin.py
```diff
@@ -312,6 +336,42 @@ def apply_settings(self, project_settings):
...
+        staging_dir_info = super(HoudiniCreator, self).get_staging_dir(instance)
+
+        staging_dir_path =  self.default_staging_dir        
+        if staging_dir_info is not None:
+            staging_dir_path = staging_dir_info.directory
+    
+        return staging_dir_path.replace("\\", "/").rstrip("/")
+
+    def set_node_staging_dir(self, node: hou.Node, staging_dir: str, instance: CreatedInstance, pre_create_data: dict):
```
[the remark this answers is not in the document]
MustafaJafar: https://github.com/ynput/ayon-houdini/pull/200/commits/385a239730cc2926528d813d77f13875baac954d

client/ayon_houdini/api/plugin.py
MustafaJafar: https://github.com/ynput/ayon-houdini/pull/200/commits/385a239730cc2926528d813d77f13875baac954d

MustafaJafar:
> 1. Shouldn't we always use staging dir logic? I see it is defaulting to `$HIP/ayon` so probably not, but that's why there is question mark.

This is the main point of this PR to provide default value if there's no staging directory configured.
This default value is overridden by the staging directory configuration. 
Default value and staging are only used if the path management setting is enabled. 

> 2. What `enable_staging_path_management` acually do? We need staging all the time, or?

It completely disables the default value and staging. 

> 3. I think the settings to expand variables makes it uber confusing. I guess you can't use `$HIP/ayon` anymore? Does it make sense to have the option to change it?

When enabling expanding the `$HIP` evaluates to the workfile location. Some admins may prefer to disable that setting to leave the output directory relative to the workfile location.

> If 1. makes sense, then `expand_staging_dir` is kinda requirement for `enable_staging_path_management` automatically, or?

`expand_staging_dir` is not a requirement it's more of a personal preference.

client/ayon_houdini/plugins/create/create_vray_rop.py
MustafaJafar: https://github.com/ynput/ayon-houdini/pull/200/commits/498e473a147502ee3c310cd5ea51afb15fa59115

tatiana-ynput:
Moving to Blocked as its not a priority for a client atm

client/ayon_houdini/api/lib.py
```diff
@@ -1687,3 +1687,24 @@ def save_slapcomp_to_file(
 
     # Clean up.
     sopnet.destroy()
+
+
+def expand_houdini_string(text: str, pattern=r"\$[a-zA-Z0-9_]+") -> str:
```
MustafaJafar: Opening a fresh conversation about your question in https://github.com/ynput/ayon-houdini/pull/200/files#r1932144992 

> Does the `hou.text.expandString` `expand_tilde` argument suffice for our needs? or do we still need to wrap this?

Firstly, what are "our need" here? Is it expected for admins to pick a default output directory with `~` which refers to user data like ``~/ayon/`chs("AYON_productName")` `` ?

Also, there's a bug the `tilda` is not expanded because we are expanding the matched sting only which doesn't include `~`.. also, that's why `` `chs("AYON_productName")` ``  is not expanded.
```python
import hou
from ayon_houdini.api.lib import expand_houdini_string

# ~/documents/ayon/`chs("AYON_productName")`/$OS.abc
text = hou.node("/out/cameraMain").parm('filename').unexpandedString()

text_expanded_ayon = expand_houdini_string(text, pattern=r"\$[a-zA-Z0-9_]+")
text_expanded_ayon2 = expand_houdini_string(text, pattern=r"(\$[a-zA-Z0-9_]+|~)")

print(">>>", text_expanded_ayon)
print(">>>", text_expanded_ayon2)
```
output
```python
>>> ~/documents/ayon/`chs("AYON_productName")`/Director.abc
>>> C:/Users/MUSTAF~1/DOCUME~1/documents/ayon/`chs("AYON_productName")`/Director.abc
```

MustafaJafar:
I believe this PR is missing any product types that were added since this PR set on hold.

client/ayon_houdini/api/lib.py
BigRoy: Personally I'm fine with keeping the `~` representing the user folder. If someone is using that for whatever reason they may have their reasons - doesn't seem like something we'd want to expand.

But that expanded URL turning into a path containing `~` should be avoided if we can.
Does that make sense?

MustafaJafar: > But that expanded URL turning into a path containing `~` should be avoided if we can.
> Does that make sense?

I believe you are referring to something like `C:/Users/MUSTAF~1/DOCUME~1/`. 
Actually, the current state of this PR doesn't expand `~` as the current regex pattern `\$[a-zA-Z0-9_]+` doesn't capture it.
```python
import hou
from ayon_houdini.api.lib import expand_houdini_string

# ~/documents/ayon/`chs("AYON_productName")`/$OS.abc
text = hou.node("/out/cameraMain").parm('filename').unexpandedString()
text_expanded_ayon = expand_houdini_string(text, pattern=r"\$[a-zA-Z0-9_]+")

print(">>>", text_expanded_ayon)

```

MustafaJafar: It's not needed as the function just ignores tilda.

For reference, the function `expand_houdini_string` uses Houdini's function [hou.text.expandString](https://www.sidefx.com/docs/houdini/hom/hou/text.html) to expand the string. however, we want this expansion to be selective. 
Here are some example inputs and outputs of the function.
```python
"""
    Example:
        >>> expand_houdini_string("~/documents/ayon/`chs('AYON_productName')`", pattern=r"\$[a-zA-Z0-9_]+")
        "~/documents/ayon/`chs('AYON_productName')`"
        >>> expand_houdini_string("./ayon/`chs('AYON_productName')`", pattern=r"\$[a-zA-Z0-9_]+")
        "./ayon/`chs('AYON_productName')`"
        >>> expand_houdini_string("$HIP/ayon/`chs('AYON_productName')`", pattern=r"\$[a-zA-Z0-9_]+")
        "H:/AYON/projects/AY_CG_demo/assets/characters/sloth/work/FX/ayon/`chs('AYON_productName')`"
        >>> expand_houdini_string("$HIP/ayon/`chs('AYON_productName')`", pattern= r"`[^`]+`")  # Used for HDA product.
        "$HIP/ayon/"
"""
```

BigRoy:
> Kuba suggested `raise NotImplementedError` to avoid developer errors.

Fine with me if there's no hard requirement for having it implemented.

client/ayon_houdini/api/lib.py
BigRoy: No - I think this is a decent starting point, and let's see where it ends up being used.

client/ayon_houdini/api/plugin.py
```diff
@@ -361,6 +395,61 @@ def apply_settings(self, project_settings):
...
+
+        This method falls back to the default output path defined in settings
+        `ayon+settings://houdini/general/rop_output/default_output_dir`
+
+        Args:
+            instance (CreatedInstance): Instance for which should be staging
+                dir gathered.
+
+        Returns:
+            str: Staging dir path
```
BigRoy: Return type stated here is different than above on the function?

BigRoy (approved):
Overall this looks good and seems to work - I just had some minor closing comments worth to look at.

client/ayon_houdini/api/plugin.py
MustafaJafar: https://github.com/ynput/ayon-houdini/pull/200/commits/cb185d2d9b51c62122d405d42490f837cf1f8225

server/settings/general.py
```diff
@@ -36,6 +72,11 @@ class GeneralSettingsModel(BaseSettingsModel):
 
 DEFAULT_GENERAL_SETTINGS = {
     "add_self_publish_button": False,
+    "set_rop_output": {
+        "enabled": True,
+        "expand_vars": False,
+        "default_output_dir": "$HIP/ayon/`chs(\"AYON_productName\")`"
```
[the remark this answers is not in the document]
MustafaJafar: This is an old discussion. maybe started in https://github.com/ynput/OpenPype/pull/3839
```
        filepath = "{root}/{subset}/{subset}.$F4.{ext}".format(
            root=hou.text.expandString("$HIP/pyblish"),
            subset="`chs(\"subset\")`",  # keep dynamic link to subset
            ext=pre_create_data.get("image_format") or "png"
        )
```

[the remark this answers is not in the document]
BigRoy: Yes - but... that was the filename to my knowledge, not the folder? Anyway - not critical for me, it's a setting so customizable. Just in my test run it felt a bit verbose.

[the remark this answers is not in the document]
MustafaJafar: Well, I can pull my magic wand and move it to all file names?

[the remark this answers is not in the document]
BigRoy: Since that one _isn't_ customizable as far as I can see I'd rather keep it as is. 👼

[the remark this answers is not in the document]
BigRoy: Yup, I see - thanks. It's mostly on the ROPs that I rarely used... those that I did didn't have it that's why it stood out.
Absolutely fine that it's unified in that way, thanks.

BigRoy (approved):

moonyuet (approved):
Image
The publishing of the general product type is correctly set to the custom staging from the core addon while the publishing of the render product(aka ROPs) is correctly located to the custom ROP staging

PR #202, opened 2024-12-23
Bugfix/201 fix some typo and correct expected return types

### LINKED ISSUE #201: Fix some typo and correct expected return types

## Current Behavior
Issue mentioned here https://github.com/ynput/ayon-houdini/pull/200#issuecomment-2559222904

## Expected Behavior
it should work fine

### Steps To Reproduce:

1. create model
2. change frame range
3. publish

### Version

Windows

### Relevant log output:

```shell

```

MustafaJafar:
## Changelog Description
resolve #201 

Note: 
`get_invalid` should return `None` or list of `hou.Node`/ `hou.OpNode`

## Testing notes:
1. Error reported in the linked issue should be fixed
2. Everything should work as before

### DIFF
--- a/client/ayon_houdini/plugins/publish/validate_export_is_a_single_frame.py
+++ b/client/ayon_houdini/plugins/publish/validate_export_is_a_single_frame.py
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
 """Validator for checking that export is a single frame."""
+import hou
+
 from ayon_core.pipeline import (
     PublishValidationError,
     OptionalPyblishPluginMixin
@@ -35,8 +37,6 @@ class ValidateSingleFrame(plugin.HoudiniInstancePlugin,
     @classmethod
     def get_invalid(cls, instance):
 
-        invalid = []
-
         frame_start = instance.data.get("frameStartHandle")
         frame_end = instance.data.get("frameEndHandle")
 
@@ -48,12 +48,12 @@ class ValidateSingleFrame(plugin.HoudiniInstancePlugin,
             return
 
         if frame_start != frame_end:
-            invalid.append(instance.data["instance_node"])
+            rop = hou.node(instance.data["instance_node"])
             cls.log.error(
                 "Invalid frame range on '%s'."
                 "You should use the same frame number for 'f1' "
                 "and 'f2' parameters.",
-                instance.data["instance_node"].path()
+                rop.path()
             )
 
-        return invalid
+        return [rop]
--- a/client/ayon_houdini/plugins/publish/validate_mkpaths_toggled.py
+++ b/client/ayon_houdini/plugins/publish/validate_mkpaths_toggled.py
@@ -30,6 +30,6 @@ class ValidateIntermediateDirectoriesChecked(plugin.HoudiniInstancePlugin):
         for node in instance[:]:
             if node.parm("mkpath").eval() != 1:
                 cls.log.error("Invalid settings found on `%s`" % node.path())
-                result.append(node.path())
+                result.append(node)
 
         return result
--- a/client/ayon_houdini/plugins/publish/validate_render_product_paths_unique.py
+++ b/client/ayon_houdini/plugins/publish/validate_render_product_paths_unique.py
@@ -95,7 +95,7 @@ class ValidateRenderProductPathsUnique(plugin.HoudiniContextPlugin,
         # Get instances matching this plugin families
         instances = pyblish.api.instances_by_plugin(list(context), cls)
         if not instances:
-            return []
+            return
 
         # Get expected rendered filepaths
         paths_to_instance_id = defaultdict(list)
@@ -119,7 +119,7 @@ class ValidateRenderProductPathsUnique(plugin.HoudiniContextPlugin,
                 invalid_paths.append(path)
 
         if not invalid_instance_ids:
-            return []
+            return
 
         # Log invalid sequences as single collection
         collections, remainder = clique.assemble(invalid_paths)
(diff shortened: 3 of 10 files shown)


moonyuet (changes requested):
`ValidateSingleFrame` has been errored out 
image
[Uploading publish-report-241224-17-57.json…]()

client/ayon_houdini/plugins/publish/validate_export_is_a_single_frame.py
```diff
@@ -48,12 +48,12 @@ def get_invalid(cls, instance):
...
             cls.log.error(
                 "Invalid frame range on '%s'."
                 "You should use the same frame number for 'f1' "
                 "and 'f2' parameters.",
-                instance.data["instance_node"].path()
+                rop.path()
             )
 
-        return invalid
+        return [rop]
```
moonyuet, line 59: You dont need to remove invalid variable as you can use `invalid` to append `instance.data["instance_node"]` value and  then return `invalid` to solve the variable issue found in the requested changes I showed

MustafaJafar: my bad. I didn't notice the indentation.
https://github.com/ynput/ayon-houdini/pull/202/commits/6caa2c826e42e17b09d80f2c626fc05d1ffd4119

moonyuet (approved):
Looks good now!
image

BigRoy:
I wonder if it makes sense to let our IDEs help us out a hand in the future - and add Type Hints to the `get_invalid` methods so it's hinted to be `Optional[List["hou.Node"]]`?

MustafaJafar:
> I wonder if it makes sense to let our IDEs help us out a hand in the future - and add Type Hints to the `get_invalid` methods so it's hinted to be `Optional[List["hou.Node"]]`?

This will be cool! 
Note: 
It's expected that `get_invalid` returns `Optional[List["hou.Node"]]` only when we are using select invalid action.

client/ayon_houdini/plugins/publish/validate_mkpaths_toggled.py
```diff
@@ -30,6 +30,6 @@ def get_invalid(cls, instance):
         for node in instance[:]:
             if node.parm("mkpath").eval() != 1:
                 cls.log.error("Invalid settings found on `%s`" % node.path())
-                result.append(node.path())
+                result.append(node)
```
BigRoy, line 33: Making this change does mean that the validation error itself will now present itself with a less readable node path. So maybe in the raised `PublishValidationError` we also want to make sure we still format using the rop's path?

BigRoy (commented):
Overall the change looks good -  but in most cases you're not refactoring how we're formatting the node into the `PublishValidationError` which means the message there will turn out worse. So we may want to make sure there that we're still formatting using the `path` of the nodes.

BigRoy:
> Note: It's expected that `get_invalid` returns `Optional[List["hou.Node"]]` only when we are using select invalid action.

Yup, am aware - but I have a feeling that should be 99% of the use cases of the `get_invalid` class methods or static methods. (And I'd rather have it for all of them to avoid confusion if suddenly one does not behave that way.) And if there's indeed a reason for one of them to deviate, hopefully seeing the type hints differ there will also help make sense of things.

client/ayon_houdini/plugins/publish/validate_mkpaths_toggled.py
MustafaJafar: I believe this should work https://github.com/ynput/ayon-houdini/pull/202/commits/475c73f47889e3a5dae0b6fbf944e9b9eaf93096

MustafaJafar:
> > Note: It's expected that `get_invalid` returns `Optional[List["hou.Node"]]` only when we are using select invalid action.
> 
> Yup, am aware - but I have a feeling that should be 99% of the use cases of the `get_invalid` class methods or static methods. (And I'd rather have it for all of them to avoid confusion if suddenly one does not behave that way.) And if there's indeed a reason for one of them to deviate, hopefully seeing the type hints differ there will also help make sense of things.

what about implementing a  `get_invalid`  on `HoudiniInstancePlugin` and `HoudiniContextPlugin`.

client/ayon_houdini/plugins/publish/validate_render_product_paths_unique.py
```diff
@@ -95,7 +95,7 @@ def get_invalid(cls, context) -> "List[hou.Node]":
         # Get instances matching this plugin families
         instances = pyblish.api.instances_by_plugin(list(context), cls)
         if not instances:
-            return []
+            return
```
BigRoy, line 98: This does not meet the return type on this method

MustafaJafar: https://github.com/ynput/ayon-houdini/pull/202/commits/6cc88bde1e694fde3869b6331343887f5344a954

client/ayon_houdini/plugins/publish/validate_subset_name.py
```diff
@@ -39,17 +39,14 @@ def process(self, instance):
 
         invalid = self.get_invalid(instance)
         if invalid:
-            nodes = [n.path() for n in invalid]
             raise PublishValidationError(
                 "See log for details. "
-                "Invalid nodes: {0}".format(nodes)
+                "Invalid nodes: {0}".format(invalid[0].path())
```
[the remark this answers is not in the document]
MustafaJafar: https://github.com/ynput/ayon-houdini/pull/202/commits/c8393db638f9e2b2e8eb712df6d8f8e59c2f702e

moonyuet (approved):
Tested with publishing the product types in Houdini. Validators looks good for me. 
image

moonyuet:
One thing is a worth mention that somehow validate resolution setting does not validate the resolution correctly in my case. Even if I set the resolution back to 1920X1080, the validator still blocks me from publishing the render
image

MustafaJafar:
> One thing is a worth mention that somehow validate resolution setting does not validate the resolution correctly in my case. Even if I set the resolution back to 1920X1080, the validator still blocks me from publishing the render (Do it in separate PR as it's not related to this PR) image

This is weird, I was not able to replicate it on my side. 
could you create an issue with reproducible steps ?

Issue #210, opened 2025-01-27
AY-7399_Houdini: Collapse loader/saver paths to AYON root vars so they are cross-OS compatible.


ynbot:
### Please describe the feature you have in mind and explain what the current shortcomings are?

Collapse all project root vars in AYON to Houdini variable equivalents, so that all the project paths start with that variable inside houdini and then based on what OS I'm running the variable would be expanded to G:... or /mnt/...

This would make the file work out of the box on different platforms if they were using e.g. $AYON_ROOT in their paths instead of using absolute paths. This could be used for Loader paths, but also for default ROP output paths (in cases where $HIP may not make sense?)

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1332464559469887669/1332478326530506823)
_(might be a private channel)_

_This issue was automatically created from Clickup ticket [AY-7399](https://app.clickup.com/t/86c1t6cxh)_

PR #212, opened 2025-01-28
Collapse loaded paths to AYON project root env vars

BigRoy:
## Changelog Description

Collapse loaded paths to AYON project root env vars

Fix the loader part of #210 

Loading assets should now load the paths using e.g. `$AYON_PROJECT_ROOT_WORK` variable instead of the absolute path making the scene directly cross-OS compatible within AYON.

> [!Note]
> One important note, it does not work with generic loaders and consequently does not work with loader HDAs. Which isn't a problem because those are **always** live evaluated on scene open based on your OS and local system - so should still evaluate to the correct paths dynamically as needed.

### TODO

- [x] Optimize the logic for collapsing the paths?
- [x] Allow enabling/disabling of the feature

## Testing notes:

Enable the setting: `ayon+settings://houdini/load/collapse_path_to_project_root_vars`

image

1. Loading assets should now load the paths using e.g. `$AYON_PROJECT_ROOT_WORK` variable.
2. Loading should work.
3. Updating loaded containers should work.


client/ayon_houdini/api/plugin.py
```diff
@@ -325,7 +326,27 @@ def filepath_from_context(cls, context):
...
 
-        return super(HoudiniLoader, cls).filepath_from_context(context)
+        path = super().filepath_from_context(context)
+
+        # Remap project roots to the relevant environment variables
+        # TODO: Refactor this to more optimal/better code
+        mapping = {}
+        for key, value in os.environ.items():
+            if key.startswith("AYON_PROJECT_ROOT_"):
+                mapping[key] = value
```
MustafaJafar, line 336: I'm not sure if this is faster, but we may not need to go through the env keys because the anatomy object already has this piece of information. 
I wonder if there's a cached anatomy object instead of creating/querying it. 
```suggestion
        anatomy = Anatomy(context["project"]["name"], project_entity=context["project"])
        mapping = anatomy.root_environments()
```

client/ayon_houdini/api/plugin.py
```diff
@@ -325,7 +326,27 @@ def filepath_from_context(cls, context):
...
+            match_path = path.replace("\\", "/")
+            # Sort by length to ensure that the longest matching key is first
+            # so that the nearest matching root is used
+            for key, value in sorted(mapping.items(),
+                                     key=lambda x: len(x[1]),
+                                     reverse=True):
+                if value and match_path.startswith(value.replace("\\", "/")):
+                    # Replace start of string with the key
+                    path = f"${key}" + path[len(value):]
+                    break
```
MustafaJafar, line 347: I wonder if going the other way is easier to understand and maintain. 
I'm assuming the longest match will result in a smaller string (unless the root name is big enough to break this assumption)
Also, we can skip sorting if we the matching is one path.
```suggestion
        paths = []
        for key, value in mapping.items():
            if value and path.startswith(value.replace("\\", "/")):
                # Replace start of string with the key
                paths.append(f"${key}{path[len(value):]}")

        if len(paths) == 1:
            return paths[0]

        # sort and return the smallest path.
        return sorted(paths)[0]
```

client/ayon_houdini/api/plugin.py
BigRoy: I'm thinking it may be better to just 'cache it once' in the current session since that part of `os.environ` (for the roots) shouldn't update on context changes regardless. Right @iLLiCiTiT ?

client/ayon_houdini/api/plugin.py
BigRoy: I like this less because it'd mean we'd always be checking against all roots, even if we had already matched one. So it'd perform less optimal. However, sure - the amount of roots may never grow that large that it'd really be a problem.

Overall I just did a prototype here and did not look at 'cleaning the code' at all - it was merely testing whether this _could_ be functional.

client/ayon_houdini/api/plugin.py
iLLiCiTiT: > I mostly refrained from doing this due to speed reasons - getting the Anatomy object usually results in quite some queries which I hoped to avoid.

If you pass if `project_entity` it does not do any additional queries.

But what if you're loading from library project? The environments are completelly different in that case, or?

MustafaJafar: Does having the key in os.environ mean it's actually added to hscript or doesn't it?
Honestly, I don't know because the last item I plated with env vars I was checking if the variables is set in both environ and hscript. https://github.com/ynput/ayon-houdini/blob/26f8cfb9d7f1b0a3a651d89161a825e8617a7d91/client/ayon_houdini/api/lib.py#L1013

BigRoy: It is not set to `hscript` no, however houdini vars still expand them.

Houdini expands them in order:
- is hscript var? -> expand
- is env var? -> expand

iLLiCiTiT: BTW with Anatomy you could just use this and skip `filepath_from_context`.
```python
env_roots = anatomy.roots.root_environmets_fill_data()
return get_representation_path(representation, env_roots)
```

BigRoy: Does that _leave_ the roots in tact? Also, good point on only supporting _current project_ there!

> The envs were aded in Pype 2. We didn't have real usage for them since OpenPype, but didn't remove them.

If we do remove them - then it makes sense to explicitly 'enable' them in `ayon-houdini` to support this Houdini feature. We can make a PR for that once we need at that point in time.

MustafaJafar (commented):
`client/ayon_houdini/plugins/load/load_image.py` is failing because of 

https://github.com/ynput/ayon-houdini/blob/26f8cfb9d7f1b0a3a651d89161a825e8617a7d91/client/ayon_houdini/plugins/load/load_image.py#L137-L138

client/ayon_houdini/api/plugin.py
iLLiCiTiT: > Does that leave the roots in tact? Also, good point on only supporting current project there!

Default behavior of `root_environmets_fill_data` is to return env keys as values for root `{"work": "$AYON_AYON_PROJECT_ROOT_WORK", ...}` -> which is what you're guessing here, whereas the suggested code does fill it with the root that is really requested by the template.

client/ayon_houdini/api/plugin.py
MustafaJafar: > the amount of roots may never grow that large that it'd really be a problem.

I agree. but, the code just looks hard to read that's why I tried picking a different approach.

client/ayon_houdini/api/plugin.py
BigRoy: Ended up sticking closer to what I already had - just felt more reliable trusting on values that Houdini would actually currently detect as valid keys - no matter what the anatomy may say currently.

client/ayon_houdini/api/plugin.py
```diff
@@ -310,13 +313,59 @@ class HoudiniLoader(load.LoaderPlugin):
...
+    def apply_settings(cls, project_settings):
+        # Prepare collapsible variable mapping using entries in `os.environ`
+        # that are set to the project root paths
+        cls.collapse_paths_to_root_vars: bool = (
+            project_settings["houdini"]["load"]
+            .get("collapse_path_to_project_root_vars", False)
+        )
+
+        super().apply_settings(project_settings)
+
```
iLLiCiTiT, line 328: Looks like this is not needed at all
```suggestion
```

BigRoy: Ah yes - valid point. I originally had more logic in there as "cache" but wasn't actually more optimal. Thanks.

MustafaJafar: It has stopped working since applying this change.

MustafaJafar:
> `client/ayon_houdini/plugins/load/load_image.py` is failing because of
> 
> https://github.com/ynput/ayon-houdini/blob/26f8cfb9d7f1b0a3a651d89161a825e8617a7d91/client/ayon_houdini/plugins/load/load_image.py#L137-L138

Further Info, It turns out it's just the image loader but also these loaders are affected too.

image

client/ayon_houdini/api/plugin.py
BigRoy: Of course, totally missed that - it needs to be there because it's NOT the default apply settings behavior. Re-added it.

BigRoy:
> > `client/ayon_houdini/plugins/load/load_image.py` is failing because of
> > https://github.com/ynput/ayon-houdini/blob/26f8cfb9d7f1b0a3a651d89161a825e8617a7d91/client/ayon_houdini/plugins/load/load_image.py#L137-L138
> 
> Further Info, It turns out it's just the image loader but also these loaders are affected too.
> 
> image

Thanks @MustafaJafar - I removed these checks. I'm pretty sure these originated from the 'legacy' way of how some of the frame logi was implemented where we were querying the files on disk to define the frame filenames, etc. but that has been removed a long time ago and the logic doesn't actually require `os` to touch the files, it's only there for path manipulation still.

client/ayon_houdini/plugins/load/load_image.py
```diff
@@ -134,26 +129,20 @@ def remove(self, container):
...
-        filename = filename.replace("\\", "/")
+                              filename)
+            path = os.path.join(folder, filename)
 
-        return filename
+        path = os.path.normpath(path)
+        path = path.replace("\\", "/")
+        return path
 
     def get_colorspace_parms(self, representation: dict) -> dict:
```
MustafaJafar, line 147: There's a bug in this function (not related to this PR)
if there's no colorspace, it returns `None` which cause `parms.update` to error.

BigRoy: Would you like me to set up a separate fix PR?

MustafaJafar (commented):
This PR works. 
One important note, it does not work with generic loaders and consequently does not work with loader HDAs.

client/ayon_houdini/plugins/load/load_image.py
MustafaJafar: Never mind, I'll shoot one.

BigRoy:
> One important note, it does not work with generic loaders and consequently does not work with loader HDAs.

Correct - however, those are always live evaluated on your local machine. Hence, should not introduce an issue when opening the scene regardless.

MustafaJafar (approved):

PR #213, opened 2025-01-29
Fix Load Filepath legacy update

BigRoy:
## Changelog Description

Fix update of legacy load filepaths

Make function non-private because it's used [here](https://github.com/ynput/ayon-houdini/blob/c31d90e7a03271f98053abd1648753ee11331f8b/client/ayon_houdini/plugins/load/load_filepath.py#L61).

## Testing notes:

1. Legacy load filepaths should be able to update.


MustafaJafar (approved):
Looks good.

PR #214, opened 2025-01-29
Change render creator product types to `render` so they are all just render products

BigRoy:
## Changelog Description

Change render creator product types to `render` so they are all just render products

This introduces a settings toggle for backwards compatibiliity: `ayon+settings://houdini/create/render_rops_use_legacy_product_type`

image

When copying settings from Addon versions before 0.5.1 this checkbox will automatically enable on the copy to remain backwards compatible.

_Note:_ Due to limitations in the setting conversion logic this will **also** flip the toggle on Studio Settings. See https://github.com/ynput/ayon-backend/issues/581

The changes for product types are:
- `arnold_rop` -> `render`
- `karma_rop` -> `render`
- `mantra_rop` -> `render`
- `redshift_rop` -> `render`
- `usdrender` -> `render`
- `vray_rop` -> `render`

Now each of those would end up in e.g. `renderLightingMain` (if it uses the `render` product name profile by default because that includes `{task}`)

⚠️ This PR also requires `ayon-deadline` change: https://github.com/ynput/ayon-deadline/pull/172 and hence should have a soft compatibility requirement to a future release that includes that fix. ⚠️ 

### Why?

This came up by @MustafaJafar who mentioned that to target these individual renderer specific ROPs in settings with profiles one would need to know to use `karma_rop`, `vray_rop`, etc. previously even though technically they are all just renders.

By changing this - what's instantly noticed is that now for example it also picks up the default `render` creator product name profile that includes the task type, generating e.g. `renderLightingMain` (with Lighting task) by default instead of `karma_ropMain`.

## Testing notes:

1. Rendering should still work


BigRoy:
This might be better solved in a backwards compatible manner with "aliases" potentially.

However, if it were - then the question rises. Which of the aliases would be the one to pick in e.g. profiles filtering for `Product Name Profiles` or `Custom Staging Dir` profiles.

antirotor:
we were in the similar situation with.... what was it.... review product? With the underscore after, or something like that - basically change that will be backwards incompatible in very hard way (break version streams, profiles, everything based on product name). I think we discussed then that it is fine unless done automatically on running project so - we should add this option to settings and clearly state that from next version this will be by default. I think that we still need some clear communication channel about these breaking changes to TD installing that particular addon - something on addon market.

BigRoy:
Well, yes - existing scenes will still work. But if you were to create a new one, you wouldn't be able to produce the same named outputs as before. Which is, still an incompatibility. ;)

But maybe does reduce the level of disaster it actually is.

MustafaJafar:
Fun fact, the `AttachReviewables` won't be visible in the publisher with render instances in Houdini without this PR. 
I'll keep you updated if there are any needed additional changes.

MustafaJafar:
> Well, yes - existing scenes will still work. But if you were to create a new one, you wouldn't be able to produce the same named outputs as before. Which is, still an incompatibility. ;)

Yeah, I can confirm.
image

what about adding a toggle `Use legacy names`?

MustafaJafar:
> The question is, how can we move it as much as possible to the new behavior by default.

Do you mean : when disabling the setting by default, How can we enable it by default.

BigRoy:
> > The question is, how can we move it as much as possible to the new behavior by default.
> 
> Do you mean : when disabling the setting by default, How can we enable it by default?

How can we ensure that just to now support 'backwards compatible' we won't still be using the legacy way in years to come just because we decided to default the new way to be off by default.

E.g. I'd prefer NEW projects to default to the new way, but existing projects to default to the old for example.

MustafaJafar:
> How can we ensure that just to now support 'backwards compatible' we won't still be using the legacy way in years to come just because we decided to default the new way to be off by default.
> 
> E.g. I'd prefer NEW projects to default to the new way, but existing projects to default to the old for example.

how have we dealt with `use_legacy_product_names_for_renders`?
image

MustafaJafar:
There's something weird issue in this PR when publishing to Deadline. 
Instances created using it, get a different plugin in deadline. 

Note in the screenshot below there are two jobs with a cube icon, these always fail on my side unlike the other jobs with render, they have the regular houdini icon.
image

They have `Render` plugin? yeah, that is arnold.
image


I've found a nasty bug! 

https://github.com/ynput/ayon-deadline/blob/dc980a4a75a840013a2c5fabc86a64d20b121dae/client/ayon_deadline/plugins/publish/houdini/submit_houdini_render_deadline.py#L241-L278

server/settings/conversion.py
```diff
@@ -15,10 +23,22 @@ def _convert_validate_subset_name(overrides: dict[str, Any]) -> None:
             "ValidateSubsetName"
         )
 
+def _enable_create_render_rops_use_render_product_type(
+        overrides: dict[str, Any]
+) -> None:
+    # Enforce render creators `render_rops_use_render_product_type` to True
+    # to remain backwards compatible with older versions
+    create = overrides.setdefault("create", {})
+    create["render_rops_use_legacy_product_type"] = True
```
BigRoy, line 32: Is there any way to apply this backward compatible override like this only to project overrides, and not to studio overrides so that the studio default for new project still start of using the NEW default, even when updating from older addons in that particular AYON server?

client/ayon_houdini/api/plugin.py
```diff
@@ -362,6 +362,25 @@ def apply_settings(self, project_settings):
             setattr(self, key, value)
 
 
+class RenderLegacyProductTypeCreator(HoudiniCreator):
```
BigRoy, line 365: I was wondering whether this would've been possible with some type of MixIn class that didn't require inheriting from `HoudiniCreator` so it's more explicit that it's just a backward compatible 'fix' on the Creators, but couldn't find any sensible way to still be able to access the upstream `apply_settings`...

Not sure if @iLLiCiTiT has any ideas how to do this in a nicer way.

server/settings/conversion.py
BigRoy: Confirmed by @martastain and @iLLiCiTiT 

There currently isn't.
And:
> tbf: it never occured to me this may be needed since they use the same model

client/ayon_houdini/api/plugin.py
iLLiCiTiT: Is `legacy_product_type` used only in `product_type` property?

BigRoy: Thanks, that's actually what I did at first.. but my IDE complained because in the parent class `product_type` is defined as a property which means it actually can't be assigned to ;) ... the only reason it works on other classes is because we do it at the class definition and set class attribute:
```python
class MyCreator(Creator):
    product_type = "render"
    ...
```
Which essentially takes it back from a property to just a string attribute... but just from the inheritance from `Creator` it actually isn't.

iLLiCiTiT: > Thanks, that's actually what I did at first.. but my IDE complained because in the parent class product_type is defined as a property which means it actually can't be assigned to ;)

Yeah, but it is abstract property, meaning it has to be overriden. I would not worry about IDE if you're handling backwards compatibility to avoid exactly what you did...

BigRoy: Tweaked it with [`7599e07` (#214)](https://github.com/ynput/ayon-houdini/pull/214/commits/7599e07e4b163c1a6d9840efab852bd42ac9f2c5)

MustafaJafar (approved):
It works, using different render targets work, using legacy and new works, copy settings conversion works (it falls back to false when copying from older addon). 
 
This makes much sense now. Thanks!
1. I can tell which runtime instances are reltated to which instance. 
    - currently 
    image
    - legacy
    image
2. I can render the same render product regardless of the renderer. and I can see it visually before triggering the validation or publish
image
I did it intentiall here. 
image
_Maybe the next milestone to sync AOVs products._ 
3. Attach render/review products logic is working now.
image

MustafaJafar:
FYI, this https://github.com/ynput/ayon-houdini/pull/257 is needed to have a consistent naming when rendering locally and on farm.

PR #215, opened 2025-02-03
Chore: Stricter ruff rules

iLLiCiTiT:
## Changelog Description
Make ruff rules stricter.

Modified code accordingly to follow stricter rules. There is one line where I don't know why it is done the way containing ``f' {file_parm.node().path()} {file_parm.name()} \`{expression}\`'`` where ruff suggesto to skip espace char `\` -> can be changed.

## Testing notes:
1. Validate changes


client/ayon_houdini/api/lib.py
```diff
@@ -1595,7 +1596,7 @@ def connect_file_parm_to_loader(file_parm: hou.Parm):
     expression = rf'chs\(\"{relative_path}/file\"\)'  # noqa
     hou.hscript(
         'opparm -r'
-        f' {file_parm.node().path()} {file_parm.name()} \`{expression}\`'
+        f' {file_parm.node().path()} {file_parm.name()} \`{expression}\`'  # noqa W605
```
BigRoy, line 1599: Yeah, not sure about this actually - I believe since we're passing this to `hscript` we need to "escape" the character for that command's execution.

However, I think the correct fix here is to use `rf` instead of `f` strings:
```suggestion
        rf' {file_parm.node().path()} {file_parm.name()} \`{expression}\`'
```
Likely ruff then does allow it.

iLLiCiTiT: But that effectivelly converts it to `` {file_parm.node().path()} {file_parm.name()} \\`{expression}\\` `` -> I don't understand what is the reasoning behind the backslashes, and I can't test it so I can't test it. It is either not having them, or having them doubled (escaped).

iLLiCiTiT: ok, double backslash it is

BigRoy (approved):
Code changes look good to me - I didn't do a test run. Maybe @MustafaJafar can do a quick run?

MustafaJafar (approved):
Everything seems to work as usual.

PR #219, opened 2025-02-11
Create destination folder if it doesn't already exist when saving

### LINKED ISSUE #218: Houdini fails to save if the work directory is not created

## Current Behavior
.

## Expected Behavior
.

### Steps To Reproduce:

Mostly programmatically, by opening a hip file, then trying to save to a path that doesn't exist.
1. Launch Houdini
2. Run this in the source code editor
```
from ayon_core.pipeline import registered_host

host = registered_host()

file_path = r"{location to a folder or task that exist already on disk}\my_file.hip"
host.save_workfile(file_path)
```

### Version

Windows
Linux - Rocky9
Linux - Ubuntu
MacOS

### Relevant log output:

```shell
The attempted operation failed.
  File "hou.session", line 6, in <module>
  File "E:\Ynput\ayon-houdini\client\ayon_houdini\api\pipeline.py", line 111, in save_workfile
    hou.hipFile.save(file_name=dst_path,
  File "C:\PROGRA~1/SIDEEF~1/HOUDIN~1.370/houdini/python3.11libs\houpythonportion\ui.py", line 1158, in decorator
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "C:\PROGRA~1/SIDEEF~1/HOUDIN~1.370/houdini/python3.11libs\hou.py", line 57326, in save
    return _hou.hipFile_save(self, file_name, save_to_recent_files)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error creating file: H:/AYON/projects/Experiments/things/Folder/work/Generic/my_file.hip
```

MustafaJafar:
## Changelog Description
resolve #218 

## Testing notes:
1. Launch Houdini
2. Run this in the source code editor
```
from ayon_core.pipeline import registered_host

host = registered_host()

file_path = r"{location to a folder or task that exist already on disk}\my_file.hip"
host.save_workfile(file_path)
```

### DIFF
--- a/client/ayon_houdini/api/pipeline.py
+++ b/client/ayon_houdini/api/pipeline.py
@@ -108,6 +108,11 @@ class HoudiniHost(HostBase, IWorkfileHost, ILoadHost, IPublishHost):
         # Force forwards slashes to avoid segfault
         if dst_path:
             dst_path = dst_path.replace("\\", "/")
+
+        # Create destination folder if it doesn't already exist.
+        if not os.path.exists(os.path.dirname(dst_path)):
+            os.makedirs(os.path.dirname(dst_path))
+
         hou.hipFile.save(file_name=dst_path,
                          save_to_recent_files=True)
         return dst_path
--- a/client/ayon_houdini/api/pipeline.py
+++ b/client/ayon_houdini/api/pipeline.py
@@ -110,7 +110,7 @@ class HoudiniHost(HostBase, IWorkfileHost, ILoadHost, IPublishHost):
             dst_path = dst_path.replace("\\", "/")
 
         # Create destination folder if it doesn't already exist.
-        if not os.path.exists(os.path.dirname(dst_path)):
+        if dst_path and not os.path.exists(os.path.dirname(dst_path)):
             os.makedirs(os.path.dirname(dst_path))
 
         hou.hipFile.save(file_name=dst_path,


client/ayon_houdini/api/pipeline.py
```diff
@@ -108,6 +108,11 @@ def save_workfile(self, dst_path=None):
         # Force forwards slashes to avoid segfault
         if dst_path:
             dst_path = dst_path.replace("\\", "/")
+
+        # Create destination folder if it doesn't already exist.
+        if dst_path and not os.path.exists(os.path.dirname(dst_path)):
+            os.makedirs(os.path.dirname(dst_path))
```
BigRoy, line 114:
```suggestion
        if dst_path:
            os.makedirs(os.path.dirname(dst_path), exist_ok=True)
```

BigRoy (commented):
I have my doubts about this - I really doubt whether triggering just a `save_workfile` should also generate the folders for it. It feels like it's trying to do more than what it should - I can't think of any host that actually generates the folders on save when calling their `save` methods. All DCCs and APIs seem pretty consistent in that.

@antirotor @philippe-ynput thoughts?

moonyuet:
> I have my doubts about this - I really doubt whether triggering just a save_workfile should also generate the folders for it. 
Not sure if it helps but the houdini docs tells how the `hou.hipFIle.Save` works, it should be working for either absolute or relative to the current directory. 
```
save(file_name=None, save_to_recent_files=True)

Saves the current scene to a .hip file.

file_name

If this is a string, saves the current scene to a file at this path, like File ▸ Save As. The path can be absolute or relative to the current directory.

If this is None, saves the scene to its current path ([hou.hipFile.path](https://www.sidefx.com/docs/houdini/hom/hou/hipFile.html#path)), like File ▸ Save.

If intermediate directories in the path do not exist, Houdini will create them.
```

MustafaJafar:
@moonyuet you are actually right. I've tested the my snippet above and it worked. So, this PR is not needed. 
Currently, I'm investigating to know why doesn't it work in my pdg setup.

MustafaJafar:
Okay, so the Houdini documentation is wrong. But this should be solved in the wokfile save.

PR #223, opened 2025-02-13
expand string before returning path in generic loaders

### LINKED ISSUE #222: referencing loaded sequences with Houdini generic loader fails

## Current Behavior
As mentioned in this comment https://github.com/ynput/ayon-houdini/issues/220#issuecomment-2657435423

> However.., I've just discovered a bad issue while making these visuals for you.. 
> 
> Although these paths are exactly the same but one of them is referencing a parm `chs("../../usd_assets_render_BlueCyan_Beauty/file")`

> Image

## Expected Behavior
It should work.

### Steps To Reproduce:

1. Load any sequence with generic loader
2. Reference file parm
3. the referenced parm doesn't evaluate the path at the current frame.

### Version

Windows
Linux - Rocky9
Linux - Ubuntu
MacOS

### Relevant log output:

```shell

```

MustafaJafar:
## Changelog Description
resolve #222 

## Testing notes:
1. Load any sequence with generic loader
2. Reference file parm
3. the referenced parm has the evaluated path at the current frame.

### DIFF
--- a/client/ayon_houdini/api/hda_utils.py
+++ b/client/ayon_houdini/api/hda_utils.py
@@ -885,11 +885,11 @@ def expression_get_representation_path() -> str:
     use_entity_uri = bool(hou.evalParm("use_ayon_entity_uri"))
     hash_value = project_name, repre_id, use_entity_uri
     if hash_value in cache:
-        return cache[hash_value]
+        return hou.text.expandString(cache[hash_value])
 
     path = get_representation_path(project_name, repre_id, use_entity_uri)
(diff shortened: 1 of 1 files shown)


BigRoy:
Interesting, it's indeed described in Houdini docs here:  https://www.sidefx.com/docs/houdini/hom/expressions.html#string-parameter-expressions

With it expanded, is it still 'live' to context options and alike?

BigRoy (approved):

BigRoy:
Works for me! Thanks!

PR #226, opened 2025-02-19
fix parameters names of lop import and load shot

### LINKED ISSUE #225: Parameters of lop import HDA have changed causing the logic to break

## Current Behavior
Some parameter names like `foler name` were changed maybe while working on #199 
Image

## Expected Behavior
.

### Steps To Reproduce:
1. load asset using lop import
2. change presets to `folder path`. It'll break.

### Version

Windows
Linux - Rocky9
Linux - Ubuntu
MacOS

### Relevant log output:

```shell

```

MustafaJafar:
## Changelog Description
resolve #225 

## Testing notes:
1. load usd products using lop import and load shot
2. change node configuration and everything should work normally.


BigRoy (approved):
Thanks - clean is better!

moonyuet (approved):
LGTM
image

BigRoy:
Unfortunately @MustafaJafar this PR changed more than we would have liked.

E.g. parm `primpath1` turned to `primpath` even though the parm has always been `primpath1` before. Due to this change/rename any existing scenes using the LOP Import (Load Asset LOP) now has lost any overrides from the defaults to the Primitive Root parm.

BigRoy:
I've taken a look at the git history, and the only parm that should've been changed here was `folder_name`.
I'd vote for making a new PR and a hotfix release to get back to the original behavior. So that only the short usage of the current release is what will be 'broken' instead. @MustafaJafar thoughts?

MustafaJafar:
> I've taken a look at the git history, and the only parm that should've been changed here was `folder_name`. I'd vote for making a new PR and a hotfix release to get back to the original behavior. So that only the short usage of the current release is what will be 'broken' instead. @MustafaJafar thoughts?

Let me give it a try.

PR #233, opened 2025-02-25
Load Asset/Shot LOP: revert parm name change except for the fixed `folder_name`

BigRoy:
## Changelog Description

Revert parm names back to prior #226 to align with earlier releases, only keep the fixed `folder_name` parm on LOP import change from that PR.

**Critical note** This would be useful for a hotfix release - so that scenes remain backwards compatible to loaded content from ayon-houdini 0.4.2 and below. Only 0.4.3 release would then be the outlier with the invalid parm names.

Without this any parm overrides, e.g. to Primitive Root `primpath1` on the LOP import (load asset LOP) would be lost in the existing scenes.

Fixes what's described here: https://github.com/ynput/ayon-houdini/pull/226#issuecomment-2682993899

## Testing notes:

1. Load Asset LOP and Load Shot LOP should still work as intended (including all their attributes, like Primitive Root, time offset, etc.)
2. Load Asset LOP and Load Shot LOP overrides (from ayon-houdini 0.4.2 and before) stored on these parms should be maintained.


MustafaJafar (approved):
I was able to reproduce the issue and this PR works. 
Thanks.

PR #234, opened 2025-03-03
Add Create Model (FBX) creator to create FBX with model product type

robertokker:
## Changelog Description
copied  the static mesh creator to make the productType "model" instead of staticMesh.
This so it aligns better with fbx exports from blender etc in the whole pipeline.

Also added 2 new options to the creator so you don;t have to go into the rop to change them al the time.
1-ASCII format (as blender can only read ASCII fbx files this is now a option in the creator .
2- single frame (this sets the fbx node to single frame or to frame range.


BigRoy:
> copied the static mesh creator to make the productType "model" instead of staticMesh.
> This so it aligns better with fbx exports from blender etc in the whole pipeline.

@robertokker If this code aligns 90% with the other one. Could we put the creators in the same python file and inherit one from the other, so that 99% of the code is shared?

I really hate duplicating stuff :D 

@antirotor thoughts?

robertokker:
@BigRoy  yeah i will try to put it into 1 file, and hust have the option to adjust the "productType" as i think that should work.
the only thing is it loads in the creator####.py so maybe rename than the staticMesh to FBX creator?

BigRoy:
> yeah i will try to put it into 1 file, and hust have the option to adjust the "productType" as i think that should work.
> the only thing is it loads in the creator####.py so maybe rename than the staticMesh to FBX creator?

Yes, exactly - refactor the filename to `create_fbx.py` but don't rename the old Creator class name of course, but just do something with inheritance.

It does show the potential issue with Creators being so tightly coupled to the _product type_ instead of a creator possibly being of multiple product types. Which has also come up a lot with "generic publishing" where we have some automated creator that can generate any type of product type. The system (as @iLLiCiTiT kindly pointed out) was never designed for a Creator to target more than one product type - they were very tightly coupled by design. But it popping up here and there may hint that at some point there could be reasons to avoid that tight coupling. Tagging @philippe-ynput @antirotor just in case.

robertokker:
OK i'll see the point indeed, makes a lot more sense and is a lot cleaner.

client/ayon_houdini/plugins/create/create_FBX.py
```diff
@@ -1,32 +1,34 @@
...
+
     def create(self, product_name, instance_data, pre_create_data):
 
         instance_data.update({"node_type": "filmboxfbx"})
         creator_attributes = instance_data.setdefault(
             "creator_attributes", dict())
         creator_attributes["render_target"] = pre_create_data["render_target"]
 
-        instance = super(CreateStaticMesh, self).create(
+        instance = super(CreateFBX, self).create(
```
BigRoy, line 31: This should be fine in Py3.
```suggestion
        instance = super().create(
```

client/ayon_houdini/plugins/create/create_FBX.py
```diff
@@ -104,9 +107,21 @@ def get_pre_create_attr_defs(self):
                                         "FBX unit of centimeters.",
                                 default=False,
                                 label="Convert Units")
-
+        format_ASCII = BoolDef("exportkind",
```
BigRoy, line 110: Variables should be lower cased.
```suggestion
        format_ascii = BoolDef("exportkind",
```

client/ayon_houdini/plugins/create/create_FBX.py
```diff
@@ -104,9 +107,21 @@ def get_pre_create_attr_defs(self):
...
+                              items={
+                                  0: "single frame",
+                                  1: "frame range"
+                              },
+                              default=0,
+                              label="single frame")
+        
         return attrs + [
-            createsubnetroot, vcformat, convert_units
+            createsubnetroot, vcformat, convert_units, format_ASCII, singleframe
```
BigRoy, line 124:
```suggestion
            createsubnetroot, vcformat, convert_units, format_ascii, singleframe
```

client/ayon_houdini/plugins/create/create_FBX.py
```diff
@@ -122,7 +137,7 @@ def get_dynamic_data(
         The default prodcut name templates for Unreal include {asset} and thus
         we should pass that along as dynamic data.
         """
-        dynamic_data = super(CreateStaticMesh, self).get_dynamic_data(
+        dynamic_data = super(CreateFBX, self).get_dynamic_data(
```
BigRoy, line 140:
```suggestion
        dynamic_data = super().get_dynamic_data(
```
Should be fine in Py3.

client/ayon_houdini/plugins/create/create_FBX.py
```diff
@@ -175,3 +190,15 @@ def get_selection(self):
             )
 
         return selection
+
```
BigRoy, line 193: Cosmetics (code style says we should have two line breaks between top level classes)
```suggestion


```

server/settings/create.py
```diff
@@ -9,6 +9,13 @@ class CreatorModel(BaseSettingsModel):
         default_factory=list,
     )
 
+class CreatorModelFBX(BaseSettingsModel):
+    enabled: bool = SettingsField(title="Enabled")
+    default_variants: list[str] = SettingsField(
+        title="Default Products",
+        default_factory=list,
+    )
+
```
BigRoy, line 18: Cosmetics
```suggestion

class CreatorModelFBX(BaseSettingsModel):
    enabled: bool = SettingsField(title="Enabled")
    default_variants: list[str] = SettingsField(
        title="Default Products",
        default_factory=list,
    )


```

client/ayon_houdini/plugins/create/create_FBX.py
```diff
@@ -1,32 +1,34 @@
 # -*- coding: utf-8 -*-
-"""Creator for Unreal Static Meshes."""
+"""Creator plugin for creating Model product type as FBX."""
```
BigRoy, line 2:
```suggestion
"""Creator plugins for FBX, like static mesh and model products"""
```

client/ayon_houdini/plugins/create/create_FBX.py
```diff
@@ -1,32 +1,34 @@
...
-    """Static Meshes as FBX. """
-
-    identifier = "io.openpype.creators.houdini.staticmesh.fbx"
-    label = "Static Mesh (FBX)"
-    product_type = "staticMesh"
-    icon = "fa5s.cubes"
+class CreateFBX(plugin.HoudiniCreator):
+    """Model as FBX. """
 
+    identifier = "io.openpype.creators.houdini.modelFBX.fbx"
```
BigRoy, line 12: No need for `fbx` twice in the identifier. Also, since this is a 'new' creator - let's make sure the identifier is also using the newer `ayon` instead of legacy `openpype` name.
```suggestion
    identifier = "io.ayon.creators.houdini.model.fbx"
```

client/ayon_houdini/plugins/create/create_FBX.py
```diff
@@ -175,3 +190,16 @@ def get_selection(self):
             )
 
         return selection
+
+
+class createStaticMesh(CreateFBX):
+    """Static Meshes as FBX. """
+
+    identifier = "io.openpype.creators.houdini.staticmesh.fbx"
+    label = "staticMesh (FBX)A"
```
BigRoy, line 199:
```suggestion
    label = "staticMesh (FBX)"
```

client/ayon_houdini/plugins/create/create_FBX.py
```diff
@@ -175,3 +190,16 @@ def get_selection(self):
...
+    product_type = "staticMesh"
+    icon = "cube"
```
BigRoy, line 201: What's the reason for the icon change?
```suggestion
    icon = "fa5s.cubes"
```

client/ayon_houdini/plugins/create/create_FBX.py
```diff
@@ -175,3 +190,16 @@ def get_selection(self):
...
+    default_variants = ["Main"]
+
+    # Default render target
+    render_target = "local"
```
BigRoy, line 205:
```suggestion

```
Anything you do not override does not need to be redeclared on the child class - so let's remove these.

client/ayon_houdini/plugins/create/create_FBX.py
```diff
@@ -175,3 +190,12 @@ def get_selection(self):
...
         return selection
+
+
+class createStaticMesh(CreateFBX):
+    """Static Meshes as FBX. """
+
+    identifier = "io.openpype.creators.houdini.staticmesh.fbx"
+    label = "staticMesh (FBX)"
+    product_type = "staticMesh"
+    icon = "cube"
```
BigRoy, line 201: This should override publish families again - because otherwise it'd still enforce `model` and `fbx` instead of static mesh since it inherits from the other class.
```suggestion
    icon = "cube"

    def get_publish_families(self):
        return ["fbx", "staticMesh"]
```

BigRoy:
@robertokker I've pushed some comments (and directly committed them too since they were mostly cosmetic). You can use the comments to see my reasoning, and hope there isn't any issue with me pushing them here.

Just left the icon change on static mesh creator open as a comment - any reason to not stick to the icon it had originally?

And had one comment about `get_publish_families` that would actually be a bug currently on the static mesh creator.

client/ayon_houdini/plugins/create/create_FBX.py
robertokker: @BigRoy the model abc also had cube as icon, so i think i copied it from there.

robertokker:
@BigRoy thanks for the commit and the help.

Tommorow i'll look at the model and pointcloud as they are also a full duplicate ;)

BigRoy:
Also, noticed it just now - can you make sure the `create_FBX.py` filename remains lowercase, `create_fbx.py`.

server/settings/create.py
```diff
@@ -9,6 +9,15 @@ class CreatorModel(BaseSettingsModel):
...
 
+
+class CreatorModelFBX(BaseSettingsModel):
+    enabled: bool = SettingsField(title="Enabled")
+    default_variants: list[str] = SettingsField(
+        title="Default Products",
+        default_factory=list,
+    )
+
+
```
BigRoy, line 20: In this case this extra settings model is not needed. It's exactly like the one above -and there isn't a reason for a new unique on here. Note also how the other settings entries below, multiples of them use the `CreatorModel`. It's like the base model for Creators, it's not specific to the model creator. ;)

```suggestion```

File: server/settings/create.py [hunk 14]
```diff
@@ -86,6 +95,9 @@ class CreatePluginsModel(BaseSettingsModel):
     CreateModel: CreatorModel = SettingsField(
         default_factory=CreatorModel,
         title="Create Model")
+    CreateModelFBX: CreatorModelFBX = SettingsField(
+        default_factory=CreatorModelFBX,
```
BigRoy, line 99:
```suggestion
    CreateModelFBX: CreatorModel= SettingsField(
        default_factory=CreatorModel,
```

client/ayon_houdini/plugins/create/create_fbx .py
```diff
@@ -175,3 +190,16 @@ def get_selection(self):
             )
 
         return selection
+
+
+class createStaticMesh(CreateFBX):
```
iLLiCiTiT, line 195:
```suggestion
class CreateStaticMesh(CreateFBX):
```

BigRoy: Well spotted!

iLLiCiTiT:
> The system (as @iLLiCiTiT kindly pointed out) was never designed for a Creator to target more than one product type - they were very tightly coupled by design.

I was, and still am, against the families logic. The creator and it's instances are not "very tightly" connected, there are assumptions in the codebase (only in UI if I remember correctly).  Technically if you don't care about "wrong UI" here and there, it might actually be possible. The creator is responsible for everything related to it's instances -> fill data (product type), fill product name etc. It's always matter of use case and more handling in the plugin, sometimes it's easier to have 2 plugins with single implementation, than 1 plugin that has to handle all the changes.

client/ayon_houdini/plugins/create/create_fbx .py
```diff
@@ -1,32 +1,34 @@
...
 
-class CreateStaticMesh(plugin.HoudiniCreator):
-    """Static Meshes as FBX. """
-
-    identifier = "io.openpype.creators.houdini.staticmesh.fbx"
-    label = "Static Mesh (FBX)"
-    product_type = "staticMesh"
-    icon = "fa5s.cubes"
+class CreateFBX(plugin.HoudiniCreator):
+    """Model as FBX. """
```
iLLiCiTiT, line 10:
```suggestion
    """Model as FBX."""
```

client/ayon_houdini/plugins/create/create_fbx .py
```diff
@@ -104,9 +107,21 @@ def get_pre_create_attr_defs(self):
...
+                               label="ASCII format")
+        singleframe = EnumDef("trange",
+                              tooltip="single frame or sequence",
+                              items={
+                                  0: "single frame",
+                                  1: "frame range"
+                              },
+                              default=0,
+                              label="single frame")
+
```
MustafaJafar, line 122:
```suggestion
        return attrs + [
            createsubnetroot, vcformat, convert_units,
            format_ascii, singleframe,
        ] + self.get_instance_attr_defs()
```

MustafaJafar (changes requested):
Hello, some notes to make ruff happy. 

Also, Could you remove the space in the file name `create_fbx .py`?

robertokker:
commited all changes.
Thnaks for the help

client/ayon_houdini/plugins/create/create_fbx.py
```diff
@@ -104,9 +107,20 @@ def get_pre_create_attr_defs(self):
...
+                              tooltip="single frame or sequence",
+                              items={
+                                  0: "single frame",
+                                  1: "frame range"
+                              },
+                              default=0,
+                              label="single frame")
         return attrs + [
-            createsubnetroot, vcformat, convert_units
+            createsubnetroot, vcformat, convert_units,format_ascii, singleframe,
```
iLLiCiTiT, line 123:
```suggestion
            createsubnetroot,
            vcformat,
            convert_units,
            format_ascii,
            singleframe,
```

client/ayon_houdini/plugins/create/create_fbx.py
```diff
@@ -104,9 +107,24 @@ def get_pre_create_attr_defs(self):
...
+                               default=True,
+                               label="ASCII format")
+        singleframe = EnumDef("trange",
+                              tooltip="single frame or sequence",
+                              items={
+                                  0: "single frame",
+                                  1: "frame range"
+                              },
+                              default=0,
+                              label="single frame")
```
MustafaJafar, line 121:
```suggestion
```

client/ayon_houdini/plugins/create/create_fbx.py
```diff
@@ -104,9 +107,24 @@ def get_pre_create_attr_defs(self):
...
         return attrs + [
-            createsubnetroot, vcformat, convert_units
+            createsubnetroot,
+            vcformat,
+            convert_units,
+            format_ascii,
+            singleframe,
```
MustafaJafar, line 127:
```suggestion
```

client/ayon_houdini/plugins/create/create_fbx.py
```diff
@@ -46,8 +48,9 @@ def create(self, product_name, instance_data, pre_create_data):
             "vcformat": pre_create_data.get("vcformat"),
             "convertunits": pre_create_data.get("convertunits"),
             # set render range to use frame range start-end frame
-            "trange": 1,
-            "createsubnetroot": pre_create_data.get("createsubnetroot")
+            "trange": pre_create_data.get("trange"),
```
MustafaJafar, line 51:
```suggestion
            "trange": 1,
```

MustafaJafar (commented):
So, it turns out that setting `trange` doesn't make sense. as static meshes and models both are supposed to be a single frame. 

Question: Why does it default to frame range?
Answer: It should be a frame range where `frame start` equals `frame end`. this results in publishing a single frame and it also tells us which frame was used for publishing.
Therefore, the result in this screenshot is discouraged. 
image


---
Fun fact we have two validators that do the same thing :D 
[ValidateMeshIsStatic](https://github.com/ynput/ayon-houdini/blob/develop/client/ayon_houdini/plugins/publish/validate_mesh_is_static.py) and [validate_export_is_a_single_frame.py](https://github.com/ynput/ayon-houdini/blob/develop/client/ayon_houdini/plugins/publish/validate_export_is_a_single_frame.py).

client/ayon_houdini/plugins/create/create_fbx.py
```diff
@@ -104,9 +107,24 @@ def get_pre_create_attr_defs(self):
...
+                               tooltip="ASCII format or not",
+                               default=True,
+                               label="ASCII format")
+        singleframe = EnumDef("trange",
+                              tooltip="single frame or sequence",
+                              items={
+                                  0: "single frame",
+                                  1: "frame range"
+                              },
+                              default=0,
```
BigRoy, line 120: Shouldn't this be `default=1` so it behaves like it did before?

client/ayon_houdini/plugins/create/create_fbx.py
```diff
@@ -104,9 +107,24 @@ def get_pre_create_attr_defs(self):
                                         "FBX unit of centimeters.",
                                 default=False,
                                 label="Convert Units")
-
+        format_ascii = BoolDef("exportkind",
+                               tooltip="ASCII format or not",
+                               default=True,
+                               label="ASCII format")
```
BigRoy, line 113: I think is better as `EnumDef` so it's clearer from the value labels what the different states mean. So the entries would be "ASCII" ,"Binary"

client/ayon_houdini/plugins/create/create_fbx.py
BigRoy: > So, it turns out that setting trange doesn't make sense. as static meshes and models both are supposed to be a single frame.

Maybe @MustafaJafar 's comment is more important here though.

MustafaJafar: Also, I think this setting can be confusing because it allows the artists to choose frame range and then they face this error as they can't publish a frame range. 
also, the model can't be animated. there are two validators that fails here [ValidateMeshIsStatic](https://github.com/ynput/ayon-houdini/blob/develop/client/ayon_houdini/plugins/publish/validate_mesh_is_static.py) and [validate_export_is_a_single_frame.py](https://github.com/ynput/ayon-houdini/blob/develop/client/ayon_houdini/plugins/publish/validate_export_is_a_single_frame.py).

image

client/ayon_houdini/plugins/create/create_fbx.py
robertokker: yeah agree would be cleaner, changing it right now

robertokker: but what would we label it then as it's a bit weird that the label is ASCII but you can have it in a dropdown binary?

Maybe "export format" @BigRoy

client/ayon_houdini/plugins/create/create_fbx.py
MustafaJafar: @robertokker this one is missed.

client/ayon_houdini/plugins/create/create_fbx.py
BigRoy: Yes, export format or file format.

robertokker: i kept it "Export Format" as that is in reference to the Houdini terminology

MustafaJafar (approved):
It works.

client/ayon_houdini/plugins/create/create_fbx.py
```diff
@@ -175,3 +187,16 @@ def get_selection(self):
             )
 
         return selection
+
+
+class CreateStaticMesh(CreateFBX):
+    """Static Meshes as FBX. """
+
+    identifier = "io.openpype.creators.houdini.staticmesh.fbx"
+    label = "staticMesh (FBX)"
```
BigRoy, line 196: Keep original label
```suggestion
    label = "Static Mesh (FBX)"
```

client/ayon_houdini/plugins/create/create_fbx.py
```diff
@@ -175,3 +187,16 @@ def get_selection(self):
             )
 
         return selection
+
+
+class CreateStaticMesh(CreateFBX):
+    """Static Meshes as FBX. """
```
BigRoy, line 193:
```suggestion
    """Static Meshes as FBX."""
```

BigRoy (approved):
LGTM

BigRoy:
Due to a lack of "labels" for enhancement or bugfix this PR never made it into the changelog for [0.4.5](https://github.com/ynput/ayon-houdini/releases/tag/0.4.5).

PR #235, opened 2025-03-04
Thumbnail should run before heavy processes so houdini does not crash

SzabadosA:
## Problem
We experienced segfault houdini crashes on thumbnail creation, when publishing products. This happens because heavy extractors (in our case the HDA extractor) still block the main thread at the moment when the process moves to the next extractor, which crashes the flipbook tool.

## Changes
We change the extractor order to 0.25 so the thumbnail executes before the product extractors, which mitigates the problem

### DIFF
--- a/client/ayon_houdini/plugins/publish/extract_active_view_thumbnail.py
+++ b/client/ayon_houdini/plugins/publish/extract_active_view_thumbnail.py
@@ -15,7 +15,7 @@ class ExtractActiveViewThumbnail(plugin.HoudiniExtractorPlugin,
     publishing as a fallback.
 
     """
-    order = pyblish.api.ExtractorOrder + 0.49
+    order = pyblish.api.ExtractorOrder + 0.25
     label = "Extract Active View Thumbnail"
     families = ["workfile"]
 


client/ayon_houdini/plugins/publish/extract_active_view_thumbnail.py
```diff
@@ -15,7 +15,7 @@ class ExtractActiveViewThumbnail(plugin.HoudiniExtractorPlugin,
     publishing as a fallback.
 
     """
-    order = pyblish.api.ExtractorOrder + 0.49
+    order = pyblish.api.ExtractorOrder + 0.25
```
BigRoy, line 18: Interesting to see this did fix it. Because I wonder what happens between +0.25 and +0.49.

I'd have expected this would needed to be negative like `- 0.1` so that it'd run before `ExtractROP` - assuming that the crash had something to do with the heavy processed occurring from rendering heavy rops in the scene.

Also this plug-in does this check for whether thumbnails are already set on the instance:

```
thumbnail = instance.data.get("thumbnailPath")
        if thumbnail:
            # A thumbnail was already set for this instance
            return
```

We should ensure it doesn't now suddenly run BEFORE anything that we'd expect to otherwise set the thumbnail path - otherwise this one would incorrectly always take precedence.

MustafaJafar:
@SzabadosA could you create an issue for it and link it to this PR?

client/ayon_houdini/plugins/publish/extract_active_view_thumbnail.py
MustafaJafar: `ExtractActiveViewThumbnail` is only affecting the `workfile` instance so we should consider only the plugins related to thumbnails and runs through `workfile`  instance. like
- [extract_thumbnail_from_source.py](https://github.com/ynput/ayon-core/blob/develop/client/ayon_core/plugins/publish/extract_thumbnail_from_source.py)

There's another plugin in core but not related to `workfile` instance. 
- [extract_thumbnail.py](https://github.com/ynput/ayon-core/blob/develop/client/ayon_core/plugins/publish/extract_thumbnail.py)

MustafaJafar: > In our case it failed on HDA extraction, therefor we moved it right in front of the HDA extractor. Do you have any other examples, where you could test this (sth. where you have the feeling the rop extractor fails)?

How many HDAs were in your scene?

client/ayon_houdini/plugins/publish/extract_active_view_thumbnail.py
```diff
@@ -15,7 +15,7 @@ class ExtractActiveViewThumbnail(plugin.HoudiniExtractorPlugin,
     publishing as a fallback.
 
     """
-    order = pyblish.api.ExtractorOrder + 0.49
+    order = pyblish.api.ExtractorOrder - 0.25
```
MustafaJafar, line 18: For information.
This order is smaller that the order of a plugin for extracting thumbnail and it runs over `workfile` instances.

https://github.com/ynput/ayon-core/blob/e420a97543d95bccb5fbf80d220d3ef7a4515670/client/ayon_core/plugins/publish/extract_thumbnail_from_source.py#L36C5-L36C49

Issue #237, opened 2025-03-10
AY-7587_Add "rig" product type to save APEX bgeo rigs


ynbot:
### Please describe the enhancement you have in mind and explain what the current shortcomings are?

APEX rigs are essentially SOP-level geometry data that can be written to bgeo  and loaded back in. Hence, it's possible to publish rigs in Houdini directly to bgeo as interchange format.

There is an existing pointcache (bgeo) that can publish this, however the "pointcache" product type on it is mostly confusing because it'd essentially need to be a rig.

It'd be good add an additional "rig" creator that writes out bgeo with the intent of being APEX rigging data.

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1348660662469005464)
_(might be a private channel)_

_This issue was automatically created from Clickup ticket [AY-7587](https://app.clickup.com/t/86c2gfpqh)_

PR #238, opened 2025-03-10
Add "rig" product type to publish APEX rigs to BGEO

BigRoy:
## Changelog Description

Add "rig" product type to publish APEX rigs to BGEO

Adds a "Rig" creator:

image

image

Fix https://github.com/ynput/ayon-houdini/issues/237

This required a workaround for validating the handles because the rigs aren't intended to be frame-range based. I'm not sure I _like_ the fix, but couldn't quickly think of anything nicer than this.

## Testing notes:

1. Create rig to publish any SOP level data, e.g. any APEX rig.
2. Note: There's no validation currently whether the data published is ACTUALLY an apex rig.
3. Publishes fine to bgeo.
4. The published data works fine loaded back in for animating with the APEX rig.

### DIFF
--- a/client/ayon_houdini/plugins/create/create_rig.py
+++ b/client/ayon_houdini/plugins/create/create_rig.py
@@ -1,0 +1,107 @@
+import inspect
+
+from ayon_houdini.api import plugin
+from ayon_core.lib import EnumDef
+from ayon_core.pipeline import CreatorError
+
+import hou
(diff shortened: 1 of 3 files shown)


moonyuet (approved):
Tried with not very complete skeleton(joint)
Looks good when publishing and loading the rig 
image

PR #242, opened 2025-03-13
Load Image: Fix loading render if there's no colorspace data

### LINKED ISSUE #241: loading render fails if there's no colorspace

## Current Behavior
.

## Expected Behavior
.

### Steps To Reproduce:

.

### Version

Windows
Linux - Rocky9
Linux - Ubuntu
MacOS

### Relevant log output:

```shell

```

MustafaJafar:
## Changelog Description
resolve #241
`get_colorspace_parms` should always return a dict. 

## Testing notes:
1. loading render without colorspace shouldn't error.


BigRoy (approved):
Good to go!

PR #244, opened 2025-03-24
Create USD: Do not lock `fileperframe` attribute

BigRoy:
## Changelog Description

Do not lock `fileperframe` attribute on creating USD instance.

Before this PR the attribute was locked:
image

However, it should be totally fine to enable that and there may be use cases where a user may want to do so - e.g. very memory intensive write operations - without needing to resort to USD value clips directly.

The attribute was locked since forever, e.g. it [came into `ayon-usd` initially](https://github.com/ynput/ayon-houdini/commit/7c71563ab7fb574bf41fde5a52ac6fc1c400e659#diff-37feab3f7a4a376c9bd21b742ce65466abec615d233d93eaa12ff71668ef182fR40) on the conversion from old addon structure and came once from the initial implementation in [OpenPype, coming from my custom colorbleed development at the time - 4 years ago](https://github.com/ynput/OpenPype/blame/c85bd30e1b01aa543f8a14a5b03de2a01f8fbcd7/openpype/hosts/houdini/plugins/create/create_usd.py#L34).

## Testing notes:

1. Publishing should work fine with the option disabled and enabled.


MustafaJafar (approved):
changes works as expected. 
publishing works. 
loading works.

PR #246, opened 2025-03-25
Manage: Add "Select in scene" inventory action

BigRoy:
## Changelog Description

Manage: Add "Select in scene" inventory action

image

## Testing notes:

1. Scene Inventory "Select in scene" action should work
2. It should also work for multiple selected containers, however only last will be marked "current"


MustafaJafar (approved):
This is cool! 
I love it. 
image

PR #247, opened 2025-03-25
Manage: Add "Show parameters" action

BigRoy:
## Changelog Description

Add "Show Parameters" action which allows to quickly show the node in a pop-up parameter window for a loaded product. Making it easy to quickly tweak a setting on a loaded object, especially useful for e.g. custom loaders using e.g. Generic Loaders internally.

image

Shows e.g.

image

## Testing notes:
1. Show Parameters action should work
2. Action should be useful


MustafaJafar (approved):
This is super cool!
I also like how it shows the geo obj node for other products. 

image

PR #251, opened 2025-04-08
Create Review: Fix setting `imageFormat` extension on create

BigRoy:
## Changelog Description

Create Review: Fix setting `imageFormat` extension on create

Previously the output path would always have been set to `png`. Now it should adhere to what user selected on create.

## Testing notes:

1. Creating and publishing should work as inteded.


MustafaJafar (approved):
It works.

Issue #179, opened 2024-11-20
Repair action on review colorspace validation returns an invalid value


MustafaJafar:
### Is there an existing issue for this?

- [x] I have searched the existing issues and added correct labels.

## Current Behavior
When disabling the color management toggle in Houdini addon settings, the `Reset OCIO colorspace parm` action on `Validate Review Colorspace` plugin returns an invalid value `<USE_DISPLAY_NAME>`. _that value maybe returned from [here](https://github.com/ynput/ayon-core/blob/a181fc897d16db9563fca73473eddee590cdd427/client/ayon_core/pipeline/colorspace.py#L1486) in the code._
![image](https://github.com/user-attachments/assets/0a1bca71-2baf-42cd-bb92-4969f02611c5)

## Expected Behavior
The repair action should return the correct colorspace from the current `display` and `view` values.

## Additional Info
The validation is working properly as I can pass the validation when choosing something manually from the drop down menu. (even if I disable the color management toggle in Houdini addon settings)
![image](https://github.com/user-attachments/assets/0bcfc966-db73-4434-b06d-085c170449c5)

### Steps To Reproduce:

1. Disabling the color management toggle in Houdini addon settings
2. Create a review instance in Houdini
3. You'll find the value of OCIO parm is set to `<USE_DISPLAY_NAME>`.

### Version
Version of the server, ayon-core and the addon that's affected

### What platform were you running when you found the bug?
Windows

You can also provide any additional information relevant to the issue.

### Relevant log output:

```shell

```

BigRoy:
See:
```python
>>> from ayon_core.pipeline.colorspace import _get_display_view_colorspace_name
>>> import os
>>> _get_display_view_colorspace_name(os.getenv("OCIO"), "sRGB - Display", "Un-tone-mapped")
'<USE_DISPLAY_NAME>'
```

Apparently that's a special return value from OCIO itself when [searching its codebase](https://github.com/search?q=repo%3AAcademySoftwareFoundation%2FOpenColorIO%20USE_DISPLAY_NAME&type=code) like [here](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/ddb9b120624541ca11734c9aa4024e7472ceac83/docs/configurations/ocio-v2_demo.ocio#L129)

Also, this issue doesn't only happen on repair - it also happens when Creating the review instance when OCIO is enabled.

PR #252, opened 2025-04-08
Create Review: Fix `<USE_DISPLAY_NAME>` value in color correction create and review

BigRoy:
## Changelog Description

Create Review: Fix `<USE_DISPLAY_NAME>` value in color correction create and review

Fix https://github.com/ynput/ayon-houdini/issues/179

## Testing notes:

1. Create review instance with OCIO management enabled should set correct Color Correction OCIO Colorspace value/
2. Validation should also run correctly.

### DIFF
--- a/client/ayon_houdini/api/colorspace.py
+++ b/client/ayon_houdini/api/colorspace.py
@@ -52,14 +52,17 @@ def get_scene_linear_colorspace():
     return colorspaces["roles"].get("scene_linear", {}).get("colorspace")
 
 
-def get_default_display_view_colorspace():
+def get_default_display_view_colorspace() -> str:
     """Returns the colorspace attribute of the default (display, view) pair.
 
     It's used for 'ociocolorspace' parm in OpenGL Node."""
 
     prefs = get_color_management_preferences()
-    return get_display_view_colorspace_name(
+    colorspace = get_display_view_colorspace_name(
         config_path=prefs["config"],
         display=prefs["display"],
         view=prefs["view"]
     )
+    if colorspace == "<USE_DISPLAY_NAME>":
+        colorspace = prefs["display"]
+    return colorspace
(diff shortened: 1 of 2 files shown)


MustafaJafar (commented):
Using `prefs["display"]` causes `Validate Review Colorspace` to fail.

image
The value set by the creator actaully doesn't exist in the `OCIO colorspace` drop down menu, so I guess the validator is working properly.
image
image

client/ayon_houdini/api/colorspace.py
```diff
@@ -52,14 +52,17 @@ def get_scene_linear_colorspace():
...
 
     prefs = get_color_management_preferences()
-    return get_display_view_colorspace_name(
+    colorspace = get_display_view_colorspace_name(
         config_path=prefs["config"],
         display=prefs["display"],
         view=prefs["view"]
     )
+    if colorspace == "<USE_DISPLAY_NAME>":
+        colorspace = prefs["display"]
```
MustafaJafar, line 67: `prefs["display"]` value doesn't exist in `hou.Color.ocio_spaces()` 

'ociocolorspace' parm only accepts color spaces not displays.

MustafaJafar: what happens when you leave the parm empty? 
I _think_ it'll apply linear conversion when using values that don't exist in the drop down menu (which is the same as leaving it empty).

MustafaJafar (approved):
It works on my side. 
I had to disable the colorspace management and validate review colorspace. 
Maybe we can add to the validator to not fall if the values is also a display. 

image

convert this 

https://github.com/ynput/ayon-houdini/blob/2d4ad82aa8ce0377f4ea923710e37756cbdbefb0/client/ayon_houdini/plugins/publish/validate_review_colorspace.py#L84

to 

```python
            if current_color_space not in (hou.Color.ocio_activeDisplays() + hou.Color.ocio_spaces()) :
```

MustafaJafar (approved):
Brilliant, I love it! 
image

PR #257, opened 2025-04-29
use `variant` instance data not `product name`

### LINKED ISSUE #256: Local Render, variant template key evaluates to product name.

## Current Behavior

Image

## Expected Behavior
It should be `variant` not product name. 

### Steps To Reproduce:

1. Create a render instance
2. Set render target to local
3. Check instances names

### Version

Windows
Linux - Rocky9
Linux - Ubuntu
MacOS

### Relevant log output:

```shell

```

MustafaJafar:
## Changelog Description
resolve #256 

image
for the above image, Template: `{product[type]}{Variant}<_{Aov}>`
for run time instances: 
- type is `render`
- variant is `Main`

## Testing notes:
1. Create a render instance
2. Set render target to local
3. Check instances names


BigRoy (approved):
LGTM

PR #277, opened 2025-06-27
Feature/geoclip copy publish

MahanthLBS:
## Changelog Description
Brief description of the solution this PR is implementing. This should be only a few sentences and it will be publicly visible in the changelog. You can highlight importance of the change and add links to further information or supporting documentation

I have collected the clip, manifest, and topology paths from the GeoClipSequence node, and while I have explored multiple solutions to copy them into the publish folder, any best solution for this. 

## Testing notes:
1. start with this step
2. follow this step


MahanthLBS:
@BigRoy

client/ayon_houdini/plugins/publish/collect_usd_layers.py
```diff
@@ -70,6 +70,39 @@ def process(self, instance):
...
+                clip_path = None
+            manifest_path = geoclip_node.evalParm("manifestfile")
+            topology_path = geoclip_node.evalParm("topologyfile")
+
+            self.log.debug(f"Found GeoClipSequence Node: {geoclip_node.path()}")
+            self.log.debug(f"Clip Path: {clip_path}")
+            self.log.debug(f"Manifest Path: {manifest_path}")
+            self.log.debug(f"Topology Path: {topology_path}")
+        else: 
+            self.log.debug("False not a geoclipsequence node")
```
BigRoy, line 105: Thanks for the contribution - even though this shows what _may need to be collected_ from a GeoClip this is not the approach we should be looking for unfortunately. This is _VERY SPECIFIC_ to the GeoClip node - and NOT to USD data in general. Instead we should be querying the USD scene and its state to know about the files.

The GeoClip node may not be there (it may have somehow been sublayered in, or come from an object merge from somewhere else, etc. etc. Or it could be inside a custom HDA.

We _should_ be querying the relevant files from the USD Stage or its Sdf Layers. It's something I can look into given priority and time - however so far that hasn't been there. :)

PR #278, opened 2025-06-27
USD Asset Creators: Add `Asset` to the USD asset creator labels

BigRoy:
## Changelog Description

USD Asset Creators: Add `Asset` to the USD asset creator labels

Our team sometimes missed that e.g. "assembly" referred to an asset assembly, and not the assembly/layout of e.g. a shot. As such, adding 'Asset' to the labels helped a lot to avoid that confusion.

However, it's a bit of personal taste perhaps - so please be opinionated in your review.

## Testing notes:

1. Check whether you like this or not.

### DIFF
--- a/client/ayon_houdini/plugins/create/create_usd.py
+++ b/client/ayon_houdini/plugins/create/create_usd.py
@@ -85,10 +85,10 @@ class CreateUSD(plugin.HoudiniCreator):
 
 class CreateUSDModel(CreateUSD):
     identifier = "io.ayon.creators.houdini.model.usd"
-    label = "USD Model"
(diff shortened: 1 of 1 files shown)


BigRoy:
> Does it worth it to udpate description of `USD` creator as well mentioning it's a generic type?

I'm not entirely convinced - I think with it saying "Create USD" I feel like it's somewhat telling that it's generic... I'm not sure if "Create generic USD" would really be better.

@antirotor @philippe-ynput can you please review on whether this is a change you like or not?

BigRoy:
Merged this - we can tweak more later if needed.

PR #284, opened 2025-07-08
Adapt to opmenu hierarchy of houdini `20.5`

### LINKED ISSUE #195: Node context menu has missing label, shows "(label)" instead

## Current Behavior
There seems to be a proper definition of a label missing in when you right click on a node. Now it says "(label)":
Image

## Expected Behavior
I'm assuming there should be a proper name for the label.

### Steps To Reproduce:

Create a node and press right click.

### Version

Windows
Linux - Rocky9
Linux - Ubuntu
MacOS

### Relevant log output:

```shell

```

MustafaJafar:
## Changelog Description
Adapt to opmenu hierarchy of houdini `20.5` as mentioned https://github.com/ynput/ayon-houdini/issues/195#issuecomment-3048146337

> Image

## Testing notes:
you should see the same result as in  https://github.com/ynput/ayon-houdini/issues/195#issuecomment-3048146337
In 19.5 you'll get a warning
```
Order error: 'opmenu.vhda_create' and 'opmenu.vhda_create_ayon' don't have the same parent.
Cannot set order: element 'root_menu' has no parent.
Cannot set order: element 'root_menu' has no parent.
Error while parsing a menu definition file
 'C:/Program Files/Side Effects Software/sidefx_packages/SideFXLabs19.5/OPmenu.xml':
Cannot set order: element 'root_menu' has no parent.
Cannot set order: element 'root_menu' has no parent.
```

### DIFF
--- a/client/ayon_houdini/startup/OPmenu.xml
+++ b/client/ayon_houdini/startup/OPmenu.xml
@@ -6,14 +6,14 @@
 <menuDocument>
     <menu>
         <!-- Operator type and asset options. -->
-        <subMenu id="opmenu.vhda_options_create">
-            <insertBefore>opmenu.unsynchronize</insertBefore>
-                <scriptItem id="opmenu.vhda_create_ayon">
-                    <insertAfter>opmenu.vhda_create</insertAfter>
-                    <label>Create New (AYON)...</label>
-                    <context>
-                    </context>
-                    <scriptCode>
+
+        <insertBefore>opmenu.unsynchronize</insertBefore>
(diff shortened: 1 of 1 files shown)


MustafaJafar:
This PR is merely for documenting this solution.

BigRoy (approved):
Actually, as your screenshots suggest - the new way looks better in both versions... plus it doesn't log warnings/errors on both versions + it's better forward compatible for future releases.

I'd say, happy got merge.

PR #287, opened 2025-07-21
Add Copernicus image file loader

BigRoy:
## Changelog Description

Add Copernicus image file loader

Note that it tries to load in an active Copernicus network, because it doesn't seem that Copernicus has the equivalent of an Object Merge node so there's little use in having the loaded content inside the `AYON_CONTAINERS` object.

Also added settings to allow disabling certain Loader plug-ins:

image

## Testing notes:

1. Loading should work
2. Update should work
3. Remove should work.


MustafaJafar (approved):
It works.

PR #301, opened 2025-09-10
Support Slap Comps on farm

### LINKED ISSUE #84: AY-6686_Houdini 20.5: Publish along "Slap Comp" renders from USD Render

The USD Render ROP in Houdini 20.5 allows you to directly also write out the "Slap Comp" images from the ROP node.

image

Also see: https://www.sidefx.com/docs/houdini/copernicus/slap_comp.html

It would be great if we could publish this as additional output.

Whether this should be a separate product or representation that's a question. E.g. someone spoke about using Slap Comp to generate a "QC" view of the render, which may make more sense to be a separate representation of the main render instead of a separate product (like an AOV).

_No response_

### Additional context:

_No response_

MustafaJafar:
## Changelog Description
resolve #84

## Testing notes:
This PR is tested along 2 other PRs: 
- https://github.com/ynput/ayon-deadline/pull/201
- https://github.com/BigRoy/HuskStandaloneSubmitter/pull/7

Testing Steps:
1. Create a USD render rop with slap comp enabled, you can use file or cop.  _Here's an example slap comp file for testing: [chromaticglows.zip](https://github.com/user-attachments/files/22260838/chromaticglows.zip)_
2. In publisher, Set render target to `Farm Rendering - Split export & render jobs` and publish.

> [!Note] 
> If your outputfile include `Beauty`, a review will be created from it.
> image

### DIFF
--- a/client/ayon_houdini/api/lib.py
+++ b/client/ayon_houdini/api/lib.py
@@ -1666,8 +1666,8 @@ def save_slapcomp_to_file(
         a relative or absolute path.
     """
 
-    sopenet = slapcomp_out.parent().createNode("sopnet")
-    copnet = sopenet.createNode("copnet")
+    sopnet = slapcomp_out.parent().createNode("sopnet")
+    copnet = sopnet.createNode("copnet")
     blocktogeo = copnet.createNode("blocktogeo")
     blocktogeo.parm("blockpath").set(slapcomp_out.path())
 
@@ -1685,4 +1685,4 @@ def save_slapcomp_to_file(
     copnet.geometry().saveToFile(filepath)
 
     # Clean up.
-    sopenet.destroy()
+    sopnet.destroy()
--- a/client/ayon_houdini/api/lib.py
+++ b/client/ayon_houdini/api/lib.py
@@ -1677,10 +1677,11 @@ def save_slapcomp_to_file(
         os.makedirs(os.path.dirname(filepath), exist_ok=True)
     except OSError as e:
         if e.errno != errno.EEXIST:
-            print(
+            raise RuntimeError(
                 f"Failed to create {os.path.dirname(filepath)}"
                 " dir. Maybe due to insufficient permissions."
             )
+        
 
     copnet.geometry().saveToFile(filepath)
 
--- a/client/ayon_houdini/api/lib.py
+++ b/client/ayon_houdini/api/lib.py
@@ -1651,3 +1651,38 @@ def format_as_collections(
     result = [collection.format(pattern) for collection in collections]
     result.extend(remainder)
     return result
+
+
+def save_slapcomp_to_file(
+        slapcomp_out: hou.CopNode,
+        filepath: str
+):
+    """Save slapcomp to file
+    
+    Args: 
+    slapcomp_out (hou.CopNode): A Block to Geometry Copernicus node
+        to export as geometry.
+    filepath (str): where to save the slapcomp. It can be 
+        a relative or absolute path.
+    """
+    
+    sopenet = slapcomp_out.parent().createNode("sopnet")
+    copnet = sopenet.createNode("copnet")
+    blocktogeo = copnet.createNode("blocktogeo")
+    blocktogeo.parm("blockpath").set(slapcomp_out.path())
+
+    # Create the destination folder as `saveToFile`
+    # doesn't create missing intermediate directories.
+    try:
+        os.makedirs(os.path.dirname(filepath), exist_ok=True)
+    except OSError as e:
+        if e.errno != errno.EEXIST:
+            print(
+                f"Failed to create {os.path.dirname(filepath)}"
+                " dir. Maybe due to insufficient permissions."
+            )
+
+    copnet.geometry().saveToFile(filepath)
+    
+    # Clean up.
+    sopenet.destroy()
--- a/client/ayon_houdini/api/lib.py
+++ b/client/ayon_houdini/api/lib.py
@@ -1658,14 +1658,14 @@ def save_slapcomp_to_file(
         filepath: str
 ):
     """Save slapcomp to file
-    
-    Args: 
+
+    Args:
     slapcomp_out (hou.CopNode): A Block to Geometry Copernicus node
         to export as geometry.
-    filepath (str): where to save the slapcomp. It can be 
+    filepath (str): where to save the slapcomp. It can be
         a relative or absolute path.
     """
-    
+
     sopenet = slapcomp_out.parent().createNode("sopnet")
     copnet = sopenet.createNode("copnet")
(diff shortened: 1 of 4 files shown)


client/ayon_houdini/plugins/publish/collect_slapcomps.py
```diff
@@ -0,0 +1,92 @@
...
+
+        ropnode = hou.node(node_path)
+
+        slapcomp_sources =  []
+
+        comp_numbers = ropnode.evalParm("husk_slapcomp")
+        for i in range(1, comp_numbers+1):
+        
+            if ropnode.evalParm(f'husk_sc_enable{i}'):
+
```
BigRoy, line 42: This could be less indented overall if you make this:


```suggestion
            if not ropnode.evalParm(f'husk_sc_enable{i}'):
                continue
```

And then unindent the below.

client/ayon_houdini/plugins/publish/collect_slapcomps.py
```diff
@@ -0,0 +1,93 @@
...
+        comp_numbers = ropnode.evalParm("husk_slapcomp")
+        for i in range(1, comp_numbers+1):
+        
+            if not ropnode.evalParm(f'husk_sc_enable{i}'):
+                continue
+
+            # slap comp cli expects a path to apex node.
+            if ropnode.evalParm(f'husk_sc_source{i}') != "file":
+                self.log.warning(
+                    "Slap Comp on farm only works with slap files!"
```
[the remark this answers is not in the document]
MustafaJafar: Sorry for my false feedback above.
it works in hou.

package.py
```diff
@@ -10,5 +10,5 @@
     "core": ">=1.0.8",
 }
 ayon_compatible_addons = {
-    "deadline": ">=0.5.11",
+    "deadline": ">=0.5.17",
```
MustafaJafar, line 13: @iLLiCiTiT This PR's deadline twin PR was merged and this addon version. 
I don't think the next dl addon release will get a minor bump therefore, I updated the patch version. 

let me know what do you think.

client/ayon_houdini/plugins/publish/collect_slapcomps.py
```diff
@@ -0,0 +1,121 @@
...
+
+    # This specific order value is used so that
+    # this plugin runs after CollectRenderProducts
+    order = pyblish.api.CollectorOrder + 0.1
+    label = "Collect Slap Comps"
+    families = ["usdrender"]
+
+    # Copernicus was introduced in Houdini 20.5 so we only enable this
+    # Collect Slap Comps if the Houdini version is 20.5 or higher.
+    enabled = hou.applicationVersion() >= (20, 5, 0)
```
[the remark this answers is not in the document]
BigRoy: I think it's fine. Slap comps aren't a thing prior to that houdini release and hence during publishing the user won't be missing out on anything.

client/ayon_houdini/api/lib.py
```diff
@@ -1651,3 +1651,38 @@ def format_as_collections(
...
+    # Create the destination folder as `saveToFile`
+    # doesn't create missing intermediate directories.
+    try:
+        os.makedirs(os.path.dirname(filepath), exist_ok=True)
+    except OSError as e:
+        if e.errno != errno.EEXIST:
+            print(
+                f"Failed to create {os.path.dirname(filepath)}"
+                " dir. Maybe due to insufficient permissions."
+            )
```
BigRoy, line 1683: Should this still continue to `raise`? or? Seems odd that this would just print yet still try to continue?

MustafaJafar: it's inside a try except.

BigRoy: Correct - which measn this except case just prints and does not re-raise.. and hence we're continuing even though the folder does not exist. Also `errno.EEXIST` prettys ure that shouldn't be raised if `exists_ok=True` as argument right?

client/ayon_houdini/plugins/publish/collect_slapcomps.py
```diff
@@ -0,0 +1,121 @@
...
+                continue
+
+            # slap comp cli expects a path to apex node.
+            slapcomp_src = ""
+
+            if ropnode.evalParm(f'husk_sc_source{i}') != "file":
+                slapcomp_out = ropnode.parm(f'husk_sc_cop{i}').evalAsNode()
+
+                slapcomp_dir = os.path.dirname(instance.data["files"][0])
+                slapcomp_src = f"{slapcomp_dir}/{slapcomp_out.name()}.bgeo"
```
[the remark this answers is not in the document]
BigRoy: Aren't we also passing the USD file path to husk? So we must know about it?

It should be [`instance.data["ifdFile"]`](https://github.com/ynput/ayon-deadline/blob/d0f4010b1732d603488dec0927e01630293a1047/client/ayon_deadline/plugins/publish/houdini/submit_houdini_render_deadline.py#L346).

client/ayon_houdini/api/lib.py
BigRoy: Let me rephrase.

Your current code - will not error. It will pass through, if the folder does not exist because it only prints in the except clause, it does not (re-)raise the error. Hence, currently your code allows continuing without errors (only a message will be printed) if the folder does not exist?

MustafaJafar: > Or am I misreading it?

No, I'm the one who is misreading it.

MustafaJafar: as far as I know, `exist_ok=True`  means do nothing if it already exists (no error)

client/ayon_houdini/plugins/publish/collect_slapcomps.py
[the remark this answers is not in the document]
MustafaJafar: Thank you. this should work
https://github.com/ynput/ayon-houdini/pull/301/commits/070f6e9b11af5ead001639fc0820b92c8a3f4239

client/ayon_houdini/api/lib.py
BigRoy: Correct - then you end up implementing a `try.. except` meaning that if it failed to create.. it'd end up going into the except clause... and then just continue onwards as if it's nothing because all the except clause is doing is print a message, it doesn't make it raise the error. You're lacking a `raise` statement?

MustafaJafar: oh, thank you. I've just realized it.

MustafaJafar:
@iLLiCiTiT Any further comments to address?

iLLiCiTiT:
> Any further comments to address?

No.

PR #302, opened 2025-09-18
Host: Implement application information method

iLLiCiTiT:
## Changelog Description
Implement `get_app_information` on host class.

Currently the method is used with https://github.com/ynput/ayon-core/pull/1445 to be stored into workfile entity.

## Testing notes:
1. Validate the implementation is actually correct.

### DIFF
--- a/client/ayon_houdini/api/pipeline.py
+++ b/client/ayon_houdini/api/pipeline.py
@@ -66,8 +66,19 @@ class HoudiniHost(HostBase, IWorkfileHost, ILoadHost, IPublishHost):
     def get_app_information(self):
         from ayon_core.host import ApplicationInformation
 
+        app_name = "Houdini"
+        hou_name = hou.applicationName()
+        if hou_name == "houdinicore":
+            app_name = "Houdini Core"
+
+        elif hou_name == "houdinifx":
+            app_name = "Houdini FX"
+
+        elif hou_name == "happrentice":
+            app_name = "Houdini Apprentice"
+
         return ApplicationInformation(
-            app_name="Houdini",
+            app_name=app_name,
             app_version=hou.applicationVersionString(),
         )
 
--- a/client/ayon_houdini/api/pipeline.py
+++ b/client/ayon_houdini/api/pipeline.py
@@ -63,6 +63,14 @@ class HoudiniHost(HostBase, IWorkfileHost, ILoadHost, IPublishHost):
         self._op_events = {}
         self._has_been_setup = False
 
+    def get_app_information(self):
+        from ayon_core.host import ApplicationInformation
+
+        return ApplicationInformation(
+            app_name="Houdini",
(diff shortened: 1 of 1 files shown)


client/ayon_houdini/api/pipeline.py
```diff
@@ -63,6 +63,14 @@ def __init__(self):
         self._op_events = {}
         self._has_been_setup = False
 
+    def get_app_information(self):
+        from ayon_core.host import ApplicationInformation
+
+        return ApplicationInformation(
+            app_name="Houdini",
```
BigRoy, line 70: This can be Houdini Core, Houdini FX, Houdini Indie which may be nice to know as well. I wonder if that differentiator should be in the name or the version string, but it can be nice to see that something came from e.g. Houdini Core instead of FX.

@MustafaJafar thoughts?

BigRoy: In Houdini core:
```python
>>> hou.applicationName()
'houdinicore'
```
In Houdini fx:
```python
>>> hou.applicationName()
'houdinifx'
```
In Houdini apprentice (non-commercial) - _not sure if AYON even works in this_:
```python
>>> hou.applicationName()
'happrentice'
```

MustafaJafar: Another thought why not use the application variant name from application addon?

MustafaJafar: another thought, I wonder if it possible to get these details via api.
image

MustafaJafar: Another thought I wonder if it possible to get the details via API and save only the first line as dcc name.
image

MustafaJafar (commented):
For reference, this PR works.
image

client/ayon_houdini/api/pipeline.py
MustafaJafar: okay, I've found it using hscript command [version](https://www.sidefx.com/docs/houdini/commands/version.html). 
This one works out of the box wihtout any conditions.
```python
hou.hscript("version -n")[0].strip()
>>> 'Houdini FX Version 20.5.370'
```
I'm accessing it via python which returns a tuple and added strip as it has `\n` at the end.
```python
>>> hou.hscript("version -n")
('Houdini FX Version 20.5.370\n', '')
```

iLLiCiTiT: Wait, I still don't understand what is wrong with current implementation?

iLLiCiTiT: I honestly don't think it is necessary to do this. For me split by `Version` is worse option than explicitly using deterministic appname in api. Also I still don't understand why is the license important? It will be shown in UI, it does not have much real usage value.

BigRoy (approved):
LGTM

PR #309, opened 2025-11-04
Support publishing usd geo clip sequences

### LINKED ISSUE #259: AY-7743_Support publishing USD Value Clips through Geometry Clip Sequence

When writing USD value clips through Geometry Clip Sequence LOPs these files are currently not detected nor collected with the publish as additional output files and remapped accordingly.

Currently only the Topology layer is added as an extra sublayer product because it has an explicit layer save path set, so the "Explicit layers" logic kicks in and generates a child product of the published USD product. However, we'd also need to handle the manifest.usd and individual clip.usd frames and handle the remapping on publish to the correct relative paths.

The question down the line would be whether each geometry clip sequence would need to turn into its own product with the manifest.usd , topology.usd  and the individual frames clip.#.usd  or whether they should become resources of the initial product. At the very least the sensible approach seems to at least allow manifest, topology and clips to live together in a single USD product instead of as separate ones.

_No response_

_No response_

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/945246349333520397/1364164589599723563)

_This issue was automatically created from Clickup ticket [AY-7743](https://app.clickup.com/t/86c3a3mr7)_

MustafaJafar:
## Changelog Description
Add support for publishing usd geo clip sequences along publishing usd products.

The behaviour of this publishing usd geo clip sequences is debatable. 
So far, there are two opinions on the table:
1. Include the geo clips as resources.
2. Include geo clips as separate products.
    - product1: clip sequence
    - product 2: manifest
    - product 3: topology

## Additional review information 2
This PR is made on simple fact (deduced from trying different things) that `HoudiniLayerInfo` prim include nodes that influence the USD file export (including `geoclipsequence`). 
And, it can only include one node of  `geoclipsequence`.. Any newer node overrides previous ones.

image

## Testing notes:
1. Create a scene with Geometry Clip Sequence.
2. Try different configurations for the geoclips node.
    image
3. Try adding an intermediate USD layer via configure layer (Currently this doesn't work).
4. Publish
5. Load product, it should work as expected.

Resolve #259

### DIFF
--- a/server/settings/publish.py
+++ b/server/settings/publish.py
@@ -240,7 +240,7 @@ DEFAULT_HOUDINI_PUBLISH_SETTINGS = {
     },
     "ValidateSingleFrame": {
         "enabled": True,
-        "optional": True,
+        "optional": False,
         "active": True
     },
     "ValidateReviewColorspace": {


client/ayon_houdini/plugins/publish/collect_usd_value_clips.py
```diff
@@ -0,0 +1,94 @@
...
+            if node.type().name() == "geoclipsequence":
+                clip_node = node
+                break
+
+        if clip_node is None:
+            return
+
+        files = []
+        files.append(clip_node.evalParm('manifestfile'))
+        files.append(clip_node.evalParm('topologyfile'))
```
BigRoy: You should be able to trace actually which layer added this opinion, because it'd be in that layer's layer stack and hence you should be able to trace back to which 'layer' the value clips exist.

client/ayon_houdini/plugins/publish/collect_usd_value_clips.py
```diff
@@ -0,0 +1,94 @@
...
+        prim = stage.GetPrimAtPath("/HoudiniLayerInfo")
+
+        editor_nodes = prim.GetCustomDataByKey("HoudiniEditorNodes")
+
+        # Houdini LOPs only allows one geoclip.
+        # any newer geoclip overrides the previous one.
+        clip_node = None
+        for node_id in editor_nodes:
+            node = hou.nodeBySessionId(node_id)
+            if node.type().name() == "geoclipsequence":
```
BigRoy: Does this still detect the correct way if the node is inside an HDA? (e.g. part of some custom tooling) I wonder if it then still points to the internal node or not?

client/ayon_houdini/plugins/publish/collect_usd_value_clips.py
```diff
@@ -0,0 +1,94 @@
...
+        stage = instance.data.get("stage", None)
+        if stage is None:
+            return
+
+        prim = stage.GetPrimAtPath("/HoudiniLayerInfo")
+
+        editor_nodes = prim.GetCustomDataByKey("HoudiniEditorNodes")
+
+        # Houdini LOPs only allows one geoclip.
+        # any newer geoclip overrides the previous one.
```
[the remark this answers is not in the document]
BigRoy: @MustafaJafar can do it fine on my end.

It seems you can do a Merge after them just fine and both will be written.

Or if you want to do both, connect the upstream stage also to the LEFT input because that's what defines the INPUT stage... the RIGHT one is the one you make a value clip from.

As such, as soon as you leave the left stream dangling you're essentially saying "do not include the input stage".
Also, more likely you'd have them in the stream by the LEFT inputs, and then the RIGHT inputs would solely be adding the animation in those cases - right?

client/ayon_houdini/plugins/publish/collect_usd_value_clips.py
BigRoy: Cool - I guess reliable enough for now then.

client/ayon_houdini/plugins/publish/collect_usd_value_clips.py
BigRoy: > If geoclips node is above a layer break, it won't be detected by the logic above. which matches the ROP node behavior.
> if geoclips is above a layer break, it won't save clip files.

Interesting, since the _stage_ WILL include the nodes above the layer break. But I guess HoudiniLayerInfo does exclude it then?

MustafaJafar: > HoudiniLayerInfo does exclude it then?

yes.
image

client/ayon_houdini/plugins/publish/collect_usd_value_clips.py
```diff
@@ -0,0 +1,109 @@
...
+    label = "Collect USD Value Clips"
+    # Run after core plugin `CollectResourcesPath`
+    order = pyblish.api.CollectorOrder + 0.496
+    families = ["usd"]
+
+    def process(self, instance):
+        # For each layer in the output layer stack process any USD Value Clip
+        # nodes that are listed as 'editor nodes' in that graph.
+        for layer in instance.data.get("layers", []):
+            self._get_layer_value_clips(layer, instance)
```
MustafaJafar: I think this should be changed but I ran out of ideas. for reference, `"layers"` key doesn't include the root layer. 

What I'd like to do here is: 
Firstly, detect to which the value clips are added. are they on the root layer or an intermediate layer created by configure layer. 
Secondly, add the files to the respective publish instance. I know for sure it's either the main publish instance or one of the runtime instances (assuming the collector for generating runtime instance is working properly).

The thing is I'm not sure how to pick the relevant layer as checking if "geoclipsequence" nodes exist in editor node of the layer alone is not enough.

e.g. the relevant layer can be implicit so, the resources should be added to the main instance.
image

**OR** It can be explicit so the runtime instance that includes that path should be used.
image

This snippet only works if the the nodes were added to explicit layer.
Just dropping it as a food for thought.
```suggestion
        layers = instance.data.get("runTimeInstances", {})

        for layer in layers:
            layer_instance = layers.get(layer) or instance
            self.log.debug(f"{layer.identifier}, {layer_instance}")
            self._get_layer_value_clips(layer, layer_instance)
```

Also, here's an example file 
[Expr_usdvalueclips_generic_v001.zip](https://github.com/user-attachments/files/23405560/Expr_usdvalueclips_generic_v001.zip)

server/settings/publish.py
```diff
@@ -235,6 +238,11 @@ class PublishPluginsModel(BaseSettingsModel):
         "optional": True,
         "active": True
     },
+    "ValidateSingleFrame": {
+        "enabled": True,
+        "optional": True,
```
BigRoy: Default should be false here because this validator runs on instances that should _really_ be static by default.
```suggestion
        "optional": False,
```

Issue #292, opened 2025-07-23
AY-8028_Collect USD Look Assets fails if asset property is 'animated' and has no default value.


ynbot:
### Please describe the issue you have encountered?

The Collect USD Look Assets fails on 'NoneType' object has no attribute 'path'  asset: Sdf.AssetPath = spec.default  resolves to None  likely due to animated values instead.

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1395349924820029531)
_(might be a private channel)_

_This issue was automatically created from Clickup ticket [AY-8028](https://app.clickup.com/t/86c4j6n6f)_

PR #311, opened 2025-11-08
Collect USD Look: Avoid erroring hard on attribute specs without `.default` value.

BigRoy:
## Changelog Description

Avoid erroring hard on attribute specs without `.default` value.

Fix https://github.com/ynput/ayon-houdini/issues/292

## Testing notes:

1. Attribute specs without authored `.default` should now not fully break collecting of look assets.

### DIFF
--- a/client/ayon_houdini/plugins/publish/collect_usd_value_clips.py
+++ b/client/ayon_houdini/plugins/publish/collect_usd_value_clips.py
@@ -8,6 +8,69 @@ import pyblish.api
 from ayon_houdini.api import plugin
 
 
+def get_clip_frames_in_frame_range(
+        clip_start: int,
+        clip_end: int,
+        has_end_set: bool,
+        loop: bool,
+        range_start: int,
+        range_end: int):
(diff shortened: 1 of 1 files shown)


MustafaJafar (approved):
It works on my side. I've tested publishing an existing asset without the name conflict and another one with the name conflict, both work.

Also, loading the products back works. 
One note though: 
Although, the underscore prefix is hardcoded in our codebase however, it works seamlessly with Houdini's default! In screenshots below, you can see the folder name in folder path starts with a number and Houdini automatically uses `_` prefix. tbh, I'm not sure if we may need any fixes in Houdini addon. Also, making the prefix customizable will result in deviating from Houdini's default. (Not sure about other applications like maya though)


image
image
image

just one fix and it'll be good to go 
https://github.com/ynput/ayon-core/blob/3fcb4949f2a8f9de340153520d278c761f1d9ba9/client/ayon_core/plugins/publish/extract_usd_layer_contributions.py#L813

PR #312, opened 2025-11-10
USD contribution workflow: prefix folder names with `_` if starting with a digit to avoid invalid prim names in USD data

MustafaJafar:
## Changelog Description

Follow up PR to https://github.com/ynput/ayon-core/pull/1535

## Testing notes:
1. Create a usd rop, the folder name should be prefixed with `_` if folder name starts with number.
2. Remove the value of default prim, trigger validation, repair action should add default prim expr that leads to the same value from previous point.

### DIFF
--- a/client/ayon_houdini/plugins/create/create_usd.py
+++ b/client/ayon_houdini/plugins/create/create_usd.py
@@ -3,6 +3,7 @@
 import inspect
 
 from ayon_houdini.api import plugin
+from ayon_houdini.api.usd import DEFAULT_PRIM_EXPRESSION
 from ayon_core.lib import EnumDef
 
 import hou
@@ -96,10 +97,7 @@ class CreateUSDModel(CreateUSD):
     additional_parameters = {
         # Set the 'default prim' by default to the folder name being
         # published to
-        "defaultprim": (
-            '/`ifs(strmatch("[0-9]", strsplit(chs("folderPath"), "/", -1)[0])'
-            ', "_", "")``strsplit(chs("folderPath"), "/", -1)`'
-        ),
+        "defaultprim": DEFAULT_PRIM_EXPRESSION,
     }
 
     def get_detail_description(self):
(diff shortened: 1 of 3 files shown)


client/ayon_houdini/plugins/create/create_usd.py
```diff
@@ -96,7 +96,10 @@ class CreateUSDModel(CreateUSD):
     additional_parameters = {
         # Set the 'default prim' by default to the folder name being
         # published to
-        "defaultprim": '/`strsplit(chs("folderPath"), "/", -1)`',
+        "defaultprim": (
+            '/`ifs(strmatch("[0-9]", strsplit(chs("folderPath"), "/", -1)[0])'
+            ', "_", "")``strsplit(chs("folderPath"), "/", -1)`'
```
BigRoy: With the complexity of this expression I start having the urge to make this a constant somewhere so it's maintained in one location. :)

BigRoy (changes requested):
Unfortunately the expressions don't seem to work @MustafaJafar 

It seems that `strmatch` is not a regex pattern, but just a simpler pattern matching and the `[0-9]` does not seem to work - according to these [match](https://www.sidefx.com/docs/houdini/vex/functions/match.html) docs it supports just a list of characters, so it does match if we put `[0123456789]` 

Also, I wonder if a multistring hscript expression starts making more sense here?

This works:
```hscript
{
string asset = strsplit(chs("folderPath"), "/", -1);
string prefix = ifs(strmatch("[0123456789]", asset[0]), "_", "");
return prefix + asset;
}
```

Or this single line one:
```
`ifs(strmatch("[0123456789]", strsplit(chs("folderPath"), "/", -1)[0]), "_", "")``strsplit(chs("folderPath"), "/", -1)`
```

BigRoy (approved):
Not the nicest expression - but it works. @MustafaJafar can you confirm and merge if ok for you?

PR #316, opened 2025-11-13
Update context vars: report path when failing to create it

BigRoy:
## Changelog Description

In case of failure to create folder for context specific houdini-variables report the filepath which failed to get created.

Makes it easier to debug the issue.

## Testing notes:

1. Check code changes


MustafaJafar (approved):
Noice!
 
```
Failed to create $MANGA dir at 'I:/tmp/manga'. Maybe due to insufficient permissions.
Updated $MANGA to I:/tmp/manga
```

BTW, My test above is using a storage drive that doesn't exist.

PR #317, opened 2025-11-18
Show dl job options by default for all creators

MustafaJafar:
## Changelog Description
- Show dl job options by default for all creators
- Add support for render target in composite creators
- Refactor setting the value of export parameter logic in the extractor.

DL options are added by [collect_jobinfo.py](https://github.com/ynput/ayon-deadline/blob/develop/client/ayon_deadline/plugins/publish/global/collect_jobinfo.py). 
Publisher list the options from this plugin if instances have the proper family on refresh time. 
Since cache instances are using intermediate family `publish.hou` that added on run time, they will always be missing the proper family on refresh time. 

This PR is doing the opposite, it adds the `publish.hou` family on creators, and then removes it if it's not needed during publishing.

## Testing notes:
1. DL Job options should be visible on all instances
2. Render locally and on farm should work as before

Resolve https://github.com/ynput/ayon-deadline/issues/173

### DIFF
--- a/client/ayon_houdini/plugins/publish/validate_export.py
+++ b/client/ayon_houdini/plugins/publish/validate_export.py
@@ -37,41 +37,24 @@ class ValidateExportParameterValue(plugin.HoudiniInstancePlugin):
     label = "Validate Export Toggle"
     actions = [FixParameterAction, SelectInvalidAction]
 
+    # Per ROP node type, define the export parameter and expected values
+    # for export and no export modes. The first entry is for export mode,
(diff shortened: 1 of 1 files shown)


MustafaJafar:
Btw, I did some experimenting moving the render targets to a collector instead of the creators (which allows registering on_value_change callback) while the code become much cleaner but we miss setting the export parameter value properly in render creators.
Also, it moves the render target enum to publish attributes column. which may not align with render target option in other addons.

Issue #320, opened 2025-11-21
YN-0232: Houdini USD LOP HDAs should support latest and hero versions


ynbot:
### Additional context:

_This issue was automatically created from AYON ticket [YN-0232](https://support.ayon.app/projects/Active_Clients/overview)_

PR #321, opened 2025-11-21
Support 'hero' and 'latest' in AYON HDAs version parameters

BigRoy:
## Changelog Description

Support 'hero' and 'latest' in AYON HDAs version parameters.

Also, if `use_entity_uri` is enabled on the node do not provide the "resolved" filepath but set the filepath using the actual constructed AYON Entity URI of the inputs parameters, so that e.g. 'latest' does not turn into a path with version=1 but remains version=latest in the path parameter.

It now tries to go much more through the resolve endpoint to ensure feature set remains closer to what backend can resolve instead of what custom logic we implemented here.

Fix https://github.com/ynput/ayon-houdini/issues/320

## Testing notes:
1. Using "hero" and "latest" as values for 'version' should now work as expected.
2. The "pick representation" dropdown should also work when using "hero" or "latest" entry.
3. The entity URIs should actually work with the resolver.
4. If wrong values are entered in the parameters then it should still be somewhat clear why the node may be failing (e.g. warnings on the node should represent it or alike.)

### DIFF
--- a/client/ayon_houdini/api/hda_utils.py
+++ b/client/ayon_houdini/api/hda_utils.py
@@ -134,7 +134,7 @@ def get_available_versions(
         include_hero (bool): Whether to include hero version if it exists.
 
     Returns:
-        list[str | int]: Version label with version names for the product
+        list[str | int]: Available version names for the product
     """
 
     project_name = node.evalParm("project_name") or get_current_project_name()


MustafaJafar (commented):
This is cool! 
image
image
image

---

image

One small request. could we have latest and hero listed in versions.

https://github.com/ynput/ayon-houdini/blob/6d9fea959979a23b774bab1736b65e4b1b42bbd5/client/ayon_houdini/api/hda_utils.py#L132-L139

MustafaJafar (approved):
This is awesome! 
image
image

PR #324, opened 2025-11-28
Render: Fix OCIO display/view data not being collected for local render instances

BigRoy:
## Changelog Description

Render: Fix OCIO display/view data not being collected for local render instances

Fixes: https://github.com/ynput/ayon-core/pull/1541#issuecomment-3589795683

## Testing notes:

1. Publishing with display/view transcoding empty should work for local render instances.


MustafaJafar (approved):
works!

PR #330, opened 2025-12-03
Render: generate correct target folder for frames to fix in Extract Last Publish 

BigRoy:
## Changelog Description

Render: generate correct target folder for frames to fix in Extract Last Publish 

Source issue:
> The Frames to Fix feature may fail to copy the last published files for renders with "expected files" data collected because it doesn't collect the correct 'staging dir' (where it actually wants to take the output directory) ... and if the folder does not exist where the files will actually end up in the end then it'll error because it'll try to `os.makedirs` for the wrong path and hence it can't copy the files to the non-existing destination.

Fixes [`YN-0283`](https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=67e0ea70d08011f0ad9759a68b000877)

Fix: https://github.com/ynput/ayon-houdini/issues/331

## Testing notes:

1. Extract Last Published should correctly copy frames from last version if frames to fix is used, but the expected output folder for the render does not exist yet.

### DIFF
--- a/client/ayon_houdini/api/lib.py
+++ b/client/ayon_houdini/api/lib.py
@@ -390,7 +390,7 @@ def imprint(node, data, update=False, folder="Extra", prefix=""):
     node.setParmTemplateGroup(parm_group)
 
 
-def lsattr(attr, value=None, root="/"):
+def lsattr(attr, value=None, root="/", recurse_in_locked_nodes=True):
     """Return nodes that have `attr`
      When `value` is not None it will only return nodes matching that value
      for the given attribute.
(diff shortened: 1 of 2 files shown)


MustafaJafar (commented):
It works locally.
here I fixed frames `1003-1005`.

https://github.com/user-attachments/assets/3152d078-db77-46ec-b177-031df4dbe587

---

If I recall correctly this feature only works locally, 
for farm we should use custom frames and tell it what to do with the missing frames.
So, we may remove redundant code related to farm/expected files.
image

MustafaJafar:
> But at least this PR fixes the issue at hand as a hotfix? If so, can you approve it. Or is there still a pending issue now?

let me give it one more test run.

PR #340, opened 2025-12-20
Add Version Info (Author, Date, Status) fields to generic loader

vincentullmann:
## Changelog Description
This PR adds some additional info parameters on the generic loader to show the versions author, updated date and status.

The same changes will have to be copied to the SOP and LOP variants of the node. I've not done that yet as I wanted to wait for a first round of feedback / discussion about the change and implementation.

## Testing notes:
1. load a product via a generic loader
2. confirm that the fields show the correct data
3. change the version and confirm the parameters update as expected

## Demo:
https://github.com/user-attachments/assets/9ac1f448-09c6-4f4e-b3ad-4f9b39392d2e

### DIFF
--- a/client/ayon_houdini/api/hda_utils.py
+++ b/client/ayon_houdini/api/hda_utils.py
@@ -1,6 +1,7 @@
 """Helper functions for load HDA"""
 
 import datetime
+import json
 import os
 import re
 import uuid
@@ -1248,5 +1249,42 @@ def expression_get_representation_path() -> str:
     cache[hash_value] = path
     return hou.text.expandString(path)
 
+
+# endregion
+
+
+# region Parm Buttons
+
+def show_info(
+    info: dict,
+    title: str = "Info",
+    text: str = "",
+) -> None:
+    """Show info in a dialog."""
+    info_text = json.dumps(info, indent=4)
+
+    button = hou.ui.displayMessage(
+        title=title,
+        text=text,
+        buttons=('Copy', 'Close'),
+        details_label="details_label",
+        details=info_text,
+        details_expanded=True,
+        close_choice=1,
+    )
+    if button == 0:  # copy
+        hou.ui.copyTextToClipboard(info_text)
+
+
+def show_version_info(node: hou.OpNode) -> None:
+    """Show version detail for the given node in a dialog."""
+    info = get_version_info(node)
+    show_info(info, title="Version Detail")
+
+
+def show_representation_info(node: hou.OpNode) -> None:
+    """Show representation detail for the given node in a dialog."""
+    info = get_representation_info(node)
+    show_info(info, title="Representation Detail")
 
 # endregion
--- a/client/ayon_houdini/startup/otls/ayon.generic_loader.hda/ayon_8_8Object_1generic__loader_8_81.0/DialogScript
+++ b/client/ayon_houdini/startup/otls/ayon.generic_loader.hda/ayon_8_8Object_1generic__loader_8_81.0/DialogScript
@@ -221,12 +221,16 @@
                 label   "Version ID"
                 type    string
                 default { [ "hou.phm().get_version_info().get(\"id\") or \"\"" python ] }
+                parmtag { "script_action" "from ayon_houdini.api import hda_utils\n\nhda_utils.show_version_info(kwargs[\"node\"])" }
+                parmtag { "script_action_icon" "BUTTONS_info" }
             }
             parm {
                 name    "representation"
                 label   "Representation ID"
                 type    string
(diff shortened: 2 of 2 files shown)


client/ayon_houdini/api/hda_utils.py
```diff
@@ -506,13 +508,85 @@ def _resolve_entity_uri(entity_uri: str, resolve_roots: bool = False):
...
+        dict: Version entity.
+
+    """
+    version_id = get_version_id(
+        project_name,
+        folder_path,
+        product_name,
+        version,
+    )
+    return get_version_by_id(project_name, version_id)
```
BigRoy: These are best reversed, or the latter should not call the first. Now you're querying the database twice.

But the other way you don't with:
```python
get_version_id(...):
   get_version(...)["versionId"]
```

client/ayon_houdini/api/hda_utils.py
```diff
@@ -997,33 +1100,105 @@ def expression_clear_cache(subkey=None) -> bool:
     return False
 
 
-def expression_get_representation_id() -> str:
```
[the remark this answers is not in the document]
BigRoy: Agreed. Refactor all and cleanup what's not needed anywhere.

client/ayon_houdini/startup/otls/ayon.generic_loader.hda/ayon_8_8Object_1generic__loader_8_81.0/DialogScript
```diff
@@ -248,6 +242,70 @@
...
+                parmtag { "script_callback_language" "python" }
+            }
+            parm {
+                name    "info_comment"
+                label   "Comment"
+                type    string
+                default { [ "hou.phm().get_version_info().get(\"attrib\", {}).get(\"comment\") or \"\"" python ] }
+                parmtag { "script_callback_language" "python" }
+            }
+        }
```
BigRoy: Should we prefix the info parms with `AYON_`, e.g. `AYON_info_comment`? @MustafaJafar thoughts?

client/ayon_houdini/api/hda_utils.py
BigRoy: Ah nevermind - that's wrong because the resolve endpoint doesn't return the actual entity, just the relevant ids 🤔

client/ayon_houdini/startup/otls/ayon.generic_loader.hda/ayon_8_8Object_1generic__loader_8_81.0/DialogScript
```diff
@@ -137,6 +137,41 @@
...
+        parm {
+            name    "nodes_referencing_file"
+            label   "Nodes Referencing File"
+            type    oplist
+            default { [ "\" \".join(parm.node().path() for parm in hou.parm('./file').parmsReferencingThis())" python ] }
+            parmtag { "oprelative" "/" }
+            parmtag { "script_callback_language" "python" }
+        }
+    }
```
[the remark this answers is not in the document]
BigRoy: > I've got a couple of other minor suggestions regarding the ui layout which are related to this.
> 
> Since this PR is introducing more "info" parameters I figured I'd move the "settings" into their own subfolder. I'm happy to split those changes off into their own PR so we can discuss them separate.
> 
> quick preview of my local version:
> 
>  Screencast_2025-12-22T17.42.07.mp4

Nice - I quite like it.

I'd love for the `Nodes Referencing File` to still be visible without needing to expand it - so that it's right there directly.
I'm fine with e.g. the top Info section being expanded by default, which I think is the `group_default` tag set to `1` on the parm?

---

I'm not too sure about the `Info > Info` headers nesting. I'd rather have them be distinctly named. Some ideas:
```
Info > Metadata
Info > Entity Data
```

Also, I think we should nudge the `Loader` section in the `Info` bar to the bottom. It's the least informative section, or at least the one the majority would not look at.

So I'm thinking:
```
Info
   Usage
   Metadata
   Loader
```

---

I quite like the Descriptive Parm too. May be best left to a follow up PR instead of flooding this one.

client/ayon_houdini/startup/otls/ayon.generic_loader.hda/ayon_8_8Object_1generic__loader_8_81.0/DialogScript
MustafaJafar: The reason for the `AYON_` is because of the possible conflict on ROP nodes. but this HDA won't need it since it's an AYON HDA anyways.

PR #343, opened 2026-01-06
Add support for publishing pointcloud using PRT_ROPDriver node.

### LINKED ISSUE #341: AY-7912: PRT publisher Houdini/3ds
### Additional context:
link to discussion on [Discord](https://discordapp.com/channels/517362899170230292/1381597077745696859)

_This issue was automatically created from AYON ticket [AY-7912]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=9a9727677ea011f08f013003c82d0673)

MustafaJafar:
## Changelog Description
Add support for publishing `pointcloud` using PRT_ROPDriver node.
image

## Additional Info
This PR integrates [houdini_PRTROP](https://github.com/flipswitchingmonkey/houdini_PRTROP)
Here's a prebuilt `PRT_RopDriver.dll` for Houdini 21.0.559
[PRT_RopDriver.zip](https://github.com/user-attachments/files/24442686/PRT_RopDriver.zip)

## Testing notes:
0. Firstly ensure you have `PRT_ROPDriver` node by having the `.dll` in `dso` folder within your houdini path.
1. Once installed, you should find `Point CacheCloud(PRT)` creator plugin in creator tool. _if the addon is not installed. the create will be disabled._
2. Publishing should work for different render targets.
3. Once published, you should find a pointcloud product with `prt` representation.
image

Resolve #341

### DIFF
--- a/client/ayon_houdini/plugins/create/create_pointcloud_prt.py
+++ b/client/ayon_houdini/plugins/create/create_pointcloud_prt.py
@@ -13,7 +13,7 @@ class CreatePRTPointCloud(plugin.HoudiniCreator):
     It requires the PRT_ROPDriver to be installed, which can be found at:
         https://github.com/flipswitchingmonkey/houdini_PRTROP
     """
-    identifier = "io.openpype.creators.houdini.pointcloud.prt"
+    identifier = "io.ayon.creators.houdini.pointcloud.prt"
     label = "PointCloud (PRT)"
     product_type = "pointcloud"
     product_base_type = "pointcloud"


moonyuet (approved):
Published the pointcache successfully in PRT but as 3dsmax has its own product type for PRT publishing workflow `pointcloud`, so if we would like to load houdini publish from 3dsmax, we need to include this PR https://github.com/ynput/ayon-3dsmax/pull/84 or we need to add pointcloud as the product type of this creator.
 
image

client/ayon_houdini/plugins/create/create_pointcloud_prt.py
```diff
@@ -0,0 +1,144 @@
...
+    description = "Create PRT ROP to export pointcloud data"
+
+    # Enable if `PRT_ROPDriver` type exists.
+    enabled = hou.ropNodeTypeCategory().nodeType("PRT_ROPDriver") is not None
+
+    # Default render target
+    render_target = "local"
+
+    def get_publish_families(self):
+        return ["pointcloud", "pointcache", "prt", "publish.hou"]
```
BigRoy: Should `pointcache` still be included here, and why?

BigRoy (changes requested):
See previous comments - needs to be added to settings and it's unclear why this also need to be tagged `pointcache` family

client/ayon_houdini/plugins/create/create_pointcloud_prt.py
MustafaJafar: it shares the same publish flow as `pointcache` and the output originates from SOP/ same input for publishing pointcaches.

BigRoy: Hmm - so now `pointcache` validators suddenly also target a supposedly `pointcloud` product type. I'm not too keen on having this so mixed. Can you add a clear comment to this code here as to why `pointcache` is included?

```suggestion
        # The pointcache family is included because all of its validators
        # also apply to the pointcache family
        return ["pointcloud", "pointcache", "prt", "publish.hou"]
```

**Or** we should be more explicit and add `pointcloud` on each of them to clearly mark that it also applies to this one. Maybe more explicit is better here.

Thoughts?

MustafaJafar:
@BigRoy I've updated the PR name, description as well 
Added a description for the creator in settings so that admins can now what is this without needing to launch Houdini.

client/ayon_houdini/plugins/create/create_pointcloud_prt.py
```diff
@@ -0,0 +1,146 @@
+
+
+# -*- coding: utf-8 -*-
```
BigRoy:
```suggestion
# -*- coding: utf-8 -*-
```

client/ayon_houdini/plugins/create/create_pointcloud_prt.py
```diff
@@ -0,0 +1,146 @@
...
+    the PRT representation. It requires the PRT_ROPDriver to be installed,
+    which can be found at: 
+        https://github.com/flipswitchingmonkey/houdini_PRTROP
+    """
+    identifier = "io.openpype.creators.houdini.pointcloud.prt"
+    label = "PointCloud (PRT)"
+    product_type = "pointcloud"
+    product_base_type = "pointcloud"
+    icon = "cubes"
+    description = __doc__
```
BigRoy: I feel like you're mixing up things here - this one is best fit for `detailed_description` and it'd need to get `inspect.cleandoc` on it too (except for Py3.13+ because there the compiler dedents `__doc__` values directly, but it's fine triggering `inspect.cleandoc` another time)

This currently takes up more space than the design was intended for:
Image

So do something along the lines of:
```python
description = "PRT pointcloud creator"

def get_detail_description(self) -> str:
    return inspect.cleandoc(self.__doc__)
```

BigRoy (changes requested):
Looking good - some small notes. Should be good after.

client/ayon_houdini/plugins/create/create_pointcloud_prt.py
BigRoy: I'd say no. `utf-8` is the default for Python as far as I know. Also see: https://stackoverflow.com/a/14083123

So:
```suggestion
```

client/ayon_houdini/plugins/create/create_pointcloud_prt.py
```diff
@@ -0,0 +1,147 @@
...
+
+import hou
+
+
+class CreatePRTPointCloud(plugin.HoudiniCreator):
+    """PRT pointcloud Creator
+
+    Create point cloud instances for publishing with
+    the PRT representation. It requires the PRT_ROPDriver to be installed,
+    which can be found at:
```
BigRoy: Nitpick.
```suggestion
    Create point cloud instances for publishing with the PRT representation.
    It requires the PRT_ROPDriver to be installed, which can be found at:
```

BigRoy (approved):
Didn't test the actual publish because I don't have the plug-in.
But code looks good - ~~just the identifier should use `ayon`~~ ;) _committed_ ✅ 

There is a bit of an 'ugly' side effect of the URL in the detail description. It's unreadable, and clicking it doesn't actually open the browser to it.

Image

It's fine for now but we should probably fix in core @iLLiCiTiT ?

iLLiCiTiT:
> It's fine for now but we should probably fix in core 

I've tried. Colors of url are handled by code, so we would have to implement the paint of QDocument on our own to be able to change it. But create an issue.

PR #350, opened 2026-01-22
Product names: Update product name functionality

iLLiCiTiT:
## Changelog Description
Use new signature of `get_product_name` function and update signature to use new `product_type` argument where needed.

The signature of `get_product_name` did change in past and is reaching end of life thus should not be used anymore. With PR https://github.com/ynput/ayon-core/pull/1656 it is also added `product_type` to `get_product_name` and `get_dynamic_data` methodes on create plugins.

Remove `get_dynamic_data` that were adding `"asset"` key, which is long deprecated key from OpenPype and `folder[name]` is already available in default ayon-core at this moment. There is one `get_dynamic_data` which actually returns folder data but cannot be removed because ayon-core's implementation does not add `"label"` to the folder template, created PR to fix it https://github.com/ynput/ayon-core/pull/1655 .

Workfile creator was for some reason using `get_dynamic_data` which is not implemented and returns empty dictionary.

## Testing notes:
1. USD workflow does work and is using entities set on the instance instead of context entities to get product name template.
2. FBX and HDA can be created and published.

### DIFF
--- a/client/ayon_houdini/plugins/create/create_fbx.py
+++ b/client/ayon_houdini/plugins/create/create_fbx.py
@@ -122,30 +122,6 @@ class CreateFBX(plugin.HoudiniCreator):
             format_ascii,
         ] + self.get_instance_attr_defs()
 
-    def get_dynamic_data(
-        self,
-        project_name,
-        folder_entity,
-        task_entity,
-        variant,
-        host_name,
-        instance
-    ):
-        """
-        The default product name templates for Unreal include {asset} and thus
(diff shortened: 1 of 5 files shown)


MustafaJafar (approved):
This PR works on my  side.
I'll leave this https://github.com/ynput/ayon-houdini/pull/350#discussion_r2775055527 for @BigRoy .

BigRoy (approved):
So much ugly backwards compatible code - but ugh, we need to 😄... so LGTM.

PR #358, opened 2026-02-06
`CollectLocalRenderInstances`: use `create_instances_for_aov`

### LINKED ISSUE #356: Collect Local Render Instances: Avoid usage of `_get_legacy_product_name_and_group` and `get_product_name_and_group_from_template`

## Current Behavior

The code here is using `_get_legacy_product_name_and_group` and `get_product_name_and_group_from_template` but should not, see: 
https://github.com/ynput/ayon-houdini/blob/7aa2d781ff6dd64bc2547e0abb5953f8225ee197/client/ayon_houdini/plugins/publish/collect_local_render_instances.py#L227-L233

For more context also see; https://github.com/ynput/ayon-core/pull/1675/changes/BASE..5b31c190590b2cd07e43d92d13e9c11d8f90e9ef#r2759320315 and 

## Expected Behavior

Preferably see if we can get by with a matching implementation to local render by using a similar Collect Local Render Instance as currently brought to Maya, 3dsmax and Blender.

- https://github.com/ynput/ayon-maya/pull/383
- https://github.com/ynput/ayon-3dsmax/pull/99
- https://github.com/ynput/ayon-blender/pull/217

I think the 3dsmax one is the best reference currently because it's most recent.

### Steps To Reproduce:

We're using some deprecated functions or functions that shouldn't really be needed.

### Version

Windows
Linux - Rocky9
Linux - Ubuntu
MacOS

### Relevant log output:

```shell

```

MustafaJafar:
## Changelog Description
- CollectLocalRenderInstances: prepare instance data first then create aov instances second

Stop relying on private methods like `_get_legacy_product_name_and_group`
This PR also follow reusing same functions used with deadline and adopted with local render in max and maya.

Personally, I'm not keen on using `ayon_core.pipeline.farm.pyblish_functions.create_skeleton_instance` for local render in Houdini as it can pollute the instance data adding irrelevant and data items that are not used in Houdini addon. which can make it hard to debug in the future. 
_I'm not against using it. and, I'm going to give it a try._

## Testing notes:
1. local render should work as before.


MustafaJafar (commented):
.

MustafaJafar (commented):
chose wrong line number again.

client/ayon_houdini/plugins/publish/collect_local_render_instances.py
```diff
@@ -1,23 +1,13 @@
...
-from ayon_core.pipeline.farm.pyblish_functions import (
-    get_product_name_and_group_from_template,
-)
 from ayon_core.pipeline.publish import (
     get_plugin_settings,
     apply_plugin_settings_automatically,
     ColormanagedPyblishPluginMixin
 )
+from ayon_core.pipeline.farm.pyblish_functions import (
+    create_instances_for_aov
```
MustafaJafar:
```suggestion
    create_skeleton_instance,
    create_instances_for_aov
```

client/ayon_houdini/plugins/publish/collect_local_render_instances.py
```diff
@@ -85,166 +75,94 @@ def process(self, instance):
...
+        # may represent scene display/view, etc.
+        for key in (
+                "colorspaceConfig",
+                "colorspace",
+                "colorspaceDisplay",
+                "colorspaceView",
+        ):
+            if key in instance.data:
+                value: str = instance.data[key]
+                instance_skeleton_data[key] = value
```
MustafaJafar: The following code works/has the same result. I can see products published fine with the expected data (I only looked at frame attributes). 
I don't have a strong opinion in this regards. So, feel free to let me know what you prefer.
```suggestion
        # Use same logic as how instances get created for farm submissions
        instance_skeleton_data =  create_skeleton_instance(
            instance,
            # TODO: These should be fixed in core to just allow the default
            #  None to work
            families_transfer=[],
            instance_transfer={},
        )
        # Remove frame data like frameStart and handleStart
        # as they are added in later publisher plugins
        instance_skeleton_data = {
            k: v 
            for k, v in instance_skeleton_data.items()
            if k not in {
                "frameStart",
                "frameEnd",
                "handleStart",
                "handleEnd",
            }
        }
        instance_skeleton_data["families"].append("render.local.hou")
        instance_skeleton_data.update({
            "frameStartHandle": instance.data["frameStartHandle"],
            "frameEndHandle": instance.data["frameEndHandle"],
            "instance_node": instance.data["instance_node"],
            "creator_attributes": instance.data["creator_attributes"],
            "publish_attributes": instance.data["publish_attributes"],
        })
```

MustafaJafar (commented):
Here's it. this should work with `create_skeleton_instance`.

BigRoy (changes requested):
Please use as much as possible from core - so that it uses the _exact same logic_. We are looking to generalize and unify so that we don't need to duplicate data.

Does it generate too much data on the instance? Possibly. But that may just mean that either in core - we're just redundantly adding too much data. But I'd rather have a key too much here - than have every DCC differ in the data they contain. The end goal is that as much as possible remains the same, even to such an extent that perhaps we'd need almost no changes per integration.

As explained in the issue https://github.com/ynput/ayon-houdini/issues/356 my preference is that we just take e.g. the Collector from 3dsmax PR, 1:1 and make a comment of the differences that are actually needed.

Then when we have it working between Maya, Max, Houdini and Blender I'd compare all of them - and see where they differ and why...  and start generalizing as much as possible. This may mean that down the line we'd just have core `CollectLocalRender` plug-in one could e.g. inherit from or alike where each integration would only add minimal changes.

The issue we're currently facing is that all integrations are essentially doing their own thing. But preferably, it'd take <15 minutes to reliably add support for local rendering to an integration. Once we have this generalization I'd start rolling out to more integrations where it's currently not supported yet - etc.

In short, if you can use the core `create_skeleton_data` and it works, please do so. I'm fine with putting some slight logic directly after where we comment `# houdini-specific` or alike, etc.

client/ayon_houdini/plugins/publish/collect_local_render_instances.py
BigRoy: I think the suggestion is better - because it unifies more with the other integrations.

MustafaJafar:
I think this is the closer I can get to collect local render plugin from other Addons. 
Let me outline my findings.
### Standard Local Render Collector
_We didn't agree on a standard yet. but, let's consider these (https://github.com/ynput/ayon-maya/pull/383, https://github.com/ynput/ayon-3dsmax/pull/99 and https://github.com/ynput/ayon-blender/pull/217) as standard._
1. Create skeleton(base) instance data using `create_skeleton_instance`
2. Create instance data for each AOV using `create_instances_for_aov`
3. Add non-picklable instance data to AOV instances data.
4. for each representation within each AOV instance data fill in roots in `stagingDir`
5. Create AOV instance
6. Add necessary families e.g. `review` family or if the addon needs a specific family for render instances like `render.local.hou` in Houdini.
7. Skip integrating original instance.

### Houdini local reneder collector
0. $${\color{orange}Houdin \space Specific: \space we \space have \space configurable \space aov \space filters \space in \space Houdini \space settings.}$$
1. Create skeleton(base) instance data using `create_skeleton_instance`
2. $${\color{purple}Add \space pickable \space data \space to \space skeleton(base) \space instance \space data \space e.g. \space families, \space instance\\_node, \space creator\\_attributes, \space publish\\_attributes}$$
3. $${\color{orange}Houdinig \space Specific: \space Remove \space frame \space data}$$
4. Create instance data for each AOV using `create_instances_for_aov`
5. Add non-picklable instance data to AOV instances data.
6. for each representation within each AOV instance data fill in roots in `stagingDir`
7. Create AOV instance
8. $${\color{orange}set\\_representation\\_colorspace \space per \space representation \space}$$ this to preserve the existing logic before this PR. also, I checked `create_skeleton_instance` and `create_instances_for_aov` they don't do it.
9. ~~Add necessary families e.g. `review` family or if the addon needs a specific family for render instances like `render.local.hou` in Houdini.~~
10. Skip integrating original instance.

I can remove purple point if you want but the code might set a false thought that families must be added per aov instance at the end although it can be done once at the beginning. 
On the other hand, I can't do something to orange points.

@BigRoy Let me know your thoughts.

client/ayon_houdini/plugins/publish/collect_local_render_instances.py
```diff
@@ -85,166 +79,72 @@ def process(self, instance):
...
+        )
+        # Remove frame data like frameStart and handleStart
+        # as they are added in later publisher plugins
+        instance_skeleton_data.pop("frameStart")
+        instance_skeleton_data.pop("frameEnd")
+        instance_skeleton_data.pop("handleStart")
+        instance_skeleton_data.pop("handleEnd")
+
+        # `create_skeleton_instance` only adds `render` and `review`
+        # Houdini local render also needs `ender.local.hou`
```
BigRoy:
```suggestion
        # Houdini local render also needs `render.local.hou`
```

MustafaJafar:
> So that e.g. a DCC would solely need to e.g. (overly simplified)
> 
> ```python
> expected_files = instance.data.setdefault('expectedFiles", {})
> expected_files["beauty"] = ["/path/to/1", "/path/to/2"]
> ```

This is the current solution in Houdini.

---

I believe `RenderProduct` is something we want for a long time.
But I think it can turn into an EPIC as we have other products as well not just render products. e.g. pointcache product and etc.
Also, I wonder would it be more accurate to say Product Version ? anyways, I think this discussion should be in a dedicated place.

client/ayon_houdini/plugins/publish/collect_local_render_instances.py
```diff
@@ -85,166 +88,87 @@ def process(self, instance):
...
         instance.data["integrate"] = False
 
-    def _get_product_name_and_group(
-        self, instance, product_base_type, dynamic_data
-    ):
-        """Get product name and group
+    def post_process_skeleton_data(
+            instance: pyblish.api.Instance,
+            instance_skeleton_data: dict
+        ):
```
moonyuet, line 152:
```suggestion
    def post_process_skeleton_data(
            self,
            instance: pyblish.api.Instance,
            instance_skeleton_data: dict
        ):
```
or 
```suggestion
    @staticmethod
    def post_process_skeleton_data(
            instance: pyblish.api.Instance,
            instance_skeleton_data: dict
        ):
```

moonyuet (changes requested):
See comment above as the latest commit errored out the collector:
```
  File "C:\Users\Kayla\AppData\Local\Ynput\AYON\dependency_packages\ayon_2603121036_windows.zip\dependencies\pyblish\plugin.py", line 528, in __explicit_process
    runner(*args)
  File "D:\ayon-addon_template\ayon-houdini\client\ayon_houdini\plugins\publish\collect_local_render_instances.py", line 100, in process
    self.post_process_skeleton_data(instance, instance_skeleton_data)
TypeError: CollectLocalRenderInstances.post_process_skeleton_data() takes 2 positional arguments but 3 were given
```

MustafaJafar:
I gave this PR some test runs. And, the result are the usual before the PR.

moonyuet (approved):
lgtm

MustafaJafar:
@BigRoy Any further comments before merging this?

PR #371, opened 2026-03-11
Contribute APEX Rig publish to USD layer

### LINKED ISSUE #351: YN-0439: Houdini: Contribute a rig publish to USD layer directly

On publishing a rig product type we want to publish some primvar to a rig department layer inside a USD file that points to the published rig product, just like an AYON USD Contribution.

It would essentially write out a USD representation that would get layered into a USD asset. The USD representation would describe the root level primvar pointing to the rig. This could, if the representations live next to each other even point to the bgeo representation using a relative asset path like

```
@./rig.bgeo.sc
``` 

### Additional context:

_This issue was automatically created from AYON ticket [YN-0439]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=95cafcc0fab811f08818a1184b131b72)

MustafaJafar:
## Changelog Description
This PR adds a usd representation to the rig product and contribute it to the usdAsset.
Also, defines a standard way for loading rigs, initializing the scene and contribute animation cache to USD.

- [x] Fix bugs and add missing pieces for APEX Rig publish (render target & validate sop)
- [x] Add USD contribution publish options to APEX Rig Publish
- [x] Add an early collector that creates a usd layer, add it as another representation to the rig product, contribute it to the usdAsset. 
- [x] Check if contribution workflow picks it up
- [x] Add more metadata to the rig usd layer that points to bgeo rig  representation. this is for implementing automation.
  image
- [ ] Add automation that auto loads the rig associated with the rig usd layer and add characters and props to scene. https://github.com/ynput/ayon-houdini/pull/384

## Additional Info
This PR https://github.com/ynput/ayon-houdini/pull/376 adds support publishing animation data.

## Testing notes:
1. Create an APEX rig and a publish instance for it.
2. USD contribution options should be present in the publish tab, already initialized with rig department layer.


client/ayon_houdini/plugins/create/create_bgeo.py
```diff
@@ -120,3 +124,30 @@ def get_network_categories(self):
...
+
+    def get_detail_description(self):
+        return inspect.cleandoc(
+            """Write a BGEO output file as `rig` product type. This can be
+            used to publish APEX rigs which are in essence just SOP-level
+            data representing a rig structure."""
+        )
+
+    def get_publish_families(self):
+        return ["rig", "apex", "bgeo", "publish.hou",  "usd"]
```
BigRoy: What's the `apex` family needed for aside of the `rig` family? Can't we just target `rig` family?

client/ayon_houdini/plugins/publish/collect_apex_usd.py
```diff
@@ -0,0 +1,34 @@
...
+
+        # Save the file
+        staging_dir = instance.data.get("stagingDir")
+        filename = f"{instance.name}.usd"
+        filepath = os.path.join(staging_dir, filename)
+
+        self.log.debug(f"Saving rig layer: {filepath}")
+        sdf_layer.Export(filepath, args={"format": "usda"})
+
+        instance.data["rig_layer"] = filepath
```
moonyuet: I am a bit confused about this. 
Why are you creating and saving rig layer at the collector, Shouldn't this be either you create some temp layers and save in the temp directory when you are creating instances, or you create anonymous layer during the context and publishing that layer as usda in extractor?
And also why the filename extension is `.usd` instead of `.usda` while you are exporting with format in `.usda`?

MustafaJafar: > Why are you creating and saving rig layer at the collector, Shouldn't this be either you create some temp layers and save in the temp directory when you are creating instances, or you create anonymous layer during the context and publishing that layer as usda in extractor?

When I checked our code, I found out we either `Sdf.Layer.CreateAnonymous()` or `Sdf.Layer.OpenAsAnonymous(layer_path)`. 

What I know from the execution flow, that the layer needs to exist early enough for the `ExtractUSDAssetContribution` plugin picks it.

But, I can't fill in the essential data to the layer on creation because I need some instance data that are not collected yet. 

Therefore, I created an empty layer early, and saved its path so that I can reopen again at the correct order.
Once opened I can add the require data to it.


> why the filename extension is `.usd` instead of `.usda` while you are exporting with format in `.usda`?

I adopted the same logic in 
 https://github.com/ynput/ayon-core/blob/dd0601bdf5cb4c8ffb678388978049423eb355f7/client/ayon_core/plugins/publish/extract_usd_layer_contributions.py#L814

moonyuet: I can see why you are exporting usda while the representation data of filename is ended with usd. But the logic here you adopted is from the extractor, right? I dont think it's safe adding and exporting layer in the collector but it should be instead exporting in extractor and the addition of layer should be lasted in the creating instance?

MustafaJafar: Hey, 
in this commit https://github.com/ynput/ayon-houdini/pull/371/commits/5ce2907754a11ee7baa330238dc987f03fbd4a2a  the layer is only saved on extraction.

client/ayon_houdini/plugins/publish/extract_apex_usd.py
```diff
@@ -0,0 +1,104 @@
...
+            representation_name="bgeo",
+            ext=ext,
+            version=version_name if specific_version else None
+        )
+        # Ensure `None` for now is also a string
+        path = str(path)
+        path = os.path.normpath(path)
+        if platform.system().lower() == "windows":
+            path = path.replace("\\", "/")
+        return path
```
MustafaJafar: For reference, this function returns a full path and it won't work if someone is using multi-os in their studio.

here's an example 
```
    uniform string ayon:apex_rig = "H:/AYON/projects/Experiments/ideas/apex_usd/apex_rig_animation_no_solaris/publish/rig/rigMain/v001/Expr_apex_rig_animation_no_solaris_rigMain_v001.bgeo.sc"
```

client/ayon_houdini/plugins/create/create_bgeo.py
```diff
@@ -38,7 +39,7 @@ def create(self, product_name, instance_data, pre_create_data):
 
         parms = {}
 
-        instance_node.parm("trange").set(1)
+        self.set_trange(instance_node)
```
moonyuet, line 42: Any reason why you choose to create the function for this instead of adopting the original approach(i.e. `instance_node.parm("trange").set(1)`)?

MustafaJafar: This is shared with different creators, that only requires set range to a different value. so I moved it to a function.

MustafaJafar:
Client reported this PR works for them [here](https://discord.com/channels/517362899170230292/1495805462464168039/1496163671280455880).  and the HDA still need additional tweaking #384   
Tagging @moonyuet @BigRoy

client/ayon_houdini/plugins/publish/extract_apex_usd.py
```diff
@@ -0,0 +1,104 @@
...
+        specific_version = isinstance(version_name, int)
+        path = get_instance_expected_output_path(
+            instance,
+            representation_name="bgeo",
+            ext=ext,
+            version=version_name if specific_version else None
+        )
+        # Ensure `None` for now is also a string
+        path = str(path)
+        path = os.path.normpath(path)
```
moonyuet: `get_instance_expected_output_path` function is already returning resolved path see [here](https://github.com/ynput/ayon-core/blob/d2e17592638e2ef8e4aa685249c3f4e7da7126e8/client/ayon_core/pipeline/publish/lib.py#L1054)
```suggestion
```

client/ayon_houdini/plugins/publish/extract_apex_usd.py
moonyuet: Or we can do this?

```suggestion
        normalized path = Path(path)
        return str(normalized_path)
```

moonyuet (approved):
Looks good for me, we can merge this after you resolve the comment above.

client/ayon_houdini/plugins/publish/extract_apex_usd.py
```diff
@@ -0,0 +1,102 @@
...
+        path = get_instance_expected_output_path(
+            instance,
+            representation_name="bgeo",
+            ext=ext,
+            version=version_name if specific_version else None
+        )
+        if path:
+            return path
+
+        raise RuntimeError("Unable to resolve publish path.")
```
MustafaJafar: @moonyuet what do you think. it's inspired by https://github.com/ynput/ayon-core/blob/7b863f71efce090114225b25b3ab9c1524a2a8f8/client/ayon_core/plugins/publish/extract_usd_layer_contributions.py#L177-L179

moonyuet: I am ok with that.

client/ayon_houdini/plugins/publish/collect_apex_usd.py
```diff
@@ -0,0 +1,25 @@
...
+class CollectAPEXUSD(plugin.HoudiniInstancePlugin):
+    """Collect APEX USD Rig Layer
+
+    Create Empty USD Rig layer"""
+
+    # Run Before CollectUSDLayerContributions from core plugins
+    order = pyblish.api.CollectorOrder + 0.25
+    label = "Collect APEX To USD"
+
+    families = ["rig"]
```
BigRoy: Does it make more sense for this to target `apex` family perhaps? 🤔

client/ayon_houdini/plugins/publish/collect_apex_usd.py
```diff
@@ -0,0 +1,25 @@
...
+
+    def process(self, instance):
+        sdf_layer = Sdf.Layer.CreateAnonymous()
+        self.log.debug(f"Creating USD rig layer: {sdf_layer}")
+        instance.data["rig_layer"] = sdf_layer
```
BigRoy: Remind me - why did this need to be created at Collector Order already? I see no other thing relying on `rig_layer` instance data? Why can't we create this anonymous layer in the extractor when we need it?

If there's reasons to do so - can we please add it to the docstring of the plug-in explaining why we need to. Currently it feels like it's just complicating things?

client/ayon_houdini/plugins/publish/extract_apex_usd.py
```diff
@@ -0,0 +1,101 @@
...
+        #   if it is a path, use it in the file read and add it as asset.
+        #   if not load model layer and add it as prop.
+        rig_attr = Sdf.AttributeSpec(
+            asset_prim,
+            "ayon:apex_rig",
+            Sdf.ValueTypeNames.String,
+            variability=Sdf.VariabilityUniform
+        )
+        rig_attr.default = self.get_representation_path_in_publish_context(
+            instance)
```
BigRoy: I think we should make this an AYON entity URI - possibly even _always_ regardless of settings. Why? Because how we're using this attribute and what we're loading through it is completely our logic - and _that_ logic should just be doing a call to AYON backend possibly to resolve the path? Yes, it'd require an additional database call to resolve the path. But it'd solve many other things (cross-OS most specifically, etc.)

Unless the HDA and logic we provide to load the rigs is entirely AYON-agnostic somehow I don't see a reason to use regular paths here. 🤔 But to avoid the issue that I'm wrong here (which I usually am, we can potentially allow this to be a path, but default in settings that we author AYON entity URI for the attribute, but allow to disable that and write regular paths instead. Which I suppose the "make relative USD" plug-in in `ayon-usd` would be able to make relative, right?)

client/ayon_houdini/plugins/publish/collect_apex_usd.py
MustafaJafar: > Remind me - why did this need to be created at Collector Order already? I see no other thing relying on rig_layer instance data? Why can't we create this anonymous layer in the extractor when we need it?

If I'm not mistaken, because of `CollectUSDLayerContribution`. 
It needs layers to include in the usdAsset.

client/ayon_houdini/plugins/publish/extract_apex_usd.py
MustafaJafar: IMO, making it relative is the first step. but, I'm not sure about the recommended way to do it.
that why I made this comment https://github.com/ynput/ayon-houdini/pull/371#discussion_r3072374867

I think for a second step to add option to allow using URIs. (I think it can be done in a follow up PR)
and maybe we can add some action to file parameters in compliance to `Load with AYON` that adds some expression for resolving the URI.

client/ayon_houdini/plugins/publish/collect_apex_usd.py
MustafaJafar: I believe it will work. but, I didn't test it.
tbh, I tried to make the implementation a bit generic to support more apex products. like animation that I explored here #376

client/ayon_houdini/plugins/publish/extract_apex_usd.py
BigRoy: > IMO, making it relative is the first step. but, I'm not sure about the recommended way to do it.
that why I made this comment https://github.com/ynput/ayon-houdini/pull/371#discussion_r3072374867

Well, the load logic that applies to this asset path is entirely custom - we don't NEED the AYON USD resolver to resolve these - we can just call the method explicitly from the backend, etc. As such, I don't see much reason not to use an entity URI here. We have full control, and it's the most explicit type of asset paths we have for AYON entities.

Anyway, I think making the path relative - would already be completely solved by just enabling the `USDOutputProcessorRemapToRelativePaths` (Process USD files to use relative paths - plug-in) in `ayon-usd`. It needs no changes here, right?

BigRoy: Ah wait! :) This should not be a `String`, but an `Asset` - That should solve the relative path aspect I described [here](https://github.com/ynput/ayon-houdini/pull/371/changes#r3156971658) by using the `ayon-usd` plugin.
The value should then be `Sdf.AssetPath`.

client/ayon_houdini/plugins/publish/collect_apex_usd.py
BigRoy: Ah I see - and in that scenario you feel that both are `apex` even though on is the rig and the other is the animation. Maybe I'd just expect the families to be more explicit then. I don't see the overlap where you'd use the `apex` family at the moment then. But `rig` is fine for me then.

client/ayon_houdini/plugins/publish/extract_apex_usd.py
BigRoy: Should not be a string, but `Asset` spec. See: https://github.com/ynput/ayon-houdini/pull/371#discussion_r3156981445

client/ayon_houdini/plugins/publish/collect_apex_usd.py
BigRoy: But that one also wouldn't use `rig_layer` from `instance.data` anyway, right? So I'm not sure how this currently would solve anything 🤔

client/ayon_houdini/plugins/publish/extract_apex_usd.py
MustafaJafar: So, should it be like this:
```suggestion
        rig_attr.default = Sdf.AssetPath(
            self.get_representation_path_in_publish_context(
                instance))
```

BigRoy: I think so, yes - but you'd need to adjust the AttributeSpec type as well.

client/ayon_houdini/plugins/publish/collect_apex_usd.py
BigRoy: Can you double check @MustafaJafar if you can live without this?

MustafaJafar:
@BigRoy so, the collector plugin was redundant. I've removed it in https://github.com/ynput/ayon-houdini/pull/371/commits/00ad6fd0435b58e059fcf81f584fc8955e1c9a6f

client/ayon_houdini/plugins/publish/extract_apex_usd.py
```diff
@@ -0,0 +1,106 @@
...
+    label = "Extract APEX To USD"
+
+    families = ["rig"]
+
+    def process(self, instance):
+        """Inject the current working file"""
+
+        sdf_layer = Sdf.Layer.CreateAnonymous()
+        self.log.debug(f"Creating USD rig layer: {sdf_layer}")
+        instance.data["rig_layer"] = sdf_layer
```
BigRoy: We have no other plug-in using this data, so I think can we avoid storing it. Right?

```suggestion
```

client/ayon_houdini/plugins/publish/extract_apex_usd.py
```diff
@@ -0,0 +1,106 @@
...
+        default_prim_path = Sdf.Path(f"/{default_prim.strip('/')}")
+        asset_prim = get_or_define_prim_spec(
+            sdf_layer,
+            default_prim_path,
+            "Xform"
+        )
+        asset_prim.specifier = Sdf.SpecifierOver
+        # "ayon:apex_rig" if apex rig exist:
+        #   if it is a path, use it in the file read and add it as asset.
+        #   if not load model layer and add it as prop.
```
BigRoy: Comment is not relevant/inaccurate?
```suggestion
```

client/ayon_houdini/plugins/publish/extract_apex_usd.py
```diff
@@ -0,0 +1,106 @@
...
+
+    order = pyblish.api.ExtractorOrder
+    label = "Extract APEX To USD"
+
+    families = ["rig"]
+
+    def process(self, instance):
+        """Inject the current working file"""
+
+        sdf_layer = Sdf.Layer.CreateAnonymous()
```
BigRoy: Since we now know where we'll be writing out the resulting file will need to go, I feel like we can instead now use `Sdf.Layer.CreateNew` with the identifier being the filepath, and then we can just trigger `sdf_layer.Save()` when we've completed our attribute inside the scene file?

Thoughts @MustafaJafar ?

client/ayon_houdini/plugins/publish/extract_apex_usd.py
```diff
@@ -0,0 +1,106 @@
...
+        path = get_instance_expected_output_path(
+            instance,
+            representation_name="bgeo",
+            ext=ext,
+            version=version_name if specific_version else None
+        )
+        if path:
+            return path
+
+        raise RuntimeError("Unable to resolve publish path.")
```
BigRoy: Can we add in the error the message of the "bgeo" name it looked for?

client/ayon_houdini/plugins/publish/extract_apex_usd.py
MustafaJafar: It's irrelevant. 
However, it's a useful piece of advice for making automated setup when load the apex rig.
After loading the usdAsset,
- If `ayon:apex_rig` is found, use `sceneaddcharacter`
- If it's not found, use `sceneaddprop`

client/ayon_houdini/plugins/publish/extract_apex_usd.py
MustafaJafar: I moved the layer creation to the top.

client/ayon_houdini/plugins/publish/extract_apex_usd.py
MustafaJafar: Do you mean the product name it looked for?

Unfortunately, the `get_instance_expected_output_path` function only returns path or none.
I believe we only have 
- file_path in work directory
- ext
- version_name
- representation_name

client/ayon_houdini/plugins/publish/extract_apex_usd.py
MustafaJafar: I removed these commects. do you think this piece of information should exist some where?

client/ayon_houdini/plugins/publish/extract_apex_usd.py
BigRoy: The product name would be nice - but I suppose even just reading the representation name we looked for already helps.

client/ayon_houdini/plugins/publish/extract_apex_usd.py
BigRoy: What you're describing [here](https://github.com/ynput/ayon-houdini/pull/371#discussion_r3167254675) is not at all what I got from the comments in the code. I think this sounds more like info for documentation or alike than for this part of the codebase.

client/ayon_houdini/plugins/publish/extract_apex_usd.py
```diff
@@ -0,0 +1,101 @@
...
+            variability=Sdf.VariabilityUniform
+        )
+        rig_path = self.get_representation_path_in_publish_context(
+            instance
+        )
+        self.log.debug(f"Setting 'ayon:apex_rig' attr to '{rig_path}'")
+        rig_attr.default = Sdf.AssetPath(rig_path)
+
+        # Save the file
+        sdf_layer.Export(layer_path, args={"format": "usda"})
```
BigRoy: You can specify the format on `CreateNew` and then just use `Save()` - now with still exporting, you're not getting the potential optimization benefit that you're already operating on the Sdf Layer set to the filepath. Meaning you don't need to specify layer path here again.


```suggestion
        sdf_layer.Save()
```
And add the format to the CreateNew call.

MustafaJafar:
I've gave it one more test and it works.
@BigRoy Let me know what do you think.

BigRoy (approved):
Awesome, looks good to me.

Not entirely sure whether we'll want to end up ALWAYS writing the USD file (and hence with that also deal with the USD contributions and whatnot) but since APEX rig adoption in Houdini is super low currently I think it's fine to take this along now and we can allow to disable it in some form later if requested.

Also, I'm still thinking using an entity URI for the apex rig asset path is better. But since it's now an Asset attribute, I feel it's fine for now.

PR #366, opened 2026-03-04
Enhancement: AYON APEX Publishing Animation

MustafaJafar:
## Changelog Description
- Add animation creator
- Move rig and animation creators to Bgeo file
- Add APEX animation validator

## Additional information
I've also explored more ways to publish animations, we can at some point add FBX character export.

This PR serves as an exploration working with APEX in AYON.

# Demo
[Turret Example AYON APEX Workflow.zip](https://github.com/user-attachments/files/25741133/Turret.Example.AYON.APEX.Workflow.zip)

https://github.com/user-attachments/assets/b5b85efa-bdcd-434d-97f7-25faf4b576e7

## Testing notes:
1. Add characters and props
2. Create APEX Animation
3. Publish

You can refer to the demo for more info.

PR #376, opened 2026-03-23
Publish apex animation

### LINKED ISSUE #365: AYON APEX Workflow
This is part of the ticket about APEX and USD support #351

MustafaJafar:
## Changelog Description
This PR is for publishing animation data from packed APEX assets. it was built on top of #371

- Add animation creator
- Add APEX animation validator

## A word on APEX Animation
Typically, this PR allows publishing the animation data only so that we can apply them later in different scenes instead of republishing the whole rig asset again OR publishing baked animation.

As long as artists are using `APEX Scene Copy Animation` node, then everything will be cool.

image

However, when they add a scene animate node, as soon as they go through animation mode in the scene view, the node will stash the animation data. which means the `copy animation` nodes can be removed and the animation will just be persistent. this is a result of Houdini using stashing. _and, in this sense, I'm afraid loading different animations won't have any effect. and it can be confusing for artists. and this is something not related to AYON/saving animation data as Bgeo._ 
image

I made this PR while exploring APEX and AYON, mainly in #366 .
Then to make things cleaner, any needed fixes for publishing APEX were moved to #371 as well as the original requirement for AYON APEX Rigging USD workflow while moving the animation part to this PR.

# Demo
[Turret Example AYON APEX Workflow.zip](https://github.com/user-attachments/files/25741133/Turret.Example.AYON.APEX.Workflow.zip)

https://github.com/user-attachments/assets/b5b85efa-bdcd-434d-97f7-25faf4b576e7

## Testing notes:
1. Add characters and props
2. Create APEX Animation
3. Publish

Resolve #365

### DIFF
--- a/client/ayon_houdini/plugins/publish/validate_apex_animation.py
+++ b/client/ayon_houdini/plugins/publish/validate_apex_animation.py
@@ -1,0 +1,121 @@
+# -*- coding: utf-8 -*-
+import hou
+
+import pyblish.api
+from ayon_core.pipeline import PublishValidationError
+from ayon_core.pipeline.publish import RepairAction
+from ayon_houdini.api.action import SelectInvalidAction
+from ayon_houdini.api import plugin
+
+
+class IsolateAnimation(RepairAction):
+    label = "Isolate Animation"
+
+class ValidateAPEXAnim(plugin.HoudiniInstancePlugin):
+    """Validate output node has animation APEX data.
+
+    Check if packed prim files match only "/animation".
+    This is crucial to prevent any issue when loading the animation.
+    """
+    # Should run after ValidateSopOutputNode
+    order = pyblish.api.ValidatorOrder + 0.1
+    families = ["anim"]
+    label = "Validate APEX Animation"
+    actions = [SelectInvalidAction, IsolateAnimation]
+
+    def process(self, instance):
+
+        if not instance.data.get("instance_node"):
+            # Ignore instances without an instance node
+            # e.g. in memory bootstrap instances
+            self.log.debug(
+                "Skipping instance without instance node: {}".format(instance)
+            )
+            return
+
+        invalid_nodes, message = self.get_invalid_with_message(instance)
+        if invalid_nodes:
+            raise PublishValidationError(
+                message,
+                title=self.label
+            )
+
+    @classmethod
+    def get_invalid_with_message(cls, instance):
+
+        output_node = instance.data.get("output_node")
+
+        geo = output_node.geometry()
+
+        # TODO: Check if this is a bgeo scene.
+        # Same as the error in the Rig Tree
+        # Invalid Hierarchy: No valid roots found, possible closed polygon
+
+        anim_paths = set(geo.extractPackedPaths('/anim**'))
+        if not anim_paths:
+            error = (
+                f"Output SOP node '{output_node}' doesn't"
+                " have 'animation' folder to export."
+                " Please check your RigTree."
+            )
+            return [output_node, error]
+
+        all_paths = set(geo.extractPackedPaths('/**')) - set("/")
+
+        if all_paths != anim_paths:
+            error = (
+                f"Output SOP node '{output_node.path()}' should only"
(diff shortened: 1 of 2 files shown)


client/ayon_houdini/plugins/publish/validate_apex_animation.py
```diff
@@ -0,0 +1,121 @@
...
+                " have 'animation' folder in RigTree."
+            )
+            return [output_node, error]
+
+        return [None, None]
+
+    @classmethod
+    def get_invalid(cls, instance):
+        node, _ = cls.get_invalid_with_message(instance)
+        return [node]
```
moonyuet, line 77: EDIT: ignore me my previous comment. Maybe it is more explicit to use tuple instead of list like (node, error) instead of [node, error]?

client/ayon_houdini/plugins/publish/validate_apex_animation.py
```diff
@@ -0,0 +1,121 @@
...
+    @classmethod
+    def repair(cls, instance):
+        """Isolate Animation Action.
+
+        It is a helper action more than a repair action,
+        used to add a default single value for the path.
+        """
+
+        output_node = instance.data.get("output_node")
+        if cls.get_invalid(instance) != [output_node]:
```
moonyuet, line 88: I believe it does the same thing?
```suggestion
        if output_node not in cls.get_invalid(instance):
```

MustafaJafar: It does the same thing I believe. 
I picked this way, because this is always a list with one item.
Also, it's the same as here
https://github.com/ynput/ayon-houdini/blob/d5f849fc991ec3f54b506cf48d589bb9bede33dd/client/ayon_houdini/plugins/publish/validate_primitive_hierarchy_paths.py#L147

client/ayon_houdini/plugins/publish/validate_apex_animation.py
MustafaJafar: Tbh, I don't know I just followed what we already have in 
https://github.com/ynput/ayon-houdini/blob/d5f849fc991ec3f54b506cf48d589bb9bede33dd/client/ayon_houdini/plugins/publish/validate_vdb_output_node.py#L130

moonyuet (commented):
I published animation successfully. Enclosed with the publish report.

[publish-report-260417-17-10.json](https://github.com/user-attachments/files/26817727/publish-report-260417-17-10.json)

Maybe for separate issue. When I tested with the validator(by bypassing the packed folder), it blocks the validation for sure with the sop_node validator. But if I hit select invalid It would give the error below
```
Action: SelectInvalidAction (094d16b0-5f55-4fff-88bf-a9ff5018a32b)

Error: Action failed.

  File "D:\ayon-addon_template\ayon-houdini\client\ayon_houdini\plugins\publish\validate_sop_output_node.py", line 80, in get_invalid
    output_node.cook()
  File "C:\PROGRA~1/SIDEEF~1/HOUDIN~1.512/houdini/python3.11libs\hou.py", line 19175, in cook
    return _hou.OpNode_cook(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error while cooking.

During handling of the above exception, another exception occurred:

  File "C:\Users\Kayla\AppData\Local\Ynput\AYON\dependency_packages\ayon_2603121036_windows.zip\dependencies\pyblish\plugin.py", line 528, in __explicit_process
    runner(*args)
  File "D:\ayon-addon_template\ayon-houdini\client\ayon_houdini\api\action.py", line 30, in process
    invalid_nodes = plugin.get_invalid(instance)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\ayon-addon_template\ayon-houdini\client\ayon_houdini\plugins\publish\validate_sop_output_node.py", line 82, in get_invalid
    raise PublishValidationError(
ayon_core.pipeline.publish.publish_plugins.PublishValidationError: Failed to cook node: /obj/apex_rig/turret_asset.
```

PR #377, opened 2026-03-26
Review Instance: Validate camera prim from Lops

### LINKED ISSUE #170: Review: Support OpenGL with "source" set to LOPs

Currently the review product type using the OpenGL ROP only allows publishing from object level and not from LOPs (Solaris).

### Suggested implementation?

It should completely support the LOPs setup for publishing:

Image

1. All validations should pass fine, and preferably a validator validates the ROPs configuration is correct and is capable of rendering (e.g. correct camera set, etc.)
2. [The (animated) focal length collection](https://github.com/ynput/ayon-houdini/blob/323dd09ddb6a8d7f4ed3238ae537198f69ef2991/client/ayon_houdini/plugins/publish/collect_review_data.py#L37-L56) should be support the Solaris camera so it can be applied in the burnins

---

There was once a draft in OpenPype, see here: https://github.com/ynput/OpenPype/pull/4890

A workaround would be publish review from `lopimportcam` on `/obj` level instead as requested here: https://github.com/ynput/ayon-houdini/issues/23 - which would be implemented by https://github.com/ynput/ayon-houdini/pull/169 however we should just support native Solaris OpenGL render instead.

This also came up in https://github.com/ynput/ayon-houdini/pull/167 here: https://github.com/ynput/ayon-houdini/pull/167#issuecomment-2465068633

woutbelt:
## Changelog Description
Added supoprt fro LOP cameras.

Refactored the [get_invalid_camera_path] method to handle both OBJ and LOP (USD) camera validation.
For OBJ nodes, the function checks if the camera path exists and is a valid camera node (cam or lopimportcam) stayed the same as how it was.
For LOP nodes (opsource == 1), the function retrieves the LOP network from the loppath parameter and checks if the specified USD camera prim exists in the stage.

Fix #170 

## Testing notes:
Create or open a Houdini scene with an LOP (USD) camera setups.
Assign the appropriate camera path or USD camera prim to the review node.
Run the Ayon publisher and trigger the review validation.

Confirm that:
For LOP, missing or invalid USD camera prims are correctly reported.

### DIFF
--- a/client/ayon_houdini/plugins/publish/validate_scene_review.py
+++ b/client/ayon_houdini/plugins/publish/validate_scene_review.py
@@ -47,16 +47,29 @@ class ValidateSceneReview(plugin.HoudiniInstancePlugin):
             return "Scene path does not exist: '{}'".format(path)
 
     def get_invalid_camera_path(self, rop_node):
-        camera_path_parm = rop_node.parm("camera")
-        camera_node = camera_path_parm.evalAsNode()
-        path = camera_path_parm.evalAsString()
-        if not camera_node:
-            return "Camera path does not exist: '{}'".format(path)
-        type_name = camera_node.type().name()
-        if type_name not in {"cam", "lopimportcam"}:
-            return "Camera path is not a camera: '{}' (type: {})".format(
-                path, type_name
-            )
+        opsource = rop_node.parm("opsource").eval()
+
+        if opsource != 1:
+            camera_path_parm = rop_node.parm("camera")
+            camera_node = camera_path_parm.evalAsNode()
+            path = camera_path_parm.evalAsString()
+            if not camera_node:
+                return "Camera path does not exist test: '{}'".format(path)
+            type_name = camera_node.type().name()
+            if type_name not in {"cam", "lopimportcam"}:
+                return "Camera path is not a camera: '{}' (type: {})".format(
+                    path, type_name
+                )
+        else:
+            lop_path = rop_node.parm("loppath").evalAsString()
+            camera_path_parm = rop_node.parm("cameraprim")
+
+            stage = hou.node(lop_path).stage()
+
+            if not stage.GetPrimAtPath(camera_path_parm.evalAsString()):
+                return "Camera prim does not exist: '{}'".format(
+                    camera_path_parm.evalAsString()
+                )
 
     def get_invalid_resolution(self, rop_node):
 
--- a/client/ayon_houdini/plugins/publish/validate_scene_review.py
+++ b/client/ayon_houdini/plugins/publish/validate_scene_review.py
@@ -49,7 +49,8 @@ class ValidateSceneReview(plugin.HoudiniInstancePlugin):
     def get_invalid_camera_path(self, rop_node):
         opsource = rop_node.parm("opsource").eval()
 
-        if opsource != 1:
+        if opsource == 0:
+            # Object-level
             camera_path_parm = rop_node.parm("camera")
             camera_node = camera_path_parm.evalAsNode()
             path = camera_path_parm.evalAsString()
@@ -60,7 +61,8 @@ class ValidateSceneReview(plugin.HoudiniInstancePlugin):
                 return "Camera path is not a camera: '{}' (type: {})".format(
                     path, type_name
                 )
-        else:
+        elif opsource == 1:
+            # LOP-level
             lop_path = rop_node.parm("loppath").evalAsString()
             camera_path_parm = rop_node.parm("cameraprim")
 
@@ -70,6 +72,8 @@ class ValidateSceneReview(plugin.HoudiniInstancePlugin):
                 return "Camera prim does not exist: '{}'".format(
                     camera_path_parm.evalAsString()
                 )
+        else:
+            return "Unsupported camera source type: {}".format(opsource)
 
     def get_invalid_resolution(self, rop_node):
 
--- a/client/ayon_houdini/plugins/publish/validate_scene_review.py
+++ b/client/ayon_houdini/plugins/publish/validate_scene_review.py
@@ -9,8 +9,8 @@ from ayon_houdini.api import plugin
 
 class ValidateSceneReview(plugin.HoudiniInstancePlugin):
     """Validator Some Scene Settings before publishing the review
-        1. Scene Path
-        2. Resolution
+    1. Scene Path
+    2. Resolution
     """
 
     order = pyblish.api.ValidatorOrder
@@ -35,9 +35,7 @@ class ValidateSceneReview(plugin.HoudiniInstancePlugin):
             report.extend(invalid)
 
         if report:
-            raise PublishValidationError(
-                "\n\n".join(report),
-                title=self.label)
+            raise PublishValidationError("\n\n".join(report), title=self.label)
 
     def get_invalid_scene_path(self, rop_node):
         scene_path_parm = rop_node.parm("scenepath")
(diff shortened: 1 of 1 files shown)


BigRoy:
I vaguely recall that if we support LOPS we'll also need to support the collecting of the focal length on a per frame basis.

But likely for now you're just getting a warning logged from here: https://github.com/ynput/ayon-houdini/blob/b0e60d70c13b65c5b6423755f182f39f0fd79b76/client/ayon_houdini/plugins/publish/collect_review_data.py#L31-L53

Best case scenario we can somehow get the animated focal length from the USD stage. 🤔 But we'd need to evaluate the stage over time range etc. and may have massive impact on performance - so likely not trivial unfortunately.

client/ayon_houdini/plugins/publish/validate_scene_review.py
```diff
@@ -47,16 +47,29 @@ def get_invalid_scene_path(self, rop_node):
...
+        else:
+            lop_path = rop_node.parm("loppath").evalAsString()
+            camera_path_parm = rop_node.parm("cameraprim")
+
+            stage = hou.node(lop_path).stage()
+
+            if not stage.GetPrimAtPath(camera_path_parm.evalAsString()):
+                return "Camera prim does not exist: '{}'".format(
+                    camera_path_parm.evalAsString()
+                )
```
BigRoy: I'd rather be more explicit and do

```python
if opsource == 0:
    # Object-level

elif opsource == 1:
    # LOPs

else:
    raise ValueError(f"Unsupported source for OpenGL rop: {opsource}")
```

woutbelt:
> I vaguely recall that if we support LOPS we'll also need to support the collecting of the focal length on a per frame basis.
> 
> But likely for now you're just getting a warning logged from here:
> 
> https://github.com/ynput/ayon-houdini/blob/b0e60d70c13b65c5b6423755f182f39f0fd79b76/client/ayon_houdini/plugins/publish/collect_review_data.py#L31-L53
> 
> Best case scenario we can somehow get the animated focal length from the USD stage. 🤔 But we'd need to evaluate the stage over time range etc. and may have massive impact on performance - so likely not trivial unfortunately.

Do you recall why we would need to keep track of the focal length?, having it anyazlie the stage every frame would be quite heavy 

> I vaguely recall that if we support LOPS we'll also need to support the collecting of the focal length on a per frame basis.
> 
> But likely for now you're just getting a warning logged from here:
> 
> https://github.com/ynput/ayon-houdini/blob/b0e60d70c13b65c5b6423755f182f39f0fd79b76/client/ayon_houdini/plugins/publish/collect_review_data.py#L31-L53
> 
> Best case scenario we can somehow get the animated focal length from the USD stage. 🤔 But we'd need to evaluate the stage over time range etc. and may have massive impact on performance - so likely not trivial unfortunately.

Would it then be best to check if the focal length primvar has time samples? If not, we can presume the focal length stays the same throughout the full frame range and only do a check on one frame. Otherwise, we analyze the full range. I do still think having it check the full range would become extremely slow for heavy scenes.

BigRoy:
> Do you recall why we would need to keep track of the focal length?, having it anyazlie the stage every frame would be quite heavy

It's mostly for the burnin to display animated focal lengths - if we can somehow get the OpenGL ROP to write the focal length metadata into the image sequence files (or some other means as it's calculating anyway) that'd be great because then we could take the info from there instead of needing to query the scene and potential benefit is that it'd work for both cases.

> Would it then be best to check if the focal length primvar has time samples? If not, we can presume the focal length stays the same throughout the full frame range and only do a check on one frame. Otherwise, we analyze the full range. I do still think having it check the full range would become extremely slow for heavy scenes.

I recall that because Houdini only caches the current frame that it'd assume the camera attribute wasn't animated, even if technically it may be over time in the scene. So it was rather tricky.

I believe some of it was mentioned in some of the links that came up here: https://github.com/ynput/ayon-houdini/issues/170#issue-2647760083

woutbelt:
I have done some testing. You can detect if something has a time sample, even if it's in an implicit or explicit layer in Houdini by using GetNumTimeSamples() instead of ValueMightBeTimeVarying() what you would normaly use.

```python
from pxr import UsdGeom

node = hou.pwd()
stage = node.editableStage()

camera_prim = stage.GetPrimAtPath('/camera')

camera = UsdGeom.Camera(camera_prim)
focal_attr = camera.GetFocalLengthAttr()

value = focal_attr.GetNumTimeSamples()

print(value)
```

This will either return 1 or 0 if it is animated in an implicit layer, or a higher number if it is a layer written to disk.

I don't believe there is a proper way to write out metadata with either the flipbook or OpenGL nodes in Houdini natively, with the exception of doing a post-script after the render. But that sounds like quite a hassle.

client/ayon_houdini/plugins/publish/validate_scene_review.py
```diff
@@ -47,16 +47,33 @@ def get_invalid_scene_path(self, rop_node):
...
+            # LOP-level
+            lop_path = rop_node.parm("loppath").evalAsString()
+            camera_path_parm = rop_node.parm("cameraprim")
+
+            stage = hou.node(lop_path).stage()
+
+            if not stage.GetPrimAtPath(camera_path_parm.evalAsString()):
+                return "Camera prim does not exist: '{}'".format(
+                    camera_path_parm.evalAsString()
+                )
```
MustafaJafar:
```suggestion
            camera_path = rop_node.parm("cameraprim").evalAsString()

            stage = hou.node(lop_path).stage()
            camera_prim = stage.GetPrimAtPath(camera_path)

            if not camera_prim or not camera_prim.IsValid():
                return f"Camera prim does not exist: '{camera_path}'"

            if not camera_prim.IsA(pxr.UsdGeom.Camera):
                return f"Camera path '{camera_path}' is not a camera."
```

client/ayon_houdini/plugins/publish/validate_scene_review.py
```diff
@@ -47,16 +47,33 @@ def get_invalid_scene_path(self, rop_node):
...
-        camera_node = camera_path_parm.evalAsNode()
-        path = camera_path_parm.evalAsString()
-        if not camera_node:
-            return "Camera path does not exist: '{}'".format(path)
-        type_name = camera_node.type().name()
-        if type_name not in {"cam", "lopimportcam"}:
-            return "Camera path is not a camera: '{}' (type: {})".format(
-                path, type_name
-            )
+        opsource = rop_node.parm("opsource").eval()
```
MustafaJafar, line 51:
```suggestion
        import pxr

        opsource = rop_node.parm("opsource").eval()
```

MustafaJafar (commented):
The validation works.

btw, I like how it looks like in https://github.com/ynput/ayon-houdini/blob/d5f849fc991ec3f54b506cf48d589bb9bede33dd/client/ayon_houdini/plugins/publish/validate_usd_render_arnold.py#L265-L278

MustafaJafar:
Hey, I made this one #382 to allow selecting a lop node on creating the review instance.
Btw, I think we may create another PR for the issue of focal length.

client/ayon_houdini/plugins/publish/validate_scene_review.py
```diff
@@ -47,16 +47,39 @@ def get_invalid_scene_path(self, rop_node):
...
+            # LOP-level
+            lop_path = rop_node.parm("loppath").evalAsString()
+            camera_path = rop_node.parm("cameraprim").evalAsString()
+
+            stage = hou.node(lop_path).stage()
+            camera_prim = stage.GetPrimAtPath(camera_path)
+
+            if not camera_prim or not camera_prim.IsValid():
+                return f"Camera prim path does not exist: '{camera_path}'"
+
```
MustafaJafar: There's a white space on this line causing the code linting check to fail
```suggestion

````

MustafaJafar (approved):
It works well on my side.
Imo, we can mention it on that this feature work but it misses getting focal length.

BigRoy (approved):

BigRoy:
@woutbelt can you pull upstream into your branch (it's out of date) then we can merge it.

woutbelt:
@BigRoy All done my branch should be up to date with the latest develop branch.

PR #382, opened 2026-03-30
CreateReview: allows selecting a Lop node

### LINKED ISSUE #170: Review: Support OpenGL with "source" set to LOPs

MustafaJafar:
## Changelog Description
- CreateReview: allows selecting a Lop node

It'll select the first camera it finds.

## Testing notes:
1. select a Lop node
2. Create a review instance
3. You should see the camera prim in the node.

Resolve #170


client/ayon_houdini/plugins/create/create_review.py
```diff
@@ -82,19 +82,32 @@ def create(self, product_name, instance_data, pre_create_data):
...
+                elif node.type().category().name() == "Lop":
+                    for prim in node.stage().Traverse():
+                        if prim.GetTypeName() == "Camera":
+                            camera = prim.GetPath().pathString
+                            break
+                    parms.update({
+                        "opsource": 1,
+                        "loppath": path,
+                        "cameraprim": camera or "",
+                    })
```
BigRoy: This doesn't actually break on the first found valid entry in the selection. Can you change it so that it does?

So whenever it has found any object or lop camera, then stop the outer for loop over selected nodes.

BigRoy (changes requested):
Implement https://github.com/ynput/ayon-houdini/pull/382/changes#r3013870514 and this looks good to me after.

client/ayon_houdini/plugins/create/create_review.py
MustafaJafar: should be fixed with https://github.com/ynput/ayon-houdini/pull/382/commits/b30467ce7cfa9bb136ed19703e4b197514b4fa7a

BigRoy (approved):

PR #384, opened 2026-04-15
APEX Rig: Add AYON Load APEX Rig OTL

MustafaJafar:
## Changelog Description
-  Add automation that auto loads the rig associated with the rig usd layer and add characters and props to scene.
image

Why am I making a different PR instead of pushing it directly to #371?
Because this Houdini, there can be different ways to implement this node. e.g. we can add the lop import inside the HDA and expose all of the parameters and always use a for loop.
Therefore, I created this PR to discuss it separately from the base PR.

Additional thoughts:
- Personally, I went minimal and add the core building block for loading the primitive and adding a character or prop to scene based on the existence of `apex_rig` parameter. and left the rest for artists/TDs to use it the way they see correct.
- This setup doesn't require any post modification to unpack the assets with the expected data so that you can import them directly to solaris after unpacking them.
    image
- **This setup requires clean paths in the asset geometry before publishing as USD Rig. e.g. you can check Houdini's Turret Asset Example from [KINEFX | APEX GRAPH AND SCRIPT](https://www.sidefx.com/tutorials/kinefx-apex-graph-and-script/).** 

## Testing notes:
Here's a workfile example if you would like one [generic.zip](https://github.com/user-attachments/files/26752200/generic.zip)
1. load a USD Asset with USD APEX Rig Layer, see #371 
2. As the screen shows,
    - Go to SOPs, either `/obj/geo_node` or add a SOPs network in `/stage`. (It doesn't work with "sop modify" no idea why and not related to this PR).
    - Add either one or many "LOP import" and select the Lop Path, primitives, enable path and name attributes, and any configurations you want. (The only requirement here to have the primitives list without any child prims)
    - add ayon_load_apex_rig either directly below the node if it has one primitive or in a for loop if it has a list of primitives


moonyuet (dismissed):
Tested successfully with Houdini 21.0 in some simple tests.
With the rig loader I can load accordingly the rig hooked to the shape
Image

moonyuet:
@MustafaJafar @BigRoy is it good to merge?

MustafaJafar:
> @MustafaJafar @BigRoy is it good to merge?

I'm converting this PR as Draft to avoid merging it by accident. 
Client reported back _see [here](https://discord.com/channels/517362899170230292/1495805462464168039/1496163671280455880)_ they have some suggestions and they will share it with us.

PR #388, opened 2026-05-23
387 yn 0733 add usd resolver clear cache button or hda to houdini

### LINKED ISSUE #387: YN-0733: Add USD Resolver Clear Cache button or HDA to Houdini

Add USD Resolver Clear Cache button or HDA to Houdini 

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1494321623606628526/1503705218271936563)

_This issue was automatically created from AYON ticket [YN-0733]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=f793c7ecea9c45239a2911c402e3b06f)

MustafaJafar:
## Changelog Description
- [x] Add `Clear USD Resolver Cache` to the AYON Menu
- [x] Add clear resolver cache functionality to `Clear Cache` button in AYON loader HDAs
- [x] ~~Reset View~~ Reload layers and views after clearing the resolver cache

## Additional Information
Some Screenshots 
image
image

## Testing notes:
one way of testing is with an assembly asset that references two assets with URIs and latest version token.
Then publishing a new version of the inner asset and clearing the cache in the scene and it should work.
image

Resolve #387


client/ayon_houdini/api/usd.py
```diff
@@ -425,3 +425,16 @@ def get_ayon_entity_uri_from_representation_context(context: dict) -> str:
...
+def clear_resolver_cache(reset_view=True):
+    try: 
+        from usdAssetResolver import AyonUsdResolver
+
+        ctx = AyonUsdResolver.ResolverContext()
+        # Clear the resolver cache.
+        ctx.ClearCache()
+        # TODO: Reset the viewport
+        if reset_view:
+            ...
```
MustafaJafar: A word on resetting view

It seems the only true way for resetting the view is via destroying the current scene views and recreate them.
Resetting function can be relatively big to ensure copying the settings properly from the old scene view to the new scene view.
This function provides an example about how to achieve it.
```python
import labsopui
labsopui.reset_viewport.resetSceneViewers()
```

I'll try to find a shortcut or just have a minimal version of that function.

client/ayon_houdini/api/usd.py
```diff
@@ -425,3 +425,16 @@ def get_ayon_entity_uri_from_representation_context(context: dict) -> str:
             f"Received data: {response.data}"
         )
     return uris[0]["uri"]
+
+def clear_resolver_cache(reset_view=True):
+    try:
```
BigRoy: I think since we now are exposing this on default buttons - let's make it first check if we have the AYON USD resolver?

E.g. by starting with:
```python
try:
    from usdAssetResolver import AyonUsdResolver
except ModuleNotFoundError:
    return
```

client/ayon_houdini/startup/MainMenuCommon.xml
```diff
@@ -114,6 +114,15 @@ ayon_houdini.api.lib.update_houdini_vars_context_dialog()
 ]]></scriptCode>
             </scriptItem>
 
+            <separatorItem/>
+            <scriptItem id="clear_resplver_cache">
```
BigRoy:
```suggestion
            <scriptItem id="clear_resolver_cache">
```

client/ayon_houdini/api/usd.py
MustafaJafar: @BigRoy Back to this.

I've just seen this:
> SideFX Labs is a free, open-source, and artist-friendly toolset developed by SideFX.

And, the code actually is on GH -> [reset_viewport.py](https://github.com/sideeffects/SideFXLabs/blob/Development/scripts/python/labsopui/reset_viewport.py) <- so, I think it's safe to have it here?

MustafaJafar:
@BigRoy 
Tbh, after adding the last command to reload all layers, I feel users may be confused that clearing cache on an HDA will also clear the cache for other HDAs. I'm not sure if this what an artist may anticipate. 
But since the priority of this PR has been lowered, I believe having the current state is better than nothing.

client/ayon_houdini/api/usd.py
```diff
@@ -425,3 +425,23 @@ def get_ayon_entity_uri_from_representation_context(context: dict) -> str:
...
+    ctx = AyonUsdResolver.ResolverContext()
+    # Clear the resolver cache.
+    ctx.ClearCache()
+
+    # This can take time in heavy scenes.
+    # For future consideration, we may do
+    # this per layer via `hou.lop.reloadLayer("uri_path")`
+    if reload_all_files:
+        hou.lop.forceReloadAllFilesFromDisk(
+            reload_viewports=reload_viewports)
```
BigRoy: Would technically `hou.lop.reloadLayer(layer_path, recursive=True)` be better on the LOP themselves? 🤔 Sounds like a minor tweak to change it and may make the logic a bit better on that regard?

BigRoy (commented):
Nice. Should we still spend 15 minutes to try to get `reloadLayer` to do what we hope?

client/ayon_houdini/api/usd.py
MustafaJafar: > Would technically `hou.lop.reloadLayer(layer_path, recursive=True)` be better on the LOP themselves? 🤔 Sounds like a minor tweak to change it and may make the logic a bit better on that regard?

this is for the value in the `file` parameter if I'm not mistaken?

client/ayon_houdini/api/usd.py
```diff
@@ -425,3 +425,23 @@ def get_ayon_entity_uri_from_representation_context(context: dict) -> str:
...
+        reload_viewports=True
+    ):
+    try:
+        from usdAssetResolver import AyonUsdResolver
+    except ModuleNotFoundError:
+        return
+
+    ctx = AyonUsdResolver.ResolverContext()
+    # Clear the resolver cache.
+    ctx.ClearCache()
```
BigRoy, line 449: Even if resolver is not available - should we still force the houdini reload thing below?

client/ayon_houdini/api/usd.py
BigRoy: maybe - it may need to be the absolute layer identifier, or the _resolved_ path even? (meaning we may need to get path from the resolver BEFORE we clear its cache?)

Anyway, depends on what `reloadLayer` really takes as value here.

client/ayon_houdini/api/usd.py
MustafaJafar: Honestly, I don't have a strong opinion.  but, personally I expect the function to reload after clearing the cache but if cache can't be cleared, I'd expect a popup message that tells something like there's no cache and etc.

MustafaJafar:
@BigRoy okay I've a riddle. 

We can use the USD resolver for all Houdini contexts where we can load anything as URI whether it's an image or bgeo or usd. 
When clearing the cache we need to reload any loaded file.

Using this function is only for LOPs
```python
hou.lop.forceReloadAllFilesFromDisk(reload_viewports=reload_viewports)
```
Also, this one
```python
layer_path = hou.evalParm('./file')
hou.lop.reloadLayer(layer_path, recursive=True)
```

I think a better option for this matter is to chase the nodes referencing the file and press reload button on them. 
This should work regardless of the houdini context loading the file.

MustafaJafar:
with https://github.com/ynput/ayon-houdini/pull/388/commits/e94bddd187aa2547041cf9e077867ed9290793ac the function goes to the load node and press on reload button which works whether you are in LOPs or SOPs or any context as Houdini is consistent about `reload` parm name.

And if the function triggered from menu, it'll reload all LOPs.

client/ayon_houdini/api/usd.py
MustafaJafar: @BigRoy  A popup message like this. 

image

```suggestion
    try:
        from usdAssetResolver import AyonUsdResolver
    except ModuleNotFoundError:
        from ayon_core.tools.utils import show_message_dialog
        show_message_dialog(
            "Resolver Cache Not Cleared",
            "AYON USD Resolver was not found. Cache clearing was skipped.",
            "warning"
        )
        return

    ctx = AyonUsdResolver.ResolverContext()
    # Clear the resolver cache.
    ctx.ClearCache()
```

client/ayon_houdini/api/usd.py
```diff
@@ -425,3 +425,49 @@ def get_ayon_entity_uri_from_representation_context(context: dict) -> str:
...
+        # of all nodes that reference the file parameter.
+        ref_nodes_parm = node.parm('nodes_referencing_file')
+        if ref_nodes_parm:
+            for node_path in ref_nodes_parm.eval().split(" "):
+                reload_button = None
+                reload_button = hou.node(node_path).parm("reload")
+                if reload_button:
+                    reload_button.pressButton()
+        # If this triggered from node, we don't want to reload everything.
+        return
```
BigRoy: We may want to handle the case that actually both weren't match so we can catch bugs.



```suggestion
    # Reload loaded files by pressing reload button
    # either on the generic loader or the import asset or shot.
    if node:
        # if it's an import asset or shot, then it has a reload button.
        reload_button_parm = node.parm("reload")
        if reload_button_parm:
            reload_button_parm.pressButton()
            return
        # if not, then it's a generic lop node, which always has a list
        # of all nodes that reference the file parameter.
        refreshed = False
        ref_nodes_parm = node.parm('nodes_referencing_file')
        if ref_nodes_parm:
            for node_path in ref_nodes_parm.eval().split(" "):
                reload_button = None
                reload_button = hou.node(node_path).parm("reload")
                if reload_button:
                    reload_button.pressButton()
                    refreshed  = True
                    
        # If this triggered from node, we don't want to reload everything.
        if refreshed:
            return
```
And I'd argue we need to log a warning if no parms are found on the node to process, etc?

client/ayon_houdini/api/usd.py
```diff
@@ -425,3 +425,49 @@ def get_ayon_entity_uri_from_representation_context(context: dict) -> str:
             f"Received data: {response.data}"
         )
     return uris[0]["uri"]
+
+def clear_resolver_cache(
+        node=None,
+        reload_all_files=True,
```
BigRoy, line 431: When specifying both `node` and `reload_all_files=True` it doesn't actually reload all files, because it already returns on the `node` processing.

I'd actually say that whenever `reload_all_files` is True that it should always reload all. And that `node` is only relevant when `reload_all_files=False`. Otherwise the `reload_all_files` argument starts to turn rather confusing, right?

BigRoy (commented):
LGTM - just some small notes @MustafaJafar

client/ayon_houdini/api/usd.py
MustafaJafar: Should be achieved with https://github.com/ynput/ayon-houdini/pull/388/commits/82f13b64e9ed3e82dd2e55bba04d0234d327cc19

MustafaJafar:
@BigRoy I think this should be good to go. 

Also, what about https://github.com/ynput/ayon-houdini/pull/388#discussion_r3595741576 .
IMO, if a user want to just reload files without clearing cache, then they can just click `Reload Files` . 

I wonder if adding a popup message to inform users that resolver it not available.

client/ayon_houdini/api/usd.py
```diff
@@ -425,3 +425,49 @@ def get_ayon_entity_uri_from_representation_context(context: dict) -> str:
             f"Received data: {response.data}"
         )
     return uris[0]["uri"]
+
```
BigRoy, line 428:
```suggestion


```

BigRoy (approved):

client/ayon_houdini/api/usd.py
BigRoy: Without pop-up is best, because it does still clear the expression cache as well separate from this. ✅

PR #396, opened 2026-06-11
allow CollectFarmInstances to run for all families

### LINKED ISSUE #405: YN-0784: Houdini: Copernicus local render fails on Deadline plugin

When trying to submit a copernicus local render with Deadline addon enabled it will fail on Deadline submit render logic because it lacks some 'deadline' key. Essentially when you publish with Copernicus in Houdini with Render Target set as Local Machine Rendering, it still tries to run the HoudiniCacheSubmitDeadline class.

&nbsp;

However, it should be skipping it to begin with.

&nbsp;

I think the issue is that related publish families for cops are not found in the families filter of collect farm instances plugin [https://github.com/ynput/ayon-houdini/blob/develop/client/ayon_houdini/plugins/publish/collect_farm_instances.py](https://github.com/ynput/ayon-houdini/blob/develop/client/ayon_houdini/plugins/publish/collect_farm_instances.py) 

_No response_

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1509042802326245376)

_This issue was automatically created from AYON ticket [YN-0784]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=9fdf20605e5411f1bf0d333e7681eb41)

vincentullmann:
## Changelog Description
replaces the list of `families` on `CollectFarmInstances` with `*` to allow it running on all families (including `image_rop`)

image

`CollectFarmInstances` is supposed to remove  `publish.hou` form the families when running local, but it did not run for  `image_rop` families
This later down the line causes `HoudiniCacheSubmitDeadline` to trigger, even for local products, and fail because `CollectDeadlineServerFromInstance` is using [different logic](https://github.com/vincentullmann/ayon-deadline/blob/develop/client/ayon_deadline/plugins/publish/global/collect_deadline_server_from_instance.py#L38-L40) to differentiate between local and farm products

:warning: I've only done limited testing on this. I am not sure if there are any families that do need to be excluded, in which case maintaining a long list of "allowed families" might make more sense.

## Testing notes:
1. create a cops network with some test nodes (eg.: a `constant`)
2. open the AYON Creator and create a "Composite (Copernicus)" product
3. set render target to "local" render
4. publish

failed publish log
[publish-report-260611-17-03.json](https://github.com/user-attachments/files/28846321/publish-report-260611-17-03.json)

### DIFF
--- a/client/ayon_houdini/plugins/publish/collect_farm_instances.py
+++ b/client/ayon_houdini/plugins/publish/collect_farm_instances.py
@@ -6,7 +6,6 @@ class CollectFarmInstances(plugin.HoudiniInstancePlugin):
     """Collect instances for farm render."""
 
     order = pyblish.api.CollectorOrder - 0.49
-    families = ["*"]
 
     targets = ["local", "remote"]
     label = "Collect farm instances"
--- a/client/ayon_houdini/plugins/publish/collect_farm_instances.py
+++ b/client/ayon_houdini/plugins/publish/collect_farm_instances.py
@@ -6,16 +6,7 @@ class CollectFarmInstances(plugin.HoudiniInstancePlugin):
     """Collect instances for farm render."""
 
     order = pyblish.api.CollectorOrder - 0.49
-    families = ["mantra_rop",
-                "karma_rop",
-                "redshift_rop",
-                "arnold_rop",
-                "vray_rop",
-                "usdrender",
-                "ass","pointcache", "redshiftproxy",
-                "vdbcache", "model", "staticMesh",
-                "rop.opengl", "usdrop", "camera",
-                "apex"]
+    families = ["*"]
 
     targets = ["local", "remote"]
     label = "Collect farm instances"


MustafaJafar (approved):
I think it'll be fine to include all families.
I tested rendering and publishing different products and it seems okay.

client/ayon_houdini/plugins/publish/collect_farm_instances.py
```diff
@@ -6,16 +6,7 @@ class CollectFarmInstances(plugin.HoudiniInstancePlugin):
...
-                "karma_rop",
-                "redshift_rop",
-                "arnold_rop",
-                "vray_rop",
-                "usdrender",
-                "ass","pointcache", "redshiftproxy",
-                "vdbcache", "model", "staticMesh",
-                "rop.opengl", "usdrop", "camera",
-                "apex"]
+    families = ["*"]
```
moonyuet: Have you tried the suggestion below with `workfile` product and see if it works?
```suggestion
```

vincentullmann: did some tests and it seems to work fine.

I added a print statement in my local copy to check and it seems "render_target" is set to `None` for `workfiles
```
INFO: Render target: None
DEBUG: Render on farm is disabled. Skipping farm collecting.
```

BigRoy (approved):
Better than the bulk of families admittedly - so if this works fine, let's do it.

Issue #399, opened 2026-06-16
Loader: Add rig product type to bgeo loader


semx2a:
**Is your feature request related to a problem? Please describe.**
Ability to load rig products into a Houdini scene.

**Describe the solution you'd like**
Add "rig" product type to load_bgeo.py's product types list.

MustafaJafar:
The Pr was merged already #400

PR #400, opened 2026-06-16
feat: Add 'rig' to product base types in load_bgeo

semx2a:
## Changelog Description
Ability to load rigs into Houdini

## Issue
#399

### DIFF
--- a/client/ayon_houdini/plugins/load/load_bgeo.py
+++ b/client/ayon_houdini/plugins/load/load_bgeo.py
@@ -12,7 +12,7 @@ class BgeoLoader(plugin.HoudiniLoader):
     """Load bgeo files to Houdini."""
 
     label = "Load bgeo"
-    product_base_types = {"model", "pointcache", "bgeo"}
+    product_base_types = {"model", "pointcache", "bgeo", "rig"}
     product_types = product_base_types
     representations = {"*"}
     extensions = {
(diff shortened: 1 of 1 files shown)


client/ayon_houdini/plugins/load/load_bgeo.py
```diff
@@ -12,7 +12,7 @@ class BgeoLoader(plugin.HoudiniLoader):
     """Load bgeo files to Houdini."""
 
     label = "Load bgeo"
-    product_base_types = {"model", "pointcache", "bgeo"}
+    product_base_types = {"model", "pointcache", "bgeo", "rig"}
```
BigRoy: Probably being able to load any bgeo file makes just as much sense?

```suggestion
    product_base_types = {"*"}
```

BigRoy (commented):
@MustafaJafar does this still make sense next to https://github.com/ynput/ayon-houdini/pull/384 ? 

Probably having just the ability to load directly the `bgeo` makes perfect sense.

I'm fine with the change, but may as well just allow loading any bgeo, right?

client/ayon_houdini/plugins/load/load_bgeo.py
MustafaJafar: ~~if we do this, then we should filter the representations instead, correct?~~ 
oh, we already filter by extensions.

This makes much sense.

BigRoy: No, because it already filters by `extensions` - as such, the BgeoLoader can load _anything_ that happens to have a valid extension - regardless of representation name or product types.

semx2a:
Hi there, I've added the suggestion to glob product_base_types to load all rig types. Are there any additional blockers that prevent the merge of this PR?

BigRoy (approved):
Sorry, I thought it was pending review by Mustafa - but seems he already marked the comment resolved perhaps?

PR #402, opened 2026-06-19
Expose representation filter to the load asset and shot HDAs

### LINKED ISSUE #385: YN-0708: Load Asset/Shot LOPs do not filter representations by default

Products with representations like 'camera_abc' can't be selected via the Load Asset LOP and Load Shot LOP because the internal generic loader seems to have filtering enabled, without that filtering exposed to the parent node. I'd like to propose to either disable the filtering altogether there OR at least add a way to allow disabling the filtering.

&nbsp;

The fact that they are not listed is just confusing to the user, because you can't see the filtering is being applied currently unless you dive into the (locked HDAs) graphs, etc 

### Additional context:

_This issue was automatically created from AYON ticket [YN-0708]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=9d9f4f50453611f1b45b2b1b3be889ad)

MustafaJafar:
## Changelog Description
- Add Extension filter to generic loaders
- Expose extension filter to the load asset and shot HDAs

image

## Testing notes:
1. Create load asset/shot in solaris
2. enable/disable the extension filter

Resolve #385


MustafaJafar:
The fact we have this filter enabled by default is originating from the things that reference or sublayer nodes can load in solaris. 
It's worth mentioning that representations might not always mirror the file extension e.g. representation name `camera_abc`

BigRoy:
> The fact we have this filter enabled by default is originating from the things that reference or sublayer nodes can load in solaris. It's worth mentioning that representations might not always mirror the file extension e.g. representation name `camera_abc`

Exactly, looking back I actually think it makes more sense to have an extensions filter, and have the order of the entries be the "priority" for which may be the default extension or representation to pick? But then what if you have two representations of the same extension where one is a proxy, you may want to prioritze the main one. Etc, suddenly it becomes a whole mess to configure this in sensible way. Which makes me wonder the whole filtering is a good idea to begin with and isn't just making things more complicated 🤔 Anyway, it's really only for the dropdowns and for ux for the artists.

BigRoy:
@MustafaJafar should we add the extensions filtering? 🤔 or is there a reason still not to?

MustafaJafar:
> @MustafaJafar should we add the extensions filtering? 🤔 or is there a reason still not to?

I believe there is no strong reason. it's just the way it was developed in the first place. 
The intent was to filter out the things that will fail in USD which can be done in two ways  
- representation filter
- extension filter

we were relying on python api so much so the most easy way was to continue on the same pattern and add representation filtering as follows

https://github.com/ynput/ayon-houdini/blob/d06d6190daace92f6b1e49db61208f933e24387d/client/ayon_houdini/api/hda_utils.py#L949-L960


We can extend this by extension filtering by adding `fields=["attrib.extension"]` 
But, I really wonder why its value is non.
```
{
  "data": {
    "projects": {
      "edges": [
        {
          "node": {
            "version": {
              "id": "531347337bca11f19647431bc872748a",
              "representations": {
                "edges": [
                  {
                    "node": {
                      "id": "531bb4a77bca11f18ff6431bc872748a",
                      "attrib": {
                        "extension": null
                      },
                      "name": "usd"
                    }
                  }
                ]
              }
            }
          }
        }
      ]
    }
  }
}
```

which looks like a bug.

MustafaJafar:
@BigRoy 
I've added the extension filter and exposed it instead of representation filter.
image
on generic loaders, we can see both filters.
image

BigRoy (changes requested):
With a Load Shot and Load Asset LOP it's still filtering out non-usd and non-abc files from the representation pop-up menu. E.g. if you have a version that has e.g. a `.json` file it's not listed there.

Whenever I unlock the HDA, then suddenly it does work 🤷‍♂️ 

Locked HDA:

Image

Then:

Image

Then still doesn't work:

Image

Dive into the subnet, select the Generic Loader LOP. Click on Representation there:

Image

And now suddenly outside of the subnet it also works:

Image

Tested in Hou22. @MustafaJafar any idea how to solve?

I think we should likely avoid setting the default extension filters inside the Generic Loader LOPs intside the Load Asset LOP and the Load Shot LOP. They are disabled internally anyway, and there's no way to enable the... but somehow here they do behave as if they were enabled?

MustafaJafar:
> With a Load Shot and Load Asset LOP it's still filtering out non-usd and non-abc files from the representation pop-up menu. E.g. if you have a version that has e.g. a `.json` file it's not listed there.
> 
> Whenever I unlock the HDA, then suddenly it does work 🤷‍♂️
>
> Tested in Hou22. @MustafaJafar any idea how to solve?

I know where is the origin of this.. 
When I removed the representation filter parm I didn't clear its default value first..... 🫤 
When you unlock the HDA, the parameters are reset to the defaults in the generic loaders.


I'll put them back and remove the default value then remove the parm again.
image

MustafaJafar:
> With a Load Shot and Load Asset LOP it's still filtering out non-usd and non-abc files from the representation pop-up menu. E.g. if you have a version that has e.g. a `.json` file it's not listed there.

I still can reproduce the reported issue.
But, I don't need to unlock it. I just double click the HDA and go inside it and select the generic loader inside.

https://github.com/user-attachments/assets/4b233e39-1a64-4522-9866-0b88ed7666c2

----

The logic doesn't check the toggle, but detects its effect which is the `representation_filter` and `extension_filter` are disabled. but, as shown in the video above, these parameters are always considered not disabled until I go inside and click on the generic loader inside. 
And, since I put a default value for the `extension_filter` which shouldn't take an effect but the filter works.
tbh, this doesn't make any sense!

image

---

Update I found the answer [here](https://www.sidefx.com/forum/topic/95657/?page=1#post-420524) let me quote:
> The issue is that such things as disabled/hidden are UI states and not guaranteed to be accurate at any given time. If the parameter interface hasn't had to display that node yet then its state is going to be unclear. You'll want to call `hou.OpNode.updateParmStates()` before checking the value which should give you your expected result.

That's explains why it worked when I clicked the source generic loader node.

BigRoy (approved):

PR #409, opened 2026-07-16
Add python3.13libs to support Houdini 22

BigRoy:
## Changelog Description

Add python3.13libs to support Houdini 22

Without this Houdini 22 would not trigger the startup scripts.

## Testing notes:

1. Houdini 22 should work.


MustafaJafar (approved):
So cool!

PR #410, opened 2026-07-17
combine pythonX.Ylibs into a single folder

vincentullmann:
## Changelog Description
combines all the `pythonX.Ylibs` folders into just `scripts/python`

houdini checks both `pythonX.Ylibs` and `python`. the versioned one can be useful if we need to make changes based on the what python version is used, but this is not the case (and hopefully wont be).
So we can combine them for simplicity and future proofing.

## Testing notes:
1. launch any houdini version


BigRoy:
What!? Does this work? Thank you, if it does.

@MustafaJafar can you confirm a few Houdini versions?

vincentullmann:
been working for me since houdini 15.

forum post to confirm:
https://www.sidefx.com/forum/topic/99638/?page=1#post-436687

not 100% sure but I think the `scripts/python` was used to be the default back in python2 times. iirc the split into `pythonX.Ylibs` started when people had to support python 2 and 3 in parallel during the transition period

vincentullmann:
did a bit of digging:
image

it was listed as a "backwards compatibility" feature back in h14
Screenshot_2026-07-17T10:23:40
https://web.archive.org/web/20150920072309/https://www.sidefx.com/docs/houdini14.0/hom/independent

but it hasn't been removed ove the past 10 years and instead added as a general tip, so imho it should be safe to rely on
image
https://www.sidefx.com/docs/houdini/hom/locations.html#general-tips

MustafaJafar:
Could we create an issue and link it to this PR?

MustafaJafar (approved):
> combines all the `pythonX.Ylibs` folders into just `scripts/python`

I don't know much about history of this folder. but I know it's just supported. 

---

Yes, this PR worked on my side I tested it with 5 Houdini versions from 19.5 to 22.0

BigRoy:
No need for issue, it's a simple fix that changes no behavior. In a way it's code cosmetics (until a new release of Houdini comes along with a differen Python.). This really is the best fix.

I'll double check a thing next week and then merge it.

Ahmed-Hindy:
@BigRoy I have doing this in non AYON pipeline for a couple of years now py3.7 tp py3.11 with absolutely no drawbacks. I never understood why AYON splits it in the first place.

BigRoy:
> I have doing this in non AYON pipeline for a couple of years now py3.7 tp py3.11 with absolutely no drawbacks. I never understood why AYON splits it in the first place.

Thanks - it all boiled down to the most important factor, human beings - or really, me. I just didn't know about it.

BigRoy (approved):
Thanks so much!

PR #412, opened 2026-07-31
Remove `CollectFarmCacheFamily` plugin

MustafaJafar:
## Changelog Description
- Remove `CollectFarmCacheFamily` plugin

This is a follow up PR to https://github.com/ynput/ayon-houdini/pull/317 which added the removal of `publish.hou` which is done at the same order `pyblish.api.CollectorOrder - 0.49`. 
Which makes `CollectFarmCacheFamily` redundant. 

## Testing notes:
1. Publishing cache to redner farm should work as before.

Resolve https://github.com/ynput/ayon-deadline/issues/173

### DIFF
--- a/client/ayon_houdini/plugins/publish/collect_cache_farm.py
+++ b/client/ayon_houdini/plugins/publish/collect_cache_farm.py
@@ -7,24 +7,6 @@ from ayon_houdini.api import (
 )
 
 
-class CollectFarmCacheFamily(plugin.HoudiniInstancePlugin):
-    """Collect publish.hou family for caching on farm as early as possible."""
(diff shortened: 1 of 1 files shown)


MustafaJafar:
One thing to note. `CollectFarmInstances` depends on the render target unlike the removed on `CollectFarmCacheFamily` depends on the `farm` instance data. 

Before implementing render targets, it was common to have a toggle which drives the `farm` instance data. and I believe we no longer. we always have render targets.

BigRoy (approved):
I think this is correct.

Did this work as intended on your test runs?

MustafaJafar:
> Did this work as intended on your test runs?

Everything looks as usual.

PR #413, opened 2026-08-05
disable CPU (XPU) toggle to set KARMA_XPU_DISABLE_EMBREE_DEVICE on renders

miravassor:
## Add Karma XPU "Disable CPU" toggle

Adds a `Disable CPU (XPU)` toggle to the Karma ROP and USD Render creators.

When enabled, `KARMA_XPU_DISABLE_EMBREE_DEVICE=1` is added to the farm job environment, so Karma XPU renders on GPU devices only.

Only applies to farm renders with the Karma XPU engine/renderer (ignored otherwise).

We observed some great performance enhancement on some scenes when disabling the CPU usage on our farm XPU renders.


miravassor:
Hello @MustafaJafar, do you have the time to look into this? Our studio and I believe others could really benefit from this (up to 2x improvement on render time)!
Thanks

PR #417, opened 2026-08-24
Implementation the new workfile buiilder in houdini

### LINKED ISSUE #416: Houdini: Implementation of new workfile template system
Copied from https://github.com/ynput/ayon-core/issues/1040

If we want more explicit control, we may need to clarify the exact settings configuration that makes sense:

- Workfile Template (enum):
  - Disabled
  - Always apply to empty scene (act as 'new file' template)
  - Apply on application launch if no workfile exists yet
- Save first workfile if it does not exist yet

 To me, the 'saving of first workfile' should likely need to be outside of the workfile template loading itself. Because it may very well be that someone does want to directly 'save' their filename on launch just so they can directly press CTRL+S if they start working but have no templates configured to start from.

_No response_

_No response_

moonyuet:
## Changelog Description
This PR is to implement the new workflow of workfile builder in Houdini
Resolve https://github.com/ynput/ayon-houdini/issues/416

Links https://github.com/ynput/ayon-core/issues/1040
## Testing notes:
1. Tweak `ayon+settings://houdini/templated_workfile_build/profiles/0`
2. Build workfile

### DIFF
--- a/client/ayon_houdini/api/lib.py
+++ b/client/ayon_houdini/api/lib.py
@@ -1563,14 +1563,17 @@ def prompt_reset_context():
 
 def start_workfile_template_builder():
     from .workfile_template_builder import (
-        build_workfile_template
+        trigger_on_new_file,
+        trigger_on_app_launch,
     )
-
-    log.info("Starting workfile template builder...")
-    try:
-        build_workfile_template(workfile_creation_enabled=True)
-    except TemplateProfileNotFound:
-        log.warning("Template profile not found. Skipping...")
+    host = registered_host()
+    if host.hou_initialized:
+        log.info("Starting workfile template builder...")
+        trigger_on_new_file()
+    else:
+        host.hou_initialized = True
+        if not host.get_current_workfile():
+            trigger_on_app_launch()
 
 
 def show_node_parmeditor(node):
--- a/client/ayon_houdini/api/lib.py
+++ b/client/ayon_houdini/api/lib.py
@@ -24,9 +24,6 @@ from ayon_core.pipeline import (
 from ayon_core.pipeline.create import CreateContext
 from ayon_core.pipeline.template_data import get_template_data
 from ayon_core.pipeline.context_tools import get_current_task_entity
-from ayon_core.pipeline.workfile.workfile_template_builder import (
-    TemplateProfileNotFound
-)
 from ayon_core.tools.utils import PopupUpdateKeys, SimplePopup
 from ayon_core.tools.utils.host_tools import get_tool_by_name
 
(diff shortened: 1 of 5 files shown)


moonyuet:
@BigRoy I hit some issues when building the houdini workfile template with different license type. Should we consider this as part of the issue as I can't do file-> new due to the contradictory of the different license types for built houdini workfile from that of the software we used.

MustafaJafar:
> @BigRoy I hit some issues when building the houdini workfile template with different license type. Should we consider this as part of the issue as I can't do file-> new due to the contradictory of the different license types for built houdini workfile from that of the software we used.

Does this mean the template is a `hip` file while the current workfile is `hiplc` ? or vice verse

moonyuet:
> > @BigRoy I hit some issues when building the houdini workfile template with different license type. Should we consider this as part of the issue as I can't do file-> new due to the contradictory of the different license types for built houdini workfile from that of the software we used.
> 
> Does this mean the template is a `hip` file while the current workfile is `hiplc` ? or vice verse

yes. This scenario is something we might need to consider for just in case there are some users want to build the files between different license types.

BigRoy:
Please elaborate the exact reproducable. What license types? in what scenarios does it fail? what's the resulting behavior, etc.

Preferably I'd like it to always work, but if the non commercial or lesser commercial licenses of Houdini have a "side effect" 🥁  bug I guess it's ok to just document that. However, I'm curious as to what is making it fail?

MustafaJafar (approved):
Yep works. 
Thank you.

moonyuet:
> Please elaborate the exact reproducable. What license types? in what scenarios does it fail? what's the resulting behavior, etc.
> 
> Preferably I'd like it to always work, but if the non commercial or lesser commercial licenses of Houdini have a "side effect" 🥁 bug I guess it's ok to just document that. However, I'm curious as to what is making it fail?

`hiplc` is the houdini file built in Houdini Indie, while hip is the houdini file built in Houdini Fx. There is also hipnc built in Houdini apprentice aka non commerical version. If we built the hiplc files in Houdini Fx, it will downgrade the license. You will notice this when you start a new scene through File-> New and you can notice the menu below.
It doesn't fail but you can't manage to launch the new scene unless you closed houdini and turned off your workfile build.
image

MustafaJafar:
> `hiplc` is the houdini file built in Houdini Indie, while hip is the houdini file built in Houdini Fx. There is also hipnc built in Houdini apprentice aka non commerical version. If we built the hiplc files in Houdini Fx, it will downgrade the license. You will notice this when you start a new scene through File-> New and you can notice the menu below.
> It doesn't fail but you can't manage to launch the new scene unless you closed houdini and turned off your workfile build.

I believe this is an expected behavior. Houdini users should be aware that mixing and matching hip files with different licenses.

BigRoy:
I see, everything works - Houdini just blocks you out because it's affecting this... and that's because the template file is `hipnc` or `hiplc` in that case, right? If so, I guess that's somewhat acceptable... you just shouldn't assign a template file then that enforces less permissions?

BigRoy (changes requested):
I believe this PR may need to update `package.py` to still require the newer core?

BigRoy (approved):
Mustafa tested it, package.py requirement is there now. 👍

PR #418, opened 2026-08-28
Add `ValidateFrameTokenUSDRender` plugin

### LINKED ISSUE #328: YN-0275: rendering custom frames

It seems custom frame range is not working correctly in Houdini. Render logs provided in the client thread. 

_No response_

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1445471105446973662/1445489917231628361)

_This issue was automatically created from AYON ticket [YN-0275]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=c2bc7010cff211f094c26d096f04fe4c)

MustafaJafar:
## Changelog Description
Add `ValidateFrameTokenUSDRender` plugin

Untitled

## Testing notes:
1. create a usd render 
2. change custom frame in publisher
3. publisher should fail and you can try the repair action

Resolve #328

### DIFF
--- a/client/ayon_houdini/plugins/publish/validate_frame_token_usd_render.py
+++ b/client/ayon_houdini/plugins/publish/validate_frame_token_usd_render.py
@@ -78,5 +78,9 @@ class ValidateFrameTokenUSDRender(plugin.HoudiniInstancePlugin):
             # Already fixed
             return
 
+        import os 
+
         rop_node = hou.node(instance.data["instance_node"])
-        rop_node.parm("lopoutput").set("__render__.$F4.usd")
+        lopoutput = rop_node.parm("lopoutput").eval()
+        path, ext = os.path.splitext(lopoutput)
+        rop_node.parm("lopoutput").set(f"{path}.$F4{ext}")
--- a/client/ayon_houdini/plugins/publish/validate_frame_token_usd_render.py
+++ b/client/ayon_houdini/plugins/publish/validate_frame_token_usd_render.py
@@ -1,0 +1,82 @@
+import hou
+
+import pyblish.api
+
+from ayon_core.pipeline import PublishValidationError
+from ayon_core.pipeline.publish import RepairAction
+from ayon_houdini.api import plugin
+from ayon_houdini.api.action import SelectInvalidAction
+
+
+class FixParameterAction(RepairAction):
+    label = "Add $F4 to 'lopoutput'"
+
+
+class ValidateFrameTokenUSDRender(plugin.HoudiniInstancePlugin):
+    """Validate if the unexpanded string contains the frame ('$F') token.
+
+    This validator will *only* check the output parameter of the node if
+    the "Use Custom Frames" attribute is set.
+
+    Rules:
+        If "Use Custom Frames" attribute is set in publisher
+        it is mandatory to have the frame token - '$F4' or similar -
+        to ensure that each frame gets written.
+        If this is not the case you may face black render
+        as the output USD file may not cover the whole range.
+
+    Examples:
+        Good: '__render__.$F4.usd'
+        Bad: '__render__.usd'
+
+    """
+
+    order = pyblish.api.ValidatorOrder
+    label = "Validate Frame Token (USD Render)"
+    families = ["usdrender"]
+    actions = [FixParameterAction, SelectInvalidAction]
+
+    def process(self, instance):
+        if not instance.data.get("farm"):
+            self.log.debug("Not a farm instance, skipping.")
+            return
+
+        if "deadline.submit.publish.job" not in instance.data["families"]:
+            self.log.debug(
+                "Should only checked if submitting to deadline, skipping.")
+            return
+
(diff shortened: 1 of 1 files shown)


client/ayon_houdini/plugins/publish/validate_frame_token_usd_render.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def process(self, instance):
+        if not instance.data.get("farm"):
+            self.log.debug("Not a farm instance, skipping.")
+            return
+
+        if "deadline.submit.publish.job" not in instance.data["families"]:
+            self.log.debug(
+                "Should only checked if submitting to deadline, skipping.")
+            return
```
BigRoy, line 47: If Deadline-specific; should this be inside `ayon-deadline`? 🤔

client/ayon_houdini/plugins/publish/validate_frame_token_usd_render.py
```diff
@@ -0,0 +1,82 @@
...
+            return [node]
+
+    @classmethod
+    def repair(cls, instance):
+        if not cls.get_invalid(instance):
+            # Already fixed
+            return
+
+        rop_node = hou.node(instance.data["instance_node"])
+        rop_node.parm("lopoutput").set("__render__.$F4.usd")
```
BigRoy: Should we take current filepath, split extension, add `.$F4` and then add extension back? Just so that if for whatever reason someone is using customized paths here - our repair action somewhat works as well?

client/ayon_houdini/plugins/publish/validate_frame_token_usd_render.py
```diff
@@ -0,0 +1,82 @@
...
+                "No frame token found in the USD Export Output File parm in"
+                f" '{invalid[0].path()}'"
+            )
+
+    @classmethod
+    def get_invalid(cls, instance):
+
+        node = hou.node(instance.data["instance_node"])
+        output_parm = node.parm("lopoutput")
+        unexpanded_str = output_parm.unexpandedString()
```
BigRoy, line 70: Note that this may fail with `OperationFailed` in Houdini if _for whatever reason_ some crazy person put an expression or keyframes on the attribute :)

> Suppose a string parameter contains keyframes. In this situation, Houdini will not attempt to do string expansion on the parameter’s value, so calling this method will raise hou.OperationFailed. Instead of calling this method, you can call expression() to access the first Keyframe’s expression. If there are multiple keyframes, you can call keyframes() to get a list of [hou.StringKeyframe](https://www.sidefx.com/docs/houdini/hom/hou/StringKeyframe.html) objects and call expression() on those objects to retrieve the expression. [src](https://www.sidefx.com/docs/houdini/hom/hou/Parm.html#value)

client/ayon_houdini/plugins/publish/validate_frame_token_usd_render.py
MustafaJafar: Let me spill some random thoughts.

I thought we put the missing frames info in instance data but no we set them directly in the deadline job info. 

I put that deadline check to be sure that we have the job info where I can know whether or not we are setting custom frames in DL jobs.

checking if farm instance cannot be enough as we only want to do it when we are setting custom frames. 

I wonder if checking the `publish_attributes` can be an alternative.

client/ayon_houdini/plugins/publish/validate_frame_token_usd_render.py
MustafaJafar: This looks familiar. yeah we mentioned this topic [here](https://github.com/ynput/ayon-houdini/blob/5c6d85bd8e40667cd8b9a0f87ac2182060a2ed0d/client/ayon_houdini/api/lib.py#L1630-L1637) but in the opposite situation. 

For reference, we have the same bug you mentioned in the here 
https://github.com/ynput/ayon-houdini/blob/5c6d85bd8e40667cd8b9a0f87ac2182060a2ed0d/client/ayon_houdini/plugins/publish/validate_frame_token.py#L49

client/ayon_houdini/plugins/publish/validate_frame_token_usd_render.py
BigRoy: > I wonder if checking the publish_attributes can be an alternative.

Probably not, because that one then also comes from deadline? or is this coming through core? if so, **then** we should not make this deadline specific, etc.

MustafaJafar: It's coming from deadline. 
https://github.com/ynput/ayon-deadline/blob/33dfc6ff914c70cffddd026fcf2ef7a9f402e3aa/client/ayon_deadline/plugins/publish/global/collect_jobinfo.py#L272-L285

I'll move this to deadline.

MustafaJafar:
Moving this to Deadline repo https://github.com/ynput/ayon-deadline/pull/276

PR #420, opened 2026-09-09
USD Render - use stage metadata instead of hardcoded rendersettings path

### LINKED ISSUE #419: USD Render Houdini 22 incorect default render settings

## Current Behavior
In case a render settings path is not set on the `usd render rop`, Ayon falls back to `/Render/rendersettings`. 
In Houdini 22 they renamed it to "/Render/$OS" and by default it initializes as `/Render/karmarendersettings`, causing no files to be collected if no render settings path is set.

## Expected Behavior
A better solution is to use the stage metadata for the default render settings prim instead of a hardcoded path. Husk also respects this.
Karmarendersettings already sets the render settings prim as default automatically when a it is created.

### Steps To Reproduce:

1. Submit karma render in H22 with no rendersettings path configured on usd render rop
2. Validators should fail `no render products found`

Phantom138:
## Changelog Description
Uses stage metadata as a fallback instead of hardcoded `/Render/rendersettings` path
Fixes #419 

## Testing notes:
1. Submit usd render with no render settings path set
2. Files should be collected regardless of the rendersettings prim name (as long as it's set as default on the layer)

### DIFF
--- a/client/ayon_houdini/api/usd.py
+++ b/client/ayon_houdini/api/usd.py
@@ -341,7 +341,7 @@ def get_usd_render_rop_rendersettings(rop_node, stage=None, logger=None):
     path = rop_node.evalParm("rendersettings")
     if not path:
         # Default behavior
-        path = "/Render/rendersettings"
+        path = stage.GetMetadata("renderSettingsPrimPath")
 
     prim = stage.GetPrimAtPath(path)
     if not prim:
--- a/client/ayon_houdini/api/usd.py
+++ b/client/ayon_houdini/api/usd.py
@@ -340,7 +340,7 @@ def get_usd_render_rop_rendersettings(rop_node, stage=None, logger=None):
 
     path = (
         rop_node.evalParm("rendersettings")
-        or stage.GetMetadata("renderSettingsPrimPath") 
+        or stage.GetMetadata("renderSettingsPrimPath")
         or "/Render/rendersettings"
     )
 
--- a/client/ayon_houdini/api/usd.py
+++ b/client/ayon_houdini/api/usd.py
@@ -338,10 +338,11 @@ def get_usd_render_rop_rendersettings(rop_node, stage=None, logger=None):
         lop_node = get_usd_rop_loppath(rop_node)
         stage = lop_node.stage()
 
-    path = rop_node.evalParm("rendersettings")
-    if not path:
-        # Default behavior
-        path = stage.GetMetadata("renderSettingsPrimPath")
+    path = (
+        rop_node.evalParm("rendersettings")
+        or stage.GetMetadata("renderSettingsPrimPath") 
+        or "/Render/rendersettings"
+    )
 
     prim = stage.GetPrimAtPath(path)
     if not prim:
--- a/client/ayon_houdini/plugins/publish/validate_render_products.py
+++ b/client/ayon_houdini/plugins/publish/validate_render_products.py
@@ -53,8 +53,10 @@ class ValidateUsdRenderProducts(plugin.HoudiniInstancePlugin):
 
         if not instance.data.get("files", []):
             node = hou.node(node_path)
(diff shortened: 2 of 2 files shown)


client/ayon_houdini/plugins/publish/validate_render_products.py
```diff
@@ -53,8 +53,10 @@ def process(self, instance):
 
         if not instance.data.get("files", []):
             node = hou.node(node_path)
+            lop_stage = node.parm("loppath").evalAsNode().stage()
```
BigRoy:
```suggestion
            lop_stage = instance.data["stage"]
```

client/ayon_houdini/plugins/publish/validate_render_products.py
```diff
@@ -53,8 +53,10 @@ def process(self, instance):
...
+            
             rendersettings_path = (
-                node.evalParm("rendersettings") or "/Render/rendersettings"
+                node.evalParm("rendersettings") or lop_stage.GetMetadata("renderSettingsPrimPath")
```
BigRoy: This should still fall back to `"/Render/rendersettings"` if the stage metadata is not there.
```suggestion
                node.evalParm("rendersettings") 
                or lop_stage.GetMetadata("renderSettingsPrimPath")
                or "/Render/rendersettings"
```

Also, just now learned that this is default behavior of USD apparently: https://openusd.org/release/api/usd_render_page_front.html

But we'll need the fallback, yes - because Karma Render Settings LOP node does set it, but a regular Render Settings node does not.

client/ayon_houdini/api/usd.py
```diff
@@ -341,7 +341,7 @@ def get_usd_render_rop_rendersettings(rop_node, stage=None, logger=None):
     path = rop_node.evalParm("rendersettings")
     if not path:
         # Default behavior
-        path = "/Render/rendersettings"
+        path = stage.GetMetadata("renderSettingsPrimPath")
```
BigRoy: Should fallback to `/Render/rendersettings` if metadata is not set.
```suggestion
    path: str = (
        rop_node.evalParm("rendersettings") 
        or stage.GetMetadata("renderSettingsPrimPath")
        or "/Render/rendersettings"
    )
```

BigRoy (changes requested):
Thanks, looking good - just need to add some fallbacks and use the stage from `instance.data` instead to avoid requerying it.

Phantom138:
Hi, take a look at the new commit, added the suggested changes.
Also added an extra check if the prim actually exists. That way it makes it a bit easier to debug in case an artist gets that validation error.

client/ayon_houdini/plugins/publish/validate_render_products.py
```diff
@@ -53,9 +55,16 @@ def process(self, instance):
...
             rendersettings_path = (
-                node.evalParm("rendersettings") or "/Render/rendersettings"
+                node.evalParm("rendersettings") 
+                or instance.data["stage"].GetMetadata("renderSettingsPrimPath")
+                or "/Render/rendersettings"
             )
+
+            if not instance.data["stage"].GetPrimAtPath(rendersettings_path):
+                self.log.warning(f"No render settings primitive found at: {rendersettings_path}")
+
```
BigRoy, line 62:
```suggestion
            stage = instance.data["stage"]
            rendersettings_path: str = (
                node.evalParm("rendersettings") 
                or stage.GetMetadata("renderSettingsPrimPath")
                or "/Render/rendersettings"
            )

            if not stage.GetPrimAtPath(rendersettings_path):
                self.log.warning(f"No render settings primitive found at: {rendersettings_path}")
```

BigRoy: I hate Github sometimes - haha.

BigRoy (approved):
Works for me.

PR #421, opened 2026-09-09
fix #372: allow all camera types in validate scene review

### LINKED ISSUE #372: allow different camera types in `ValidateSceneReview`

## Current Behaviour
`ValidateSceneReview` only allows `{"cam", "lopimportcam"}` 
see: https://github.com/ynput/ayon-houdini/blob/develop/client/ayon_houdini/plugins/publish/validate_scene_review.py#L56-L58

## Expected Behaviour
Allow all nodes that can be used as cameras.
This would include nodes such as "Switcher", "VR Camera" as well as in house camera nodes

a standard method to list all object camera nodes is:
```py
hou.root().recursiveGlob('*', hou.nodeTypeFilter.ObjCamera)
```
this lists the same cameras as the viewpoprt dropdown

Image

We'd still need to find a solution for Lop Cameras, which are unfortunately not included here 

### Steps To Reproduce:

- create `Switcher` or `VR Camera` node
- create a "Review" using this camera
- validate

### Version

Linux - Rocky9
Linux - Arch Linux

### Relevant log output:

Image

```shell
  File "/home/v/.local/share/AYON/dependency_packages/ayon_2602231723_linux-rocky9.zip/dependencies/pyblish/plugin.py", line 528, in __explicit_process
    runner(*args)
  File "/home/v/.local/share/AYON/addons/houdini_0.9.5/ayon_houdini/plugins/publish/validate_scene_review.py", line 38, in process
    raise PublishValidationError(
ayon_core.pipeline.publish.publish_plugins.PublishValidationError: Camera path is not a camera: '/obj/cam_switcher' (type: switcher)
```

vincentullmann:
## Changelog Description
Allow all camera types in `ValidateSceneReview`

added a small helper method `node_matches_filter` to check if a given node matches a given `hou.nodeTypeFilter` and utilise that to validate if a node matches the `hou.nodeTypeFilter.ObjCamera` filter

## Testing notes:
1. create a `Review (opengl)` product
2. create a regular camera, a "special camera" (eg.: "Switcher") and a non camera (eg.: "geo")
3. set the OpenGL rop to use each of these
4. validate

validation should pass for the Camera and swticher and fail for the geo

Example failed message: (same as before)
> Camera path is not a camera: '/obj/geo4' (type: geo)

test script for `node_matches_filter`

```py
import hou

from ayon_houdini.api.lib import node_matches_filter


cam = hou.node("/obj/").createNode("cam")
sop_node = cam.node("file1")
switch = hou.node("/obj/").createNode("switcher")
geo = hou.node("/obj/").createNode("geo")


for node in [sop_node, cam, switch, geo]:
    is_cam = node_matches_filter(node, hou.nodeTypeFilter.ObjCamera)
    is_obj = node_matches_filter(node, hou.nodeTypeFilter.Obj)
    is_sop = node_matches_filter(node, hou.nodeTypeFilter.Sop)
    print(f"{node} {is_cam=} {is_obj=} {is_sop=}")

    node.destroy()
```
expected output
```
file1 is_cam=False is_obj=False is_sop=True
cam7 is_cam=True is_obj=True is_sop=False
switcher1 is_cam=True is_obj=True is_sop=False
geo5 is_cam=False is_obj=True is_sop=False
```

Fix https://github.com/ynput/ayon-houdini/issues/372

### DIFF
--- a/client/ayon_houdini/api/lib.py
+++ b/client/ayon_houdini/api/lib.py
@@ -803,6 +803,35 @@ def get_output_children(output_node, include_sops=True):
     return out_list
 
 
+def node_matches_filter(
+    node: hou.Node,
+    node_type_filter: hou.nodeTypeFilter,
+) -> bool:
+    """Check if the given node matches a hou.nodeTypeFilter.
+
+    Args:
+        node (hou.Node): The node to check.
+        node_type_filter (hou.nodeTypeFilter): The node type filter to check.
+
+    Returns:
+        bool: True if the node matches the node type filter, False otherwise.
+
+    """
+    if node_type_filter == hou.nodeTypeFilter.NoFilter:
+        return True
+
+    name = node.name()
+    parent = node.parent() or hou.root()
+
+    # houdini does not seem to have a simple "node matches filter" function
+    # so we repurpose the "recursiveGlob" function here
+    return node in parent.recursiveGlob(
+        pattern=name,
+        filter=node_type_filter,  # ty: ignore[invalid-argument-type]  hou-types has invalid type for filter
+        include_subnets=False,
+    )
+
+
 def get_resolution_from_entity(entity):
     """Get resolution from the given entity.
 
--- a/client/ayon_houdini/plugins/publish/validate_scene_review.py
+++ b/client/ayon_houdini/plugins/publish/validate_scene_review.py
@@ -5,6 +5,7 @@ import pyblish.api
 from ayon_core.pipeline import PublishValidationError
 
 from ayon_houdini.api import plugin
+from ayon_houdini.api.lib import node_matches_filter
 
 
 class ValidateSceneReview(plugin.HoudiniInstancePlugin):
@@ -56,12 +57,15 @@ class ValidateSceneReview(plugin.HoudiniInstancePlugin):
             camera_node = camera_path_parm.evalAsNode()
             path = camera_path_parm.evalAsString()
             if not camera_node:
-                return "Camera path does not exist: '{}'".format(path)
-            type_name = camera_node.type().name()
-            if type_name not in {"cam", "lopimportcam"}:
-                return "Camera path is not a camera: '{}' (type: {})".format(
-                    path, type_name
-                )
+                return f"Camera path does not exist: '{path}'"
+
+            if not node_matches_filter(
+                camera_node,
+                hou.nodeTypeFilter.ObjCamera,  # ty: ignore[invalid-argument-type]  hou-types has invalid type for filter EnumValues
+            ):
+                type_ = camera_node.type().name()
+                return f"Camera path is not a camera: '{path}' (type: {type_})"
+
         elif opsource == 1:
             # LOP-level
             lop_path = rop_node.parm("loppath").evalAsString()


client/ayon_houdini/api/lib.py
```diff
@@ -803,6 +803,35 @@ def get_output_children(output_node, include_sops=True):
...
+    name = node.name()
+    parent = node.parent() or hou.root()
+
+    # houdini does not seem to have a simple "node matches filter" function
+    # so we repurpose the "recursiveGlob" function here
+    return node in parent.recursiveGlob(
+        pattern=name,
+        filter=node_type_filter,  # ty: ignore[invalid-argument-type]  hou-types has invalid type for filter
+        include_subnets=False,
+    )
```
BigRoy, line 832: Oh boy, this feels so overkill but I can't find another decent way either :(

client/ayon_houdini/plugins/publish/validate_scene_review.py
```diff
@@ -56,12 +57,15 @@ def get_invalid_camera_path(self, rop_node):
...
-            if type_name not in {"cam", "lopimportcam"}:
-                return "Camera path is not a camera: '{}' (type: {})".format(
-                    path, type_name
-                )
+                return f"Camera path does not exist: '{path}'"
+
+            if not node_matches_filter(
+                camera_node,
+                hou.nodeTypeFilter.ObjCamera,  # ty: ignore[invalid-argument-type]  hou-types has invalid type for filter EnumValues
+            ):
```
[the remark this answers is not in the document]
BigRoy: It passes the linter, I'm ok with it as is.

client/ayon_houdini/api/lib.py
MustafaJafar: well, there is actually. the doc string of [hou.node.recursiveGlob](https://www.sidefx.com/docs/houdini/hom/hou/Node.html#recursiveGlob) mentions the method is behaving similar to [hou.Node.glob](https://www.sidefx.com/docs/houdini/hom/hou/Node.html#glob) and [hou.NodeBundle](https://www.sidefx.com/docs/houdini/hom/hou/NodeBundle.html)

therefore, we can simply do this

```python
import hou

node = hou.node('/obj/testgeometry_rubbertoy1')
is_cam = False

try: 
    cam_bundle = hou.addNodeBundle("came_object_bundle")
    cam_bundle.setFilter(hou.nodeTypeFilter.ObjCamera)
    cam_bundle.addNode(node)  # only nodes match the filter are added.
    is_cam = True
    print(f"{node.path()} is an obj camera.")
except hou.OperationFailed as e:
    print(e)
finally:
    cam_bundle = hou.nodeBundle("came_object_bundle")
    if cam_bundle is not None:
        cam_bundle.destroy()
```

MustafaJafar, line 832:
```suggestion
    node_bundle_name = "_node_matches_filter_bundle"

    match_filter = False
    node_bundle = None
    try:
        node_bundle = hou.addNodeBundle(node_bundle_name)
        node_bundle.setFilter(node_type_filter)
        node_bundle.addNode(node)  # raises error if it doesn't match
        match_filter = True
    except hou.OperationFailed as e:
        # Expected failure mode when the node doesn't match the filter,
        # but could also fire for other reasons - print for debugging.
        print(
            f"node_matches_filter: Failed to add \"{node.path()}\""
            f" to the node bundle: {e}"
        )
    finally:
        # Clean up regardless of outcome. Re-fetch by name since `node_bundle`
        # may be None if addNodeBundle itself failed.
        cleanup_node_bundle = hou.nodeBundle(node_bundle_name)
        if cleanup_node_bundle is not None:
            cleanup_node_bundle.destroy()

    return match_filter
```

MustafaJafar (commented):
It works. (I used the provided snippet below). 
I'd recommend using the provided snippet below as it avoids unnecessary steps. 

Here's how it looks when fail.
image

--

Also, I'm not sure whether `hou.node.recursiveGlob` is slow or fast as Houdini is using dedicated functions under the hood which are hopefully optimized.

MustafaJafar:
It's worth mentioning that in Houdini 22 by default I can't see the switch. 
but, it passes the filter fine. 

image

client/ayon_houdini/api/lib.py
vincentullmann: I assumed the creation  + deleting of a bundle must be way slower, but it turns out it is not:

```plain
bundle              : total: 128.04ms | 0.0107ms avg
recursiveGlob       : total: 470.54ms | 0.0392ms avg
```

Given that both are sub 0.1ms, I still prefer the `recursiveGlob` approach.
It's ready only and doesn't mutate the scene (using an `hou.undos.disabler` hides that we'd mutate the hipfile... but still).

There are lots of things that could cause issues in the future when creating and deleting a bundle for every check.
Not sure if there is, but what if a studio had a "onBundleCreated" / "onBundleDestroyed" callback for example?


<details>
<summary>Benchmark script</summary>

```py
import time
import hou


def via_bundle(node: hou.Node, node_type_filter: hou.nodeTypeFilter) -> bool:
    bundle = None

    with hou.undos.disabler():
        try:
            bundle = hou.addNodeBundle("ayon_filter_probe")
            bundle.setFilter(node_type_filter)
            bundle.addNode(node)
        except hou.OperationFailed:
            return False
        else:
            return True
        finally:
            if bundle is not None:
                bundle.destroy()


def via_recursive_glob(
    node: hou.Node, node_type_filter: hou.nodeTypeFilter
) -> bool:
    if node_type_filter == hou.nodeTypeFilter.NoFilter:
        return True

    parent = node.parent() or hou.root()
    return node in parent.recursiveGlob(
        pattern=node.name(),
        filter=node_type_filter,
        include_subnets=False,
    )


def _bench(label: str, fn, nodes, filters, iterations: int = 1_000) -> float:
    start = time.perf_counter()

    for _ in range(iterations):
        for node in nodes:
            for filt in filters:
                fn(node, filt)

    elapsed = (time.perf_counter() - start) * 1000 # ms
    num_checks = len(nodes) * len(filters)
    avg = elapsed / iterations / num_checks
    print(f"{label:20s}: total: {elapsed:5.2f}ms | {avg:6.4f}ms avg")
    return elapsed

def main() -> None:
    filters = (
        hou.nodeTypeFilter.ObjCamera,
        hou.nodeTypeFilter.Obj,
        hou.nodeTypeFilter.Sop,
    )

    cam = hou.node("/obj/").node("cam8")
    nodes = [
        cam.node("file1"),
        cam,
        hou.node("/obj/").node("switcher2"),
        hou.node("/obj/").node("geo6"),
    ]

    print("------")
    _bench("bundle", via_bundle, nodes, filters)
    _bench("recursiveGlob", via_recursive_glob, nodes, filters)

main()
```
</details>

MustafaJafar: tbh, I don't have a strong preference here because while checking the the functions, I believe `.recursiveGlob` is optimized.

> Not sure if there is, but what if a studio had a "onBundleCreated" / "onBundleDestroyed" callback for example?

You are right, creating bundles is unnecessary.

MustafaJafar (approved):
It works. 

Additonally, it seems `recursiveGlob` is optimized regardless of it's name that gives the intent it'll do unnecessary things under the hood.

Node bundles can be used but they will alter the scene. even if cleaned up, it can trigger callbacks, _aren't they? IDK I couldn't find a reference on it_.

client/ayon_houdini/api/lib.py
MustafaJafar: btw, is `onBundleCreated` event real?

vincentullmann: > btw, is `onBundleCreated` event real?

not in the python api but the HDK does have events for `BUNDLE_ADDED`, `BUNDLE_DELETED` in [OP_BundleList.h](https://www.sidefx.com/docs/hdk/_o_p___bundle_list_8h_source.html) and `BUNDLE_FILTER_CHANGED` in [OP_Bundle.h](https://www.sidefx.com/docs/hdk/_o_p___bundle_8h_source.html)

to be fair I've not used any of these myself and can't say for sure that they'd trigger in the case suggested earlier, but there is a chance that these sort of side effects can occur.

PR #423, opened 2026-09-09
Consider frame padding when collecting frames

MustafaJafar:
## Changelog Description
Consider frame padding when collecting frames

## Testing notes:
1. create point cache product while scenefile's frame range is not one frame
2. validate and publish. 
3. additionally, you can see `frames` instance data which should be empty as the collection is empty (no frame pattern detected).


MustafaJafar:
There's an alternative way for this bug is to refactor this publish plugin and check if frame token is visible. 
but, honestly, I can't remember why we don't do this from the first place.

PR #424, opened 2026-09-09
422 yn 1073 houdini addon bug collect frames

### LINKED ISSUE #422: YN-1073: Houdini Addon Bug: Collect Frames

We think we've found a bug in the Houdini Addon.

When trying to publish a pointcache in Alembic format, if the Variant name ends with _# (for example, pointcache_AssetName_01), the CollectFrames collector incorrectly assumes (we believe due to a pattern issue) that the representation will consist of one file per frame. This causes the frame validator in ExtractROP to fail during validate_frames because it attempts to look for frames that do not exist (since the .abc is a single file).

We fixed it on our end by creating a downstream collector that restores the correct frame info, but we wanted to report it here so you can take a look and fix it in a future version. 

_No response_

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1547250308042723369)

_This issue was automatically created from AYON ticket [YN-1073]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=54a78a3023854faf99912e36df56324b)

MustafaJafar:
## Changelog Description
Refactor frame collection to follow the same way in collecting render.

## Testing notes:
1. create point cache product and try different scenarios with frame range and using frame token.
2. validate
3. verify collected frames are the same as the expected ones from the rop node.
image

Resolve #422

### DIFF
--- a/client/ayon_houdini/plugins/publish/collect_frames.py
+++ b/client/ayon_houdini/plugins/publish/collect_frames.py
@@ -50,7 +50,10 @@ class CollectFrames(plugin.HoudiniInstancePlugin):
         # clique.PATTERNS["frames"] supports only `.1001.exr` not `_1001.exr`.
         templates = instance.context.data["anatomy"]["templates"]
         frame_padding = int(templates["common"]["frame_padding"])
-        pattern = f"[_.](?P<index>(?P<padding>0*)\\d{{{frame_padding},}}+)\\.\\D+\\d?$"
+        pattern = (
+            f"[_.](?P<index>(?P<padding>0*)\\d{{{frame_padding},}}+)"
+            "\\.\\D+\\d?$"
+        )
         frame_collection, _ = clique.assemble(
             [file_name],
             patterns=[pattern],
--- a/client/ayon_houdini/plugins/publish/collect_frames.py
+++ b/client/ayon_houdini/plugins/publish/collect_frames.py
@@ -31,9 +31,6 @@ class CollectFrames(plugin.HoudiniInstancePlugin):
         output = lib.evalParmNoFrame(ropnode, output_parm.name())
         file_name = os.path.basename(output)
 
-        # todo: `frames` currently conflicts with "explicit frames" for a
-        #       for a custom frame list. So this should be refactored.
-
         frames = self.compute_frames(file_name, start_frame, end_frame)
         self.log.debug(f"Collected Frames: {frames}")
 
--- a/client/ayon_houdini/plugins/publish/collect_frames.py
+++ b/client/ayon_houdini/plugins/publish/collect_frames.py
@@ -48,7 +48,9 @@ class CollectFrames(plugin.HoudiniInstancePlugin):
         # will be <Collection "pointcacheBgeoCache_AB010.%d.bgeo [1001]">
         # we use a customized pattern as
         # clique.PATTERNS["frames"] supports only `.1001.exr` not `_1001.exr`.
-        pattern = "[_.](?P<index>(?P<padding>0*)\\d+)\\.\\D+\\d?$"
+        templates = instance.context.data["anatomy"]["templates"]
+        frame_padding = int(templates["common"]["frame_padding"])
+        pattern = f"[_.](?P<index>(?P<padding>0*)\\d{{{frame_padding},}}+)\\.\\D+\\d?$"
         frame_collection, _ = clique.assemble(
             [file_name],
             patterns=[pattern],
--- a/client/ayon_houdini/plugins/publish/collect_frames.py
+++ b/client/ayon_houdini/plugins/publish/collect_frames.py
@@ -1,8 +1,8 @@
 # -*- coding: utf-8 -*-
 """Collector plugin for frames data on ROP instances."""
 import os
+import re
 import hou  # noqa
-import clique
 import pyblish.api
 from ayon_houdini.api import lib, plugin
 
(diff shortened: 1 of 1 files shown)


MustafaJafar:
There's an alternative way for this bug is to refactor this publish plugin and check if frame token is visible. 
but, honestly, I can't remember why we don't do this from the first place.

BigRoy (changes requested):
Try to make it a bit more strict.

Perhaps we should also check `if not parm.isTimeDependent()` to not assume it to be a frame range at all. Then preferably, if it is we accurately detect where the frame number is in the filename, and rely upon that.

Enforcing the frame padding isn't strict enough.

Also, nothing enforces the actual ROP to have been set to padding that matches your project settings - so it can still break things. Anatomy applies to your outputs, ROPs shouldn't necessarily be restricted due to this.

MustafaJafar:
> Try to make it a bit more strict.
> 
> Perhaps we should also check `if not parm.isTimeDependent()` to not assume it to be a frame range at all. Then preferably, if it is we accurately detect where the frame number is in the filename, and rely upon that.
> 
> Enforcing the frame padding isn't strict enough.
> 
> Also, nothing enforces the actual ROP to have been set to padding that matches your project settings - so it can still break things. Anatomy applies to your outputs, ROPs shouldn't necessarily be restricted due to this.

The new change is generic and it works if it's one frame or many as it depends on the existence of frame token.

client/ayon_houdini/plugins/publish/collect_frames.py
```diff
@@ -28,41 +28,43 @@ def process(self, instance):
         # Evaluate the file name at the first frame.
         ropnode = hou.node(instance.data["instance_node"])
         output_parm = lib.get_output_parameter(ropnode)
-        output = output_parm.evalAtFrame(start_frame)
+        output = lib.evalParmNoFrame(ropnode, output_parm.name())
         file_name = os.path.basename(output)
 
         # todo: `frames` currently conflicts with "explicit frames" for a
         #       for a custom frame list. So this should be refactored.
```
MustafaJafar: on a separate note, this comment came from this [PR](https://github.com/ynput/OpenPype/pull/1821). 

tbh, I don't know what conflicts it refers to. all what I know that on a later plugin, we use this `frames` instance data to generate `expectedFiles`. 
difference, `frames`  is a list of file names where `expectedFiles` list of file paths (mostly `staging_dir/file_name`). 

https://github.com/ynput/ayon-houdini/blob/5c6d85bd8e40667cd8b9a0f87ac2182060a2ed0d/client/ayon_houdini/plugins/publish/collect_cache_farm.py#L33-L40

BigRoy: Once `instance.data["frames"]` was used to represent the explicit frames to render (in ayon core). So that instead of frame start - frame end it'd render e.g. only `1001, 1002, 1005, 1009`. However, I don't believe that's still the case so the comment can be ignored.

BigRoy:
```suggestion
```

client/ayon_houdini/plugins/publish/collect_frames.py
```diff
@@ -28,41 +28,43 @@ def process(self, instance):
         # Evaluate the file name at the first frame.
         ropnode = hou.node(instance.data["instance_node"])
         output_parm = lib.get_output_parameter(ropnode)
-        output = output_parm.evalAtFrame(start_frame)
+        output = lib.evalParmNoFrame(ropnode, output_parm.name())
```
BigRoy, line 31: Can we fix `evalParmNoFrame` to support the `${F}` syntax? e.g.

E.g.
```python
def evalParmNoFrame(node, parm, pad_character="#"):

    parameter = node.parm(parm)
    assert parameter, "Parameter does not exist: %s.%s" % (node, parm)

    # If the parameter has a parameter reference, then get that
    # parameter instead as otherwise `unexpandedString()` fails.
    parameter = get_top_referenced_parm(parameter)

    # Substitute out the frame numbering with padded characters
    try:
        raw = parameter.unexpandedString()
    except hou.Error as exc:
        print("Failed: %s" % parameter)
        raise RuntimeError(exc)

    def replace(match):
        padding = 1
        n = match.group("p1") or match.group("p2")
        if n and int(n):
            padding = int(n)
        return pad_character * padding

    # p1, p2 represent named groups for padding
    regex = (
        r"(\$F(?P<p1>[0-9]*))|"
        r"(\$\{F(?P<p2>[0-9]*)\})|"
    )
    print(regex)
    expression = re.sub(regex, replace, raw)

    with hou.ScriptEvalContext(parameter):
        return hou.expandStringAtFrame(expression, 0)
```

Supporting `@Frame` may be trickier, because then you'd also need to evaluate whether it's in the expression part of the string between back ticks?

MustafaJafar: I believe it'll be much better to have it in a separate PR, I'll push one later. as the focus here is on the collecting frames plugin.

BigRoy: Stupid question, but is there any reason we're not just doing this:

```python
if start_frame != end_frame and parm.isTimeDependent():
    files = [
        parm.evalAtFrame(frame)
        for frame in range(start_frame, end_frame+1)
    ]
else:
    files = parm.evalAtFrame(start_frame)
```

This may still need a check to uniqify the results in the first case, because a parm may sometimes think it's time dependent if it has an expression however that expression does not evaluate a time. Not sure if they fixed that.

BigRoy (approved):
Looks ok to me - **but** I was wondering why we're not doing [this](https://github.com/ynput/ayon-houdini/pull/424/changes#r3976301136). Is `parm.evalAtFrame()` too slow?

Can you update the PR title by the way - it's what makes it's way into the changelog on new release, would be good to have it clear and descriptive.

client/ayon_houdini/plugins/publish/collect_frames.py
MustafaJafar: well,  for 1, 10, 100 frames 
this was not tested on a heavy scene though.

using `evalAtFrame`
```
2026/09/10 20:40:23 DEBUG: Elapsed time 0.0001 seconds to complete.
2026/09/10 20:40:23 DEBUG: Elapsed time 0.0003 seconds to complete.
2026/09/10 20:43:06 DEBUG: Elapsed time 0.0013 seconds to complete.
```

using the current method
```
2026/09/10 21:05:31 DEBUG: Elapsed time 0.0001 seconds to complete.
2026/09/10 21:05:31 DEBUG: Elapsed time 0.0002 seconds to complete.
2026/09/10 20:40:23 DEBUG: Elapsed time 0.0003 seconds to complete.
```

MustafaJafar: this is the code I tested for `evalAtFrame`
```python
        if start_frame != end_frame and output_parm.isTimeDependent():
            files = [
                os.path.basename(output_parm.evalAtFrame(frame))
                for frame in range(start_frame, end_frame+1)
            ]
            staging_dir = os.path.dirname(files[0])
        else:
            files = os.path.basename(output_parm.evalAtFrame(start_frame))
            staging_dir = os.path.dirname(files)
```

BigRoy: Ok, I think it'd be more accurate - but I'm ok with what we have for now.

We may want to look at doing the frame eval some day though if we hit edge cases.

MustafaJafar: I've also these variations

this for the current method
```python
        if start_frame != end_frame and output_parm.isTimeDependent():
            output = lib.evalParmNoFrame(ropnode, output_parm.name())
            file_name = os.path.basename(output)
            frames = self.compute_frames(file_name, start_frame, end_frame)
        else:
            output = output_parm.evalAtFrame(start_frame)
            frames = os.path.basename(output)
```
this for `evalAtFrame` 
```python
        if start_frame != end_frame and output_parm.isTimeDependent():
            output = lib.evalParmNoFrame(ropnode, output_parm.name())
            frames = []
            for frame in range(start_frame, end_frame+1):
                output = output_parm.evalAtFrame(frame)
                frames.append(os.path.basename(output))
        else:
            output = output_parm.evalAtFrame(start_frame)
            frames = os.path.basename(output)
```

they also ran in the same time mentioned above.

